PR #29, opened 2015-02-17
VERISKEVALDEV-30 Replacing Natural Join With Inner join

sambaheerathan:
Replaced the Natural Join With Inner join


sambaheerathan:
Please Ignore this pull request

PR #53, opened 2015-03-06
Remove hard dependency to the class 'WSO2ExternalAPIStore' Release 1.9.0

binalip:
'WSO2APIStore' class is no longer directly referred  from the API Manager code. The class name for the external API store reads from the 'default-external-api-store.xml'.A new attribute which is  'class' is holding the name for the External API Store class.


lalaji:
Closing this pull request as it contains  a mistake and Binali is working on creating a new pull request.

PR #142, opened 2015-04-01
Adding feature.properties and change the save path

thushara35:


features/etc/feature.properties
```diff
@@ -0,0 +1,216 @@
+################################################################################
+# Copyright 2009 WSO2, Inc. (http://wso2.com)
```
prasa7: I think you need to change this year to 2014 or 2015

thushara35: Thanks will do.

PR #190, opened 2015-04-17
Fixing APIMANAGER-3452: Updating template.jag

rasika90:
Fixing APIMANAGER-3452: Incorrect POST methods in throttling tab in API Store

### DIFF
--- a/components/apimgt/api-store-web/src/site/themes/fancy/templates/api/throttling/template.jag
+++ b/components/apimgt/api-store-web/src/site/themes/fancy/templates/api/throttling/template.jag
@@ -120,6 +120,7 @@ var i, length, app, thumbURL, api = outputs.api, templatesList = outputs.uriTemp
 
 
                                 <% getAuthVal=putAuthVal=postAuthVal=deleteAuthVal=optionAuthVal=""; %>
+                                <% getThrottlingTier=postThrottlingTier=deleteThrottlingTier=optionsThrottlingTier=putThrottlingTier=""; %>
                             </tr>
                             <%
                             }}%>
--- a/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store/site/themes/fancy/templates/api/throttling/template.jag
+++ b/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store/site/themes/fancy/templates/api/throttling/template.jag
@@ -42,6 +42,7 @@ var i, length, app, thumbURL, api = outputs.api, templatesList = outputs.uriTemp
                                                   var methods = templatesList[i][1].split(",");
                                                   var authTypes = templatesList[i][2].split(",");
                                                   var throttlingTiers = templatesList[i][3].split(",");
+                                                  //below vars are reseted at the end of the loop
                                                   var getVal,putVal,postVal,deleteVal,optionVal,method;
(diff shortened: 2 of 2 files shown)


rswijesena:
I am closing this pull request as this has been sent to wrong branch. All pull requests should goes to release 1.9.0 branch. Merged the change manually and created new pull request https://github.com/wso2/carbon-apimgt/pull/224

components/apimgt/api-store-web/src/site/themes/fancy/templates/api/throttling/template.jag
```diff
@@ -120,10 +120,11 @@ var i, length, app, thumbURL, api = outputs.api, templatesList = outputs.uriTemp
 
 
                                 <% getAuthVal=putAuthVal=postAuthVal=deleteAuthVal=optionAuthVal=""; %>
+                                <% getThrottlingTier=postThrottlingTier=deleteThrottlingTier=optionsThrottlingTier=putThrottlingTier=""; %>
```
prabathabey: I believe the idea here is to initialize multiple variables at once? Would be much readable and elegant if the following format was used.

String getThrottlingTier, postThrottlingTier, deleteThrottlingTier = "";

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store/site/themes/fancy/templates/api/throttling/template.jag
```diff
@@ -118,7 +119,7 @@ var i, length, app, thumbURL, api = outputs.api, templatesList = outputs.uriTemp
 
                                     <% getVal=putVal=postVal=deleteVal=optionVal=""; %>
 
-
+                                <% getThrottlingTier=postThrottlingTier=deleteThrottlingTier=optionsThrottlingTier=putThrottlingTier=""; %>
```
prabathabey: Same goes here as well.

components/apimgt/api-store-web/src/site/themes/fancy/templates/api/throttling/template.jag
rasika90: In JavaScript, variables are functional scoped rather than block scoped. The intention is to clear up all variables into empty on each loop cycle. The other way would be;

getAuthVal="";
putAuthVal="";
postAuthVal="";
deleteAuthVal="";
optionAuthVal="";
getThrottlingTier="";
postThrottlingTier="";
deleteThrottlingTier="";
optionsThrottlingTier="";
putThrottlingTier="";

But It consumes more space in my opinion.

DilanUA: if the intention is to clear the values of the variables, better assigning undefined to the variable rather than an empty string. For better readability, format the statement with spaces as follows.
getAuthVal = putAuthVal = postAuthVal = deleteAuthVal = optionAuthVal = undefined;

manuranga: It's not a best practice to use `undefined`, you can use `null`. but in this case we can't do that since variable is used in other places, and can result in a null pointer or null getting printed to web page.
Best way to avoid this is to not to assign values to variables just in `if`s inside `for`, and set and an appropriate value in `else` as well.

PR #275, opened 2015-04-30
Release 1.9.0

madusankapremaratne:
Fix for APIMANAGER-3489

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.hostobjects/src/main/java/org/wso2/carbon/apimgt/hostobjects/APIStoreHostObject.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.hostobjects/src/main/java/org/wso2/carbon/apimgt/hostobjects/APIStoreHostObject.java
@@ -78,6 +78,7 @@ import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 
 import java.net.*;
 import java.rmi.RemoteException;
+import java.sql.Array;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.hostobjects/src/main/java/org/wso2/carbon/apimgt/hostobjects/APIStoreHostObject.java
```diff
@@ -3330,15 +3330,18 @@ public static boolean jsFunction_updateApplication(Context cx,
...
             for (Application app : apps) {
-                if (app.getName().equals(name)) {
-                    return false;
-                }
+                appsMap.put(app.getName(), app);
             }
 
             for (Application app : apps) {
-                if (app.getName().equals(oldName)) {
+                if(appsMap.containsKey(app.getName())){
```
nuwand: We seem to be iterating the same list twice. In the first iteration we're adding all items to a map and then in the second iteration we're checking if those items exists. Since we're adding all items in the first iteration (appsMap.put(app.getName(),..), they would definitely be available when we're doing the second iteration (if(appsMap.containsKey(app.getName()))).

madusankapremaratne: The first iteration reads the app names and put it into the map. (This happens before updating the apps)
And the second iteration checks whether the newName of the app (after update) is a name of already existing app by calling appsMap.containsKey(newName)
The code has been changed and the actual check in the code is if (app.getName().equals(oldName))
I have changed the name parameter to newName to reduce the confusion (Code improved in 8ecc7ed5684940f1f69ef8943428a017e8ab417d)

PR #383, opened 2015-05-13
fixed APIMANAGER-3550

kasun32:
Added a Subscription status check to resolve the issue.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
@@ -1948,7 +1948,8 @@ public class ApiMgtDAO {
                     "   AND APP.APPLICATION_ID=SUBS.APPLICATION_ID " +
                     "   AND API.API_ID=SUBS.API_ID" +
                     "   AND APP.NAME= ? " +
-                    "   AND SUBS.SUBS_CREATE_STATE = '" + APIConstants.SubscriptionCreatedStatus.SUBSCRIBE + "'";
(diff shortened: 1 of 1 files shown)


abimarank:
Hi Kasun,

In the Subscription page of the API Publisher, We have to show, BLOCKED, UNBLOCKED API Subscription. But, with this pull request, you made changes to list on the UNBLOCKED API Subscription.

It we don't show the BLOCKED API Subscription, we won't be able to unblock them through the publisher.

kasun32:
@abimarank I checked on this and found out that API Provider's Subscriptions page uses getAllAPIUsageByProvider() to load subscriptions and I have only changed getPaginatedSubscribedAPIs() which is used by API Store.

This change doesn't affect API Provider as far as I understand.

kasun32:
Hi @abimarank 

I had some misunderstanding about the issue and the earlier PR was for the API Store. Did revert the earlier commit and updated the PR with relevant changes to the API Publisher.

Please review and merge.

abimarank:
Following are status' of API subscription

```java
public static class SubscriptionStatus {
    public static final String BLOCKED = "BLOCKED";
    public static final String PROD_ONLY_BLOCKED = "PROD_ONLY_BLOCKED";
    public static final String UNBLOCKED = "UNBLOCKED";
    public static final String ON_HOLD = "ON_HOLD";
    public static final String REJECTED = "REJECTED";
}
```
Only subscription with BLOCKED, PROD_ONLY_BLOCKED, UNBLOCKED status should be listed in subscription tab of the API Publisher. REJECTED, ON_HOLD staus cann't be considered as successfull subscriptions.
Can you please filer ON_HOLD status also here with addition to REJECTED?
I'm reopning the bug.

kasun32:
@abimarank I noticed that earlier. But ON_HOLD status is already handled in API Publisher using a tip saying that the subscription is pending for approval. Does it still has to be removed? If yes I will do the necessary changes.

PR #498, opened 2015-05-26
Fixing APIM3424 : adding warnings for both tier and subscription

malinthaprasan:


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/site/themes/default/templates/item-manage/js/api-manage.js
```diff
@@ -13,6 +13,22 @@ function validate_tiers(){
     return false;
 }
 
+function validate_subscription(){
```
harsha89: Better use camel case for method signature

malinthaprasan: Fixed in #1494(https://github.com/wso2/carbon-apimgt/commit/6a4f74a2d079f3fbf27a772832e83b12d47adf3e)

PR #585, opened 2015-06-04
Fixing getUUID in publisher and store

thushara35:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
```diff
@@ -3143,5 +3143,17 @@ public JSONArray getSubscriptions(String providerName, String apiName, String ve
 
     }
 
+    /**
```
harsha89: Since this method use in both provider and consumer shall we move this to abstract api manager class. :+1:

thushara35: Updated the PR accordingly. thanks

PR #793, opened 2015-06-29
Updating JMX attributes for an end-point 

ChamNDeSilva:
Updating JMX attributes for an end-point Adding fixes of https://wso2.org/jira/browse/ESBJAVA-3271 to API Manager.


dependencies/synapse/modules/core/src/main/java/org/apache/synapse/endpoints/AbstractEndpoint.java
```diff
@@ -191,9 +191,9 @@ public void setName(String endpointName) {
                 return;
             }*/
             metricsMBean = new EndpointView(endpointName, this);
-            if(metricsMBean != null) {
-                metricsMBean.destroy();
-            }
+//            if(metricsMBean != null) {
```
laki88: is there any reason to commit commented code?

ChamNDeSilva: This commented code has been removed in the commit dc2e4245031771730bc811c0dd6e4ada87ee2aa9

PR #835, opened 2015-07-20
fixed jira APIMANAGER-3983

ruks:
https://wso2.org/jira/browse/APIMANAGER-3983


uvindra:
This fix is no longer required because of changes done in this https://github.com/uvindra/carbon-apimgt/commit/95ec882d46c633281fdabe23770fb13c2a2afe4f, the code in question has changed. Therefore we can abandon this pull request, no need to merge.

PR #902, opened 2015-08-13
Fixing https://wso2.org/jira/browse/APIMANAGER-3963

chamindias:
Fixing issue :
"Need to restart Gateways to enable/disable analytics in a distributed deployment." (APIMANAGER-3963)

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/pom.xml
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/pom.xml
@@ -82,12 +82,21 @@
         </dependency>
         <dependency>
              <groupId>org.wso2.carbon</groupId>
-             <artifactId>org.wso2.carbon.utils</artifactId>       			 
+             <artifactId>org.wso2.carbon.utils</artifactId>
          </dependency>
         <dependency>
             <groupId>org.wso2.carbon</groupId>
             <artifactId>org.wso2.carbon.identity.oauth.stub</artifactId>
             <version>4.2.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.axis2</groupId>
+            <artifactId>axis2-kernel</artifactId>
+            <version>1.6.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.wso2.carbon.apimgt</groupId>
+            <artifactId>org.wso2.carbon.apimgt.statsupdate.stub</artifactId>
         </dependency>
     </dependencies>
 
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/statshandler/GatewayStatsUpdater.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/statshandler/GatewayStatsUpdater.java
@@ -1,0 +1,108 @@
+/*
+*  Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+*
+*  WSO2 Inc. licenses this file to you under the Apache License,
+*  Version 2.0 (the "License"); you may not use this file except
+*  in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+
+package org.wso2.carbon.apimgt.gateway.statshandler;
+
+import org.apache.axis2.clustering.ClusteringAgent;
+import org.apache.axis2.clustering.ClusteringFault;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.wso2.carbon.apimgt.api.APIManagementException;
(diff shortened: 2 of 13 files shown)


components/apimgt/org.wso2.carbon.apimgt.gateway/pom.xml
```diff
@@ -82,13 +82,22 @@
...
          </dependency>
         <dependency>
             <groupId>org.wso2.carbon</groupId>
             <artifactId>org.wso2.carbon.identity.oauth.stub</artifactId>
             <version>4.2.3</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.axis2</groupId>
+            <artifactId>axis2-kernel</artifactId>
+            <version>1.6.2</version>
```
jaadds: Versions have to go into the root pom. Please put the entire dependency under dependencyManagement section and refer it from there.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/statshandler/GatewayStatsUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+                log.debug("Sending cluster message to Gateway domain to update stats publishing status.");
+                clusteringAgent.sendMessage(new StatClusterMessageUtil(statUpdateStatus), true);
+            }
+        } catch (ClusteringFault clusteringFault) {
+            //error is only logged because initial gateway has modified the status
+            log.error("Failed to send cluster message to Gateway domain and update stats publishing status.");
+        } catch (APIManagementException e) {
+            log.error("Error occurred while adding BAMServerProfile");
+        } catch (Exception e) {
+            log.error("Error occurred while updating EventingConfiguration");
```
jaadds: Also mention what can be the outcome of a failure. What functionality can get impaired.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/StatClusterMessageUtil.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.wso2.carbon.apimgt.impl.APIManagerAnalyticsConfiguration;
+
+import java.io.Serializable;
+
+/**
+ * This class provides the definition of the cluster message which is initiated from the
+ * web service call from publisher node
+ */
+public class StatClusterMessageUtil extends ClusteringMessage implements Serializable  {
```
jaadds: Clustering message is already Serializable. So no need of making it Serializable again.

jaadds: Something like StatUpdateClusterMessage would be more suitable IMO. Since this represents a message, better not to end it like Util.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/StatsUpdateUtil.java
```diff
@@ -0,0 +1,100 @@
...
+                if (receiverUrl == null || user == null || password == null) {
+                    receiverUrl = "";
+                    user = "";
+                    password = "";
+                }
+                stub.updateStatPublishGateway(receiverUrl, user, password, updatedStatus);
+
+                //send cluster message to publisher-store domain
+                storePublisherMessageAgent.updateStatsPublishStore(updatedStatus);
+            }
```
nuwand: We should not throw exceptions here. If the message update fails for one Gateway, we should still continue to send the update to the other Gateways. Therefore we need to continue even on error and then provide a log of the failures later.

pom.xml
```diff
@@ -362,6 +362,12 @@
 
             <dependency>
                 <groupId>org.wso2.carbon.apimgt</groupId>
+                <artifactId>org.wso2.carbon.apimgt.statsupdate.stub</artifactId>
+                <version>4.2.2-SNAPSHOT</version>
```
jaadds: Same as earlier - remove the version.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/StatsUpdateUtil.java
```diff
@@ -0,0 +1,100 @@
+/*
```
jaadds: Shall we merge this code into APIProviderImpl rather than having a whole new class for it. Since the code in this class gets triggered by Publisher App, better to put it inside APIProviderImpl.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/StatUpdateStorePublisherDomain.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.wso2.carbon.core.AbstractAdmin;
+
+import java.io.Serializable;
+
+/**
+ * This class contains the functions to update stat publishing in Store-Publisher domain
+ */
+public class StatUpdateStorePublisherDomain extends AbstractAdmin implements Serializable {
```
jaadds: No need to make this Serializable. And this doesn't need to be a AdminService. Since what this does is creating a cluster message and send it, same thing could have been done from StatsUpdateUtil, isn't it?

PR #1080, opened 2015-09-30
Modifying queries in "APIUsageStatisticsClient" to be compatible with Oracle/H2/MSSQL/PostgreSQL databases.

chamindias:
This is the fix to support Oracle/H2/MSSQL/PostgreSQL databases, when getting analytics data.


chamindias:
Public JIRA: APIMANAGER-4145

PR #1219, opened 2015-11-02
Fix UI alignment issues in Publisher, related to swagger doc and import swagger pop-up menu.

chamindias:


chamindias:
JIRA ID : https://wso2.org/jira/browse/REGISTRY-3091

PR #1413, opened 2015-12-11
Adding missing tenant-conf.json modification in core features

malinthaprasan:


features/apimgt/org.wso2.carbon.apimgt.core.feature/src/main/resources/tenant/tenant-conf.json
```diff
@@ -1,5 +1,29 @@
...
+        "Name": "API_CREATOR_SCOPE",
+        "Roles": "admin"
+      },
+      {
+        "Name": "API_CREATOR_PUBLISHER_SCOPE",
+        "Roles": "admin"
+      },
+      {
+        "Name": "API_SUBSCRIBER_SCOPE",
+        "Roles": "Internal/subscriber"
```
sanjeewa-malalgoda: Why this is special case? IMO we should use admin for all and let users to figure out what they exactly need.

nuwand: We discussed about this and decided to leave it as it is. Since Internal/subscriber role exists by default this doesn't cause any issues. And for demo purposes, we can sign-up a user from the Store and invoke the Store REST APIs by a token obtained by that user.

PR #1517, opened 2016-01-13
Fixing APIMANAGER-2280 and APIMANAGER-4064

tharikaGitHub:
This is a simple fix for the above issues. (A basic check to see whether it is a url) So this prevents the user entering text like 'aaaaaa' for example, in the respective fields.

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/site/themes/default/templates/utils/custom-validation/custom-validation.js
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/site/themes/default/templates/utils/custom-validation/custom-validation.js
@@ -43,6 +43,15 @@ $(document).ready(function() {
         return value.indexOf("{}") == -1
     }, 'Empty curly brackets "{}" are not allowed in context field.');
 
+    $.validator.addMethod('validateUrl', function(value, element){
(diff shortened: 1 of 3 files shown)


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/site/themes/default/templates/utils/custom-validation/custom-validation.js
```diff
@@ -43,6 +43,15 @@ $(document).ready(function() {
         return value.indexOf("{}") == -1
     }, 'Empty curly brackets "{}" are not allowed in context field.');
 
+    $.validator.addMethod('validateUrl', function(value, element){
+        var validUrlRegex = /^(http|https):\/\/(.)+/g;
```
harsha89: Is the regex validate urls starting from http and https with following two forward slashes. with (.)+ are we checking it's should contain one or more . ?

tharikaGitHub: No, with (.)+ we match a group of one or more characters except the newline.

harsha89: :+1:

PR #1550, opened 2016-01-22
Fixing some code quality issues suggested by Sonar

malinthaprasan:
Fixes include:

1. Replaced places where Throwable is caught into Exception. Catching Throwable is not good as this might include some errors an application should not try to recover such as OutOfMemoryException
2. Fixed places where original exception is not thrown/logged.


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIMWSDLReader.java
```diff
@@ -169,8 +169,9 @@ private void setServiceDefinitionForWSDL2(org.apache.woden.wsdl20.Description de
                 }
             }
         } catch (Exception e) {
-            log.error("Error occurred while getting the wsdl address location", e);
-            throw new APIManagementException(e);
+	        String errorMsg = "Error occurred while getting the wsdl address location";
```
harsha89: slight formatting issue

PR #1567, opened 2016-01-26
Fix for APIMANAGER-4488

praminda:
*getApplications* method was used to get the correct application to
update. This method takes too much time to execute when number of
applications is high (fetches application key information per each
application from the database). Therefore *getApplicationByName*
was used to directly retrieve only the application required and
update/remove.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.hostobjects/src/main/java/org/wso2/carbon/apimgt/hostobjects/APIStoreHostObject.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.hostobjects/src/main/java/org/wso2/carbon/apimgt/hostobjects/APIStoreHostObject.java
@@ -3007,25 +3007,20 @@ public class APIStoreHostObject extends ScriptableObject {
 
     public static boolean jsFunction_removeApplication(Context cx, Scriptable thisObj, Object[] args, Function funObj)
             throws ScriptException, APIManagementException {
-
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.hostobjects/src/main/java/org/wso2/carbon/apimgt/hostobjects/APIStoreHostObject.java
```diff
@@ -3094,39 +3087,35 @@ public static boolean jsFunction_updateApplication(Context cx, Scriptable thisOb
...
-                    application.setId(app.getId());
-                    application.setTier(tier);
-                    application.setCallbackUrl(callbackUrl);
-                    application.setDescription(description);
-                    apiConsumer.updateApplication(application);
-                    return true;
+            // get application with new name if exists
+            Application application = apiConsumer.getApplicationsByName(username, newName, groupingId);
+            if (!newName.equals(oldName)) {
+
```
harsha89: no need of new line here. Minor thing may be you can keep it as a reference for future codings

praminda: I think it is an Java standard to keep a blank space before a block or single line comment.

PR #2418, opened 2016-06-03
Removing non-functioning loganalyzer-dashboard block jaggery file.

ksdperera:


ksdperera:
This is fixed by another PR.

PR #2569, opened 2016-06-24
adding null check

harsha89:


harsha89:
This need to merge after siddhi version upgrade after testing it

harsha89:
Since siddhi find fuction to dealt with null inputs this no longer needed

PR #2699, opened 2016-07-11
added a new admin service operation - revokeTokensOfUserByApp

abeykoon:
This contains a feature for API manager for revoking access tokens issued for a given user (user tokens) under a given application. Key Manager server's admin service [1] should be called and it will revoke all the tokens and call gateway server so that key caches are invalidated.


abeykoon:
This PR is ready for review.

PR #2734, opened 2016-07-13
Align the publisher stat graphs.

chamindias:


chamindias:
This is related to https://wso2.org/jira/browse/APIMANAGER-5189

PR #2977, opened 2016-10-13
Add Host header for sanity

mathieu-pousse:
When the endpoint we are accessing is behind an haproxy or any system that is based on the Host header to dispatch to the proper backend, the Host header is kind of "mandatory".

Not setting the Host header can lead to invalid TLS certificate served or inconsistent behaviour.

Anyway, the header is properly set when running the api manager but was missing when clicking the "Test" button on the GUI. It ends to an error in the UI that is not blocking. 

This PR get rid of this error and simulate the HEAD request with something more similar to the production behaviour

Hope it is clear


abimarank:
@mathieu-pousse Thanks for the suggestion. But, We need to handle HTTP urls as well, also, we need to append port number too if it's not 80.

Can you enhance the PR with this suggestion?

mathieu-pousse:
Here it is, can you please mark this issue with label hacktoberfest (https://hacktoberfest.digitalocean.com/)

PR #2980, opened 2016-10-14
Adding API Gateway URL Feature

ChamNDeSilva:


sajinidesilva:
Its unnecessary to have commit messages like "removing spaces". So squash[1] them before u merge. 

[1] https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

PR #3114, opened 2016-11-02
Add support for Access-Control-Max-Age

mathieu-pousse:
Add the `Access-Control-Max-Age` header to avoid a lot of OPTIONS requests

The CORS headers can be cached by the browser accordingly to the `Access-Control-Max-Age` header. This will save dozen of OPTIONS request.

To enable this headers, simply set a property that defines the max age (in seconds) in the CORSRequestHandler: 

```xml
<handler class="org.wso2.carbon.apimgt.gateway.handlers.security.CORSRequestHandler">
    <property name="apiImplementationType" value="ENDPOINT"/>
    <property name="accessControlMaxAge" value="3600"/>
</handler>
```

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/CORSRequestHandler.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/CORSRequestHandler.java
@@ -41,7 +41,10 @@ import java.util.*;
 public class CORSRequestHandler extends AbstractHandler implements ManagedLifecycle {
 
 	private static final Log log = LogFactory.getLog(CORSRequestHandler.class);
-	private String apiImplementationType;
+	private String accessControlMaxAge;
+    // default to 15 minutes
+    private Integer accessControlMaxAgeComputed = 15 * 60;
+    private String apiImplementationType;
 	private String allowHeaders;
 	private String allowCredentials;
 	private Set<String> allowedOrigins;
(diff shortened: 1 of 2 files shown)


abimarank:
@mathieu-pousse Thanks for the contribution, We are having a discussion on this feature.

abimarank:
@mathieu-pousse Can you please add 15 mins as the default value for the cache in code level. You have declared `accessControlMaxAge` as an Integer type, but, from the configuration, it's set as String and we need to convert it to Integer.

Can you please enhance the PR with above suggestions?

abimarank:
@mathieu-pousse Can you please look into this? We are going to release next API Manager in another couple of weeks.

mathieu-pousse:
I am back. 

I am not sure to understand. Do you want me to set a default value in `CORSRequestHandler.java` ?

This will breakup the backward compatibility and force all the existing to run with the `Access-Control-Max-Age`. Is that what you want ?

Maybe I can convert the field to a String and if a value is set AND is a valid Integer then we use it and add the header, otherwise we silently ignore it and keep running as it was before ?

Let me know

abimarank:
How will it affect backward compatibility? Yes, have to set 15 mins as the default value in `CORSRequestHandler.java`.

Yes, You are correct on the conversion.

mathieu-pousse:
OK

About the backward compatibility, if we add a new feature that changes the behavior of the API Manager Client and enable it by default (15 minutes), if in some circumstances it changes the behavior of the browser up to creating a bug in the client API, this new feature break the backward compatibility.

BTW, with the default value, there is no way of disabling it

Here is my update of the code:
if the user sets a value (read as string) it is parsed and if it is a correct value it will be used

let me know if you really want a default value.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/CORSRequestHandler.java
```diff
@@ -79,7 +80,14 @@ void initializeHeaders() {
 		if (allowedMethods == null) {
 			allowedMethods = APIUtil.getAllowedMethods();
 		}
-
+		if (accessControlMaxAge != null && accessControlMaxAge.trim().length() == 0) {
```
abimarank: `accessControlMaxAge.trim().length() == 0` should be `accessControlMaxAge.trim().length() != 0`

abimarank:
Still, I don't understand the compatibility issue of having the default value. 

We can turn it off by adding the configuration with 0 as the value.

mathieu-pousse:
done, 0 or negative value disable the feature

harsha89:
@mathieu-pousse please create a JIRA ticket for the changes.

harsha89:
Closing this temporarily.

PR #3286, opened 2016-11-23
c5 added template generate and publishing

ruks:


components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/api/APIGateway.java
```diff
@@ -0,0 +1,28 @@
...
+*
+*/
+package org.wso2.carbon.apimgt.core.api;
+
+import org.wso2.carbon.apimgt.core.models.API;
+
+/**
+ * Used to define Gateway related operations
+ */
+public interface APIGateway {
```
sanjeewa-malalgoda: We can rename this as APIGatewayPublisher

ruks: fixed

PR #3522, opened 2017-01-03
commit endpoint specific swagger changes

tharindu1st:


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher/src/main/resources/publisher-api.yaml
```diff
@@ -2237,6 +2237,201 @@ paths:
...
+        This operation can be used to retrieve the list of endpoints available.
+      tags:
+        - Endpoint (Collection)
+      responses:
+        200:
+          description: |
+            OK.
+            Endpoint list is returned.
+          schema:
+            type: array
```
malinthaprasan: Shall we define this as a List model like EnvironmentList. Then we can specify other attributes like "count"  and pagination related attributes (if required)

PR #3733, opened 2017-02-23
Fix for https://wso2.org/jira/browse/APIMANAGER-5623

tharikaGitHub:


components/web/apps/org.wso2.carbon.apimgt.store/src/main/public/js/apis/{apiId}/index.js
```diff
@@ -320,8 +320,23 @@ $(function () {
                             $("#subscribe-button").html('Subscribe');
                             $("#subscribe-button").removeAttr('disabled');
                             var subscription = jsonData.obj;
-
-                            location.href = contextPath + "/apis/" + apiId;
+                            var message = "Subscription created successfully";
```
dakshika: Better to add this to the support of i18n

tharikaGitHub: Thanks for the review. Since we have discontinued UUF in favour of React.js, this file is no longer available in this repository.

PR #3766, opened 2017-03-01
Enhancement to subscription successful message in store

tharikaGitHub:


components/web/apps/org.wso2.carbon.apimgt.store/src/main/fragments/application-details/public/js/tabs.js
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Javascript to enable link to tab. Change hash for page-reload, If hashed page name is available load that page
+ * If no hash value is present load the default page which is overview-tab
+ */
+function loadFromHash() {
+    var hash = document.location.hash;
```
dakshika: location.hash is not safe for IE, recommend to use window.location

tharikaGitHub: Thanks for the review. Since we have discontinued UUF in favour of React.js, this file is no longer available in this repository.

PR #3802, opened 2017-03-10
Integrating coverage plugin and adding unit tests - Sprint 1

tharikaGitHub:


tharikaGitHub:
The check on the 70% code coverage has been commented out so that the build wont break until the coverage goal is met.

tharikaGitHub:
Sure, will do the changes. Thanks for the review.

PR #3845, opened 2017-03-21
update error handling for workflow

chamilaadhi:


components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIStoreImpl.java
```diff
@@ -603,14 +603,14 @@ public void removeObserver(EventObserver observer) {
     }
     
     public void completeWorkflow(WorkflowExecutor workflowExecutor, Workflow workflow) throws APIManagementException {
+        
+        if (workflow.getWorkflowReference() == null) {
+            String message = "Error while changing the workflow Workflow. Missing reference";
```
prasa7: Does the word workflow repeated by mistake? (workflow Workflow)

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIStoreImpl.java
```diff
@@ -662,23 +654,34 @@ public void completeWorkflow(WorkflowExecutor workflowExecutor, Workflow workflo
...
-    private void updateWorkflowEntries(Workflow workflow) throws APIMgtDAOException {
+    private void updateWorkflowEntries(Workflow workflow) throws APIManagementException {
         workflow.setUpdatedTime(LocalDateTime.now());
-        getWorkflowDAO().updateWorkflowStatus(workflow);
-        //TODO stats stuff
+        try {
+            getWorkflowDAO().updateWorkflowStatus(workflow);
+            //TODO stats stuff
+        } catch (APIMgtDAOException e) {
+            String message = "Error while updating workflow entry";
```
prasa7: Shall we add some context information to the error log so that we can narrow down the error when we are debugging other than just mentioning  "Error while updating workflow entry". May be the blow sample code will be useful.

"Error while updating workflow entry for API:" + workflow.getApiName(); + ":" + workflow.getApiVersion();

chamilaadhi: workflow object will have 'workflowReference' which won't be much of a use since it is an internal id. But the exception will have the error message related to the exception so i think that would be enough to find the error

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIStoreImpl.java
chamilaadhi: done

PR #3882, opened 2017-03-27
Unit tests for AbstractAPIManager

tharikaGitHub:
This PR contains 
1. Unit tests for  carbon-apimgt/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/AbstractAPIManager.java class. 
2. Modifications to methods in AbstractAPIManager to prevent catching and throwing the same exception - APIMgtDAOException.


pubudu538:
As a practice, it would be better to add a small description in each and every PR which allows anyone to understand what this PR contains.

tharikaGitHub:
Thanks for the review

PR #3915, opened 2017-04-04
redirection support for workflows

chamilaadhi:
- redirection support for HttpWorkflowResponse response
- implementation for Store app creation/ subscription creation
- implementation for publisher lc state change

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/HttpWorkflowResponse.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/HttpWorkflowResponse.java
@@ -19,22 +19,34 @@ package org.wso2.carbon.apimgt.core.workflow;
 
 import org.json.simple.JSONObject;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * HTTP workflow response. Can be used to pass workflow complete redirections
  */
 public class HttpWorkflowResponse extends AbstractWorkflowResponse {
 
     private String redirectUrl = "";
-    private String redirectConfirmationMsg = null;
+    private String redirectConfirmationMsg = "";
     private JSONObject jsonPayloadObj = new JSONObject();
-    private JSONObject additionalParameters = new JSONObject();
+    private Map<String, String> additionalParameters = new HashMap<>();
 
     @Override
     @SuppressWarnings("unchecked")
     public String getJSONPayload() {
+        if (additionalParameters != null && !additionalParameters.isEmpty()) {
+            redirectUrl = redirectUrl.concat("?");
+            for (Map.Entry<String, String> entry : additionalParameters.entrySet()) {
+                redirectUrl = redirectUrl.concat(((entry.getKey().concat("=")).concat(entry.getValue())).concat("&"));
+            }
+            //remove tailing "&"
+            redirectUrl = redirectUrl.substring(0, redirectUrl.length() - 1);
+        }
+
         jsonPayloadObj.put("redirectUrl", redirectUrl);
         jsonPayloadObj.put("redirectConfirmationMsg", redirectConfirmationMsg);
-        jsonPayloadObj.put("additionalParameters", additionalParameters);
+
         return jsonPayloadObj.toJSONString();
     }
 
@@ -59,7 +71,7 @@ public class HttpWorkflowResponse extends AbstractWorkflowResponse {
         additionalParameters.put(paramName, paramValue);
     }
 
-    public JSONObject getAdditionalParameterss() {
+    public Map getAdditionalParameterss() {
         return additionalParameters;
-    }
+    }    
 }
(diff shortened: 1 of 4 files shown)


components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/HttpWorkflowResponse.java
```diff
@@ -19,22 +19,34 @@
...
 
     @Override
     @SuppressWarnings("unchecked")
     public String getJSONPayload() {
+        if (additionalParameters != null && !additionalParameters.isEmpty()) {
+            redirectUrl = redirectUrl.concat("?");
+            for (Map.Entry<String, String> entry : additionalParameters.entrySet()) {
+                redirectUrl = redirectUrl.concat(((entry.getKey().concat("=")).concat(entry.getValue())).concat("&"));
+            }
+            //remove tailing "&"
```
harsha89: anything to remove?

components/web/components/org.wso2.carbon.apimgt.publisher.commons.ui/src/main/fragments/api-details/public/js/details.js
```diff
@@ -306,17 +306,46 @@ function updateLifecycleHandler(event) {
...
-                text: message,
-                type: 'success',
-                dismissQueue: true,
-                progressBar: true,
-                timeout: 5000,
-                layout: 'topCenter',
-                theme: 'relax',
-                maxVisible: 10,
-            });
+            console.log(JSON.stringify(response))
```
pubudu538: Shall we avoid console logs?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/HttpWorkflowResponse.java
```diff
@@ -19,22 +19,34 @@
...
+            redirectUrl = redirectUrl.substring(0, redirectUrl.length() - 1);
+        }
+
         jsonPayloadObj.put("redirectUrl", redirectUrl);
         jsonPayloadObj.put("redirectConfirmationMsg", redirectConfirmationMsg);
```
Rajith90: Better to move these keys to constants.

components/web/components/org.wso2.carbon.apimgt.publisher.commons.ui/src/main/fragments/api-details/public/js/details.js
Rajith90: remove the console.log

components/web/apps/org.wso2.carbon.apimgt.store/src/main/public/js/apis/{apiId}/index.js
```diff
@@ -518,26 +518,54 @@ $(function () {
...
-                                    {addClass: 'btn btn-default', text: 'Stay on this page', onClick: function($noty) {
-                                        $noty.close();
-                                        location.href = contextPath + "/apis/" + apiId;
-                                      }
-                                    }
-                                  ]
-                            });
+                            var jsonPayload = subscription.workflowResponse.jsonPayload;
+                            
+                            if(jsonPayload != null && jsonPayload != "") {
```
mushthaq33: It is better to use null check as follows rather than using jsonPayload != null

Eg: if(!jsonPayload && jsonPayload != "")

components/web/apps/org.wso2.carbon.apimgt.store/src/main/public/js/applications/add.js
```diff
@@ -116,10 +116,30 @@ var addApplication = function () {
                                 }
                             ]
                         });
+                    } else if (jsonPayload != null && jsonPayload != "") {
```
mushthaq33: Better to use if(!jsonPayload) for null checks rather than using if(jsonPayload != null)

components/web/components/org.wso2.carbon.apimgt.publisher.commons.ui/src/main/fragments/api-details/public/js/details.js
```diff
@@ -306,17 +306,46 @@ function updateLifecycleHandler(event) {
...
+            var lcResponse = response.obj;
+            var jsonPayload = lcResponse.jsonPayload;
+                            
+            if(jsonPayload != null && jsonPayload != "") {
```
mushthaq33: Same here as well. Better to use if(!jsonPayload) for null checks rather than using if(jsonPayload != null)

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/HttpWorkflowResponse.java
chamilaadhi: removal is done in the next line. when url is constructed, '&' is set at the end from the previous loop. that one is removed from this line

components/web/apps/org.wso2.carbon.apimgt.store/src/main/public/js/apis/{apiId}/index.js
chamilaadhi: what would be the benefit of using this method instead of checking the null value? In my opinion it would be much easier to understand what i was doing within that conditional check when jsonPayload != null  is used (checking for null value.)

mushthaq33: Using the way I have mentioned is the best practise in javascript. Anyway if it is more understandable this way, It's fine

mushthaq33: And when you use if if(!jsonPayload) it checks for both null and empty strings. So you will not need to check for bother as I done here. if(jsonPayload != null && jsonPayload != "")

PR #4009, opened 2017-05-11
Change INDEXER column from VARCHAR to VARCHAR2.

jonheller1:
VARCHAR2 is the standard Oracle column type, not VARCHAR.  They both work the same, and it honestly doesn't make any technical difference which one you use.  But it's one of those things people like to point out during schema reviews so changing it can help the program look better.  This issue was raised by Tim O'Newton at http://stackoverflow.com/questions/43899695/


uvindra:
However your Oracle script seems to be out of date with the latest changes. I will add your fix in to the latest script

uvindra:
Your fix has been included in https://github.com/wso2/carbon-apimgt/pull/4032

PR #4066, opened 2017-05-30
Remove Blocked state and add Maintenance state to API lifecycle

dewmini:

[List all changes you want to add here. If you fixed an issue, please
add a reference to that issue as well.]

-

### When should this PR be merged

[Please describe any preconditions that need to be addressed before we
can merge this pull request.]

### Follow up actions

[List any possible follow-up actions here; for instance, testing data
migrations, software that we need to install on staging and production
environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIGatewayPublisherImpl.java
```diff
@@ -310,4 +310,25 @@ private void saveEndpointConfig(String gwHome, String content) {
...
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override public void changeAPIState(API api, String status) throws GatewayException {
+
+        if (gwHome == null) {
+            // build the message to send
```
harsha89: what is build the message to  send means?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/api/APIGateway.java
```diff
@@ -92,4 +92,13 @@
      */
     void deleteEndpoint(Endpoint endpoint) throws GatewayException;
 
+    /**
+     * Add API state change to gateway
+     *
+     * @param api
+     * @param status
+     * @throws GatewayException
+     */
```
praminda: param documentation and exception documentation is incomplete.
Please be more descriptive on method description so that others can understand what is the purpose of this method by looking  at the javadoc

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIGatewayPublisherImpl.java
```diff
@@ -310,4 +310,25 @@ private void saveEndpointConfig(String gwHome, String content) {
             }
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override public void changeAPIState(API api, String status) throws GatewayException {
+
```
praminda: Override annotation should be in a separate line and no need of extra line before first statement in the method

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/api/APIGateway.java
```diff
@@ -92,4 +92,13 @@
      */
     void deleteEndpoint(Endpoint endpoint) throws GatewayException;
 
+    /**
+     * Add API state change to gateway
+     *
+     * @param api
```
anuruddhal: Add parameter and exception descriptions

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/APIStateChangeWorkflow.java
```diff
@@ -226,6 +232,10 @@ private void updateAPIStatusForWorkflowComplete(String apiId, String status, Str
             String errorMsg = "Couldn't change the status of api ID " + apiId;
             log.error(errorMsg, e);
             throw new APIManagementException(errorMsg, e, ExceptionCodes.APIMGT_LIFECYCLE_EXCEPTION);
+        } catch (GatewayException e) {
+            String message = "Error occurred while changing the state of api ID " + apiId + " in gateway";
```
harsha89: it's better to include the status that he trying to change to log message as well

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/api/APIGateway.java
```diff
@@ -92,4 +92,13 @@
      */
     void deleteEndpoint(Endpoint endpoint) throws GatewayException;
 
+    /**
+     * Add API state change to gateway
```
harsha89: para meter desccriptions needed and better to change the description to more meaningful manner

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIGatewayPublisherImpl.java
```diff
@@ -310,4 +310,26 @@ private void saveEndpointConfig(String gwHome, String content) {
...
+            gatewayDTO.setLabels(api.getLabels());
+            APISummary apiSummary = new APISummary(api.getId());
+            apiSummary.setName(api.getName());
+            apiSummary.setVersion(api.getVersion());
+            apiSummary.setContext(api.getContext());
+            apiSummary.setLifeCycleState(status);
+            gatewayDTO.setApiSummary(apiSummary);
+            publishToPublisherTopic(gatewayDTO);
+        } else {
+            //TODO save to file system: need to consider editor mode scenario
```
Niranjan-K: There is a ToDo here? I have completed this scenario?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/util/APIMgtConstants.java
```diff
@@ -374,7 +374,7 @@
         public static final String API_CREATE = "API_CREATE";
         public static final String API_UPDATE = "API_UPDATE";
         public static final String API_DELETE = "API_DELETE";
-        public static final String API_PUBLISH = "API_PUBLISH";
+        public static final String API_STATE_CHANGE = "API_STATE_CHANGE";
```
Niranjan-K: This is a public variable and changing this will break the backward compatibility (Major version change). Are we exposing this class (export packages). If so, then we are breaking the version strategy,

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/APIStateChangeWorkflow.java
```diff
@@ -63,13 +65,15 @@
     private ApiDAO apiDAO;
     private APISubscriptionDAO apiSubscriptionDAO;
     private APILifecycleManager apiLifecycleManager;
+    private APIGateway apiGateway;
 
     public APIStateChangeWorkflow(ApiDAO apiDAO, APISubscriptionDAO apiSubscriptionDAO, WorkflowDAO workflowDAO,
-                                  APILifecycleManager apiLifecycleManager) {
+                                  APILifecycleManager apiLifecycleManager, APIGateway apiGateway) {
```
Niranjan-K: We are changing the Constructor signature. If this is a minor version release then we need to add a new constructor. For this change, the version release should be a major version change.

components/apimgt/org.wso2.carbon.apimgt.core/src/test/java/org/wso2/carbon/apimgt/core/dao/impl/ApiDAOImplIT.java
```diff
@@ -596,10 +596,10 @@ public void testGetAPIsByStatus() throws Exception {
             apiDAO.addAPI(api);
         }
 
-        List<API> blockedAPIsSummary = new ArrayList<>();
-        for (int i = 0; i < numberOfBlocked; ++i) {
-            API api = SampleTestObjectCreator.createUniqueAPI().lifeCycleStatus(blockedStatus).build();
-            blockedAPIsSummary.add(SampleTestObjectCreator.getSummaryFromAPI(api));
+        List<API> maintenanceAPIsSummary = new ArrayList<>();
+        for (int i = 0; i < numberOfInMaintenance; ++i) {
```
Niranjan-K: Here you have used for loop but I can't find the usage of "i".

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIGatewayPublisherImpl.java
dewmini: The else part should be written considering the editor mode. Since it is not completed yet, this part cannot be completed now.

Niranjan-K: @dewmini : Noted. In which case do we have a Git Issue for this so that this is tracked? It would be better to mention the Git issue in the ToDo as well. IMO this should be a L1 in the Git issue so that we fix this before releasing the feature.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/APIStateChangeWorkflow.java
dewmini: This is for API Manager 3.0.

Niranjan-K: @dewmini : But components need to follow Version strategy. So the version for component changing in the patch range is wrong.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/util/APIMgtConstants.java
dewmini: This is for API Manager 3.0 which is under development with C5. We haven't done a major release and this change comes under API Manager 3.0.0-M3. The development is still on going.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/workflow/APIStateChangeWorkflow.java
dewmini: API Manager 3.0 is under development with C5. We haven't done a major release yet and major development is still on going. Since these classes are still subject to major changes, we are continuously doing refactoring and heavy development. After major release we will adhere to the versioning.

components/apimgt/org.wso2.carbon.apimgt.core/src/test/java/org/wso2/carbon/apimgt/core/dao/impl/ApiDAOImplIT.java
dewmini: This loop is used to create a list of APIs which are in "Maintenance" state and the "numberOfInMaintenance" variable is configurable in the test case, so that we can create many lists of APIs in "Maintenance" state with different number of APIs. In that case "i" will not be used in anywhere.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIGatewayPublisherImpl.java
dewmini: We have decided to remove API lifecycle from editor mode. Hence else part has been removed by https://github.com/wso2/carbon-apimgt/commit/a0ed6c34bb257d7a09c04e66ff5a7b4e5d8229a1

components/apimgt/org.wso2.carbon.apimgt.core/src/test/java/org/wso2/carbon/apimgt/core/dao/impl/ApiDAOImplIT.java
Niranjan-K: In that case "for" loop should not be used. You can use some other iteration like "while" loop.

uvindra: I don't see any advantage of using a while loop over a for loop here. Everything that needs to be done in the for loop has to be done in the while loop as well(initialise counter,increment counter, checking if counter has exceeded condition)

In fact since the for loop allows you define all these on a single line it is much more readable and less error prone. 

So I would argue that using a for loop is much better than a while loop in this case as it improves code readability as well.

Niranjan-K: @uvindra : Noted. I went through the code and noticed in the SampleTestObjectCreator "createUniqueAPI", it creates the API. and here the method is used and random API are created using it by changing some of the values like API ID, context.
btw in this test class the numberOfInMaintenance is only set to 1. So basically we are using the loop to run once?

uvindra: @Niranjan-K Yes currently it is 1. But the point is we can choose to change the value of numberOfInMaintenance at a later point. Variables such as numberOfInMaintenance act as parameters for the test. So by writing the code in this way we can change the parameters that goverm the test and the code will still work. When you look at it in the context of a test case, this makes sense.

Niranjan-K: Yes, it will be useful if the limit value in the for loop is changed for different test, but in this case it is hard code as 1 in this test method. IMO when the requirement comes, we can add the loop else it is not required.

PR #4127, opened 2017-06-12
New methods for identity provider implementation and changes wrt API permissions

tharikaGitHub:

- A change with respect to API permissions feature
- New methods added to Identity provider implementation
- Tests for coverage of newly added methods

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
@@ -469,7 +469,7 @@ public class APIPublisherImpl extends AbstractAPIManager implements APIPublisher
                         (apiBuilder.getVersion())) && (originalAPI.getProvider().equals(apiBuilder.getProvider())) &&
                         originalAPI.getLifeCycleStatus().equalsIgnoreCase(apiBuilder.getLifeCycleStatus())) {
 
-                    if (apiBuilder.getPermission() != null && !("[]").equals(apiBuilder.getPermission())) {
+                    if (apiBuilder.getPermission() != null && !("").equals(apiBuilder.getPermission())) {
                         Map<String, Integer> roleNamePermissionList;
                         roleNamePermissionList = getAPIPermissionArray(apiBuilder.getPermission());
                         if (checkRoleValidityForAPIPermissions(roleNamePermissionList)) {
--- a/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
@@ -337,7 +337,7 @@ public class APIPublisherImpl extends AbstractAPIManager implements APIPublisher
                 if (StringUtils.isEmpty(apiBuilder.getApiDefinition())) {
                     apiBuilder.apiDefinition(apiDefinitionFromSwagger20.generateSwaggerFromResources(apiBuilder));
                 }
-                if (apiBuilder.getPermission() != null && !("").equals(apiBuilder.getPermission())) {
+                if (apiBuilder.getPermission() != null && !StringUtils.isEmpty(apiBuilder.getPermission())) {
                     Map<String, Integer> roleNamePermissionList;
                     roleNamePermissionList = getAPIPermissionArray(apiBuilder.getPermission());
                     apiBuilder.permissionMap(roleNamePermissionList);
@@ -469,7 +469,7 @@ public class APIPublisherImpl extends AbstractAPIManager implements APIPublisher
                         (apiBuilder.getVersion())) && (originalAPI.getProvider().equals(apiBuilder.getProvider())) &&
                         originalAPI.getLifeCycleStatus().equalsIgnoreCase(apiBuilder.getLifeCycleStatus())) {
 
-                    if (apiBuilder.getPermission() != null && !("").equals(apiBuilder.getPermission())) {
+                    if (apiBuilder.getPermission() != null && !StringUtils.isEmpty(apiBuilder.getPermission())) {
                         Map<String, Integer> roleNamePermissionList;
                         roleNamePermissionList = getAPIPermissionArray(apiBuilder.getPermission());
                         if (checkRoleValidityForAPIPermissions(roleNamePermissionList)) {
--- a/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
@@ -337,7 +337,7 @@ public class APIPublisherImpl extends AbstractAPIManager implements APIPublisher
                 if (StringUtils.isEmpty(apiBuilder.getApiDefinition())) {
                     apiBuilder.apiDefinition(apiDefinitionFromSwagger20.generateSwaggerFromResources(apiBuilder));
                 }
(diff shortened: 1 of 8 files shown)


components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
```diff
@@ -87,6 +91,54 @@ public boolean isValidRole(String roleName) {
...
+    @Override
+    public String getRoleId(String roleName) throws IdentityProviderException {
+        Response role = scimServiceStub.searchGroups(FILTER_PREFIX + roleName);
+        String roleId = null;
+        if (role.status() == 200) {
+            String responseBody = role.body().toString();
+            JsonParser parser = new JsonParser();
+            JsonObject parsedResponseBody = (JsonObject) parser.parse(responseBody);
+            JsonArray roleList = (JsonArray) parsedResponseBody.get("Resources");
+            if (roleList.size() == 1) {
```
harsha89: why we check the size with only 1? Can there be more than one?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
```diff
@@ -87,6 +91,54 @@ public boolean isValidRole(String roleName) {
...
+        }
+        return roleIds;
+    }
+
+    @Override
+    public String getRoleId(String roleName) throws IdentityProviderException {
+        Response role = scimServiceStub.searchGroups(FILTER_PREFIX + roleName);
+        String roleId = null;
+        if (role.status() == 200) {
+            String responseBody = role.body().toString();
```
harsha89: can body be null?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
```diff
@@ -469,7 +469,7 @@ public void updateAPI(API.APIBuilder apiBuilder) throws APIManagementException {
                         (apiBuilder.getVersion())) && (originalAPI.getProvider().equals(apiBuilder.getProvider())) &&
                         originalAPI.getLifeCycleStatus().equalsIgnoreCase(apiBuilder.getLifeCycleStatus())) {
 
-                    if (apiBuilder.getPermission() != null && !("[]").equals(apiBuilder.getPermission())) {
+                    if (apiBuilder.getPermission() != null && !("").equals(apiBuilder.getPermission())) {
```
harsha89: cant we use StringUtils.IsEmpty?

tharikaGitHub: Yes, we can change it to use StringUtils.IsEmpty. I'll fix it.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
tharikaGitHub: When the status is 200, it means there is a role corresponding to the searched role name and the response returns a role. So it wont be null. Do you recommend to check for a null body within the status == 200 condition?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
tharikaGitHub: Usually in SCIM, it doesn't allow a role with the same display name be created. We assume that role name is unique. Hence usually a single role will be returned. But precaution is taken to check whether the list actually contains a single role. Else a null roleId will be returned. That's why we check for one.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
harsha89: ah okey. So if the search role isn't there. It will return with some other status code right? Then it should be fine. Do we require an else statement to handle other status codes?

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
harsha89: Got it :+1:

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/DefaultIdentityProviderImpl.java
tharikaGitHub: Yes, If the role is not there, it'll return a 404 response.  This method is for getting the role ID of a role if it exists. So there will not be a requirement to handle other status codes within this method AFAIU.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
```diff
@@ -337,7 +337,7 @@ public String addAPI(API.APIBuilder apiBuilder) throws APIManagementException {
                 if (StringUtils.isEmpty(apiBuilder.getApiDefinition())) {
                     apiBuilder.apiDefinition(apiDefinitionFromSwagger20.generateSwaggerFromResources(apiBuilder));
                 }
-                if (apiBuilder.getPermission() != null && !("").equals(apiBuilder.getPermission())) {
+                if (apiBuilder.getPermission() != null && !StringUtils.isEmpty(apiBuilder.getPermission())) {
```
harsha89: string utils check both null case and empty case. We can remove the null check :+1:

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
```diff
@@ -469,7 +469,7 @@ public void updateAPI(API.APIBuilder apiBuilder) throws APIManagementException {
                         (apiBuilder.getVersion())) && (originalAPI.getProvider().equals(apiBuilder.getProvider())) &&
                         originalAPI.getLifeCycleStatus().equalsIgnoreCase(apiBuilder.getLifeCycleStatus())) {
 
-                    if (apiBuilder.getPermission() != null && !("[]").equals(apiBuilder.getPermission())) {
+                    if (apiBuilder.getPermission() != null && !StringUtils.isEmpty(apiBuilder.getPermission())) {
```
harsha89: string utils check both null case and empty case. We can remove the null check :+1:

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
tharikaGitHub: Ah yes, will remove it.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/impl/APIPublisherImpl.java
tharikaGitHub: I'll remove it.

PR #4407, opened 2017-08-14
Adding workflow related apis to admin api

chamilaadhi:

Moving workflow related apis to admin apis. This is done as a part of the improvement done to introduce simple approval process to admin portal for workflows

- Moving workflow approval related api to admin api 
- Add additional apis to list pending workflows

### When should this PR be merged

[Please describe any preconditions that need to be addressed before we
can merge this pull request.]

### Follow up actions

[List any possible follow-up actions here; for instance, testing data
migrations, software that we need to install on staging and production
environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/dao/impl/WorkflowDAOImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/dao/impl/WorkflowDAOImpl.java
@@ -34,6 +34,8 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Default implementation of the WorkflowDAO interface. Uses SQL syntax that is common to H2 and MySQL DBs.
@@ -52,7 +54,7 @@ public class WorkflowDAOImpl implements WorkflowDAO {
     public void addWorkflowEntry(Workflow workflow) throws APIMgtDAOException {
 
         final String query = " INSERT INTO AM_WORKFLOWS (WF_REFERENCE,WF_TYPE,WF_STATUS,WF_CREATED_TIME,"
-                + "WF_EXTERNAL_REFERENCE, WF_ATTRIBUTES) VALUES (?,?,?,?,?,?)";
+                + "WF_EXTERNAL_REFERENCE, WF_ATTRIBUTES, WF_STATUS_DESC) VALUES (?,?,?,?,?,?,?)";
 
         try (Connection connection = DAOUtil.getConnection()) {
             connection.setAutoCommit(false);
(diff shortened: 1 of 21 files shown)


components/apimgt/org.wso2.carbon.apimgt.rest.api.admin/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/impl/WorkflowsApiServiceImpl.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public Response workflowsWorkflowReferenceIdGet(String workflowReferenceId, Request request)
+            throws NotFoundException {
+        try {
+            APIMgtAdminService apiMgtAdminService = RestApiUtil.getAPIMgtAdminService();
+            Workflow workflow = apiMgtAdminService.retrieveWorkflow(workflowReferenceId);
+            WorkflowDTO workflowDTO = WorkflowMappingUtil.toWorkflowDTO(workflow);
+            return Response.ok().entity(workflowDTO).build();
```
isurulucky: Need to handle the 404 scenario. The best option is to handle it at the DAO layer itself by throwing a DAO Exception appropriately.

chamilaadhi: Thanks for pointing the missing 404 code. I think throwing an exception if an entry is not in the database may not be a suitable approach. I think sending an empty object (in this case a null value would be much better). I added a check in the rest api layer to send a 404 based on the availability of this object

isurulucky: IMHO throwing the exception from the DAO layer itself is the suitable approach; need do it with the correct exception code and message (ex.: workflow information not found). Then we do not need to add any if checks again at the Service layer/Rest API layer to see if the returned object is null, etc. since this will be automatically handled from the rest API exception handlers. This will reduce the number of branches in the code.

components/apimgt/org.wso2.carbon.apimgt.core/src/main/java/org/wso2/carbon/apimgt/core/dao/impl/WorkflowDAOImpl.java
```diff
@@ -127,12 +130,16 @@ public void updateWorkflowStatus(Workflow workflow) throws APIMgtDAOException {
...
-        Workflow workflow;
+        Workflow workflow = null;
         try (Connection conn = DAOUtil.getConnection();
                 PreparedStatement ps = conn.prepareStatement(getworkflowQuery)) {
             ps.setString(1, workflowReference);
             try (ResultSet rs = ps.executeQuery()) {
-                workflow = this.createWorkflowFromResultSet(rs);
+                List<Workflow> list = this.createWorkflowFromResultSet(rs);
+                if (!list.isEmpty()) {
+                    workflow = list.get(0);
```
chamilaadhi: For a given referenceId there is either one entry or no entry. That is why taking the first value. retrieveWorkflow(String workflowReference) method which calls createWorkflowFromResultSet() method passes resultset which only has a single entry.  I have re-used the generic createWorkflowFromResultSet() method which is used to generate list of workflow objects using the resultset which has more than one entry.

components/apimgt/org.wso2.carbon.apimgt.rest.api.admin/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/impl/WorkflowsApiServiceImpl.java
chamilaadhi: updated the code with the suggested method and added a test case to cover the scenario.

PR #4838, opened 2017-12-18
Adding admin app to web-ui server

malinthaprasan:

- Adding admin app to new web-ui server.


malinthaprasan:
Thanks @tmkasun for the fix

PR #4940, opened 2018-01-12
Fix https://github.com/wso2/product-apim/issues/2366

Krishanx92:

Add new delete option to API overview
Fix https://github.com/wso2/product-apim/issues/2366


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Overview.js
```diff
@@ -68,6 +71,14 @@ class Overview extends Component {
         );
     }
 
+    handleRequestClose = () => {
```
tmkasun: shall we use the ES6 style class methods instead of class property style methods, because this is still in proposal stage 3 and could be changed in ES7

PR #4955, opened 2018-01-15
Change defaultTheme.js src path

Krishanx92:
Source path change to /public/app/theme/defaultTheme.js both publisher and store.


Krishanx92:
https://github.com/wso2/product-apim/issues/2396

PR #5095, opened 2018-02-22
[C5] get defined http method headers from swagger content to handle CORS requests

RenukaFernando:
- Remove hard coded `ACCESS_CONTROL_ALLOW_METHODS_HEADER` CORS preflighted header.
- Read defined HTTP headers from swagger definition and response only with readed headers.
- fix https://github.com/wso2/product-apim/issues/2792

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.commons/src/main/java/org/wso2/carbon/apimgt/rest/api/common/util/RestApiUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.commons/src/main/java/org/wso2/carbon/apimgt/rest/api/common/util/RestApiUtil.java
@@ -15,6 +15,9 @@
 */
 package org.wso2.carbon.apimgt.rest.api.common.util;
 
+import io.swagger.models.Path;
+import io.swagger.models.Swagger;
+import io.swagger.parser.SwaggerParser;
 import org.apache.commons.io.IOUtils;
 import org.json.simple.JSONObject;
 import org.slf4j.Logger;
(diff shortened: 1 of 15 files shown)


components/apimgt/org.wso2.carbon.apimgt.rest.api.commons/src/main/java/org/wso2/carbon/apimgt/rest/api/common/util/RestApiUtil.java
```diff
@@ -526,4 +534,68 @@ public static ZoneId getRequestTimeZone(String timestamp) {
...
 
+    /**
+     * Get defined HTTP methods in the swagger definition as space separated string
+     *
+     * @param request           Request
+     * @param serviceMethodInfo Method informations for the request
+     * @return Http Methods as a space separated string
+     * @throws APIManagementException if failed to get defined http methods
+     */
+    public static synchronized String getDefinedMethodHeadersInSwaggerContent(
```
malinthaprasan: Do we need to make this synchronized?

RenukaFernando: Thank you. Resolved with a static block. Parsing swagger definition is now done when the class is loaded.
fix: https://github.com/wso2/carbon-apimgt/pull/5095/commits/ac2a2a0304d573e35d3a8105350692f0beb36118

components/apimgt/org.wso2.carbon.apimgt.rest.api.commons/src/main/java/org/wso2/carbon/apimgt/rest/api/common/util/RestApiUtil.java
```diff
@@ -526,4 +549,53 @@ public static ZoneId getRequestTimeZone(String timestamp) {
...
+
+        if (requestURI.contains("/publisher")) {
+            swagger = swaggerRestAPIDefinitions.get(RestApiConstants.APPType.PUBLISHER);
+        } else if (requestURI.contains("/store")) {
+            swagger = swaggerRestAPIDefinitions.get(RestApiConstants.APPType.STORE);
+        } else if (requestURI.contains("/analytics")) {
+            swagger = swaggerRestAPIDefinitions.get(RestApiConstants.APPType.ANALYTICS);
+        } else if (requestURI.contains("/admin")) {
+            swagger = swaggerRestAPIDefinitions.get(RestApiConstants.APPType.ADMIN);
+        }
```
malinthaprasan: Shall we also have an else condition to skip this check with returning null? Otherwise, any other REST API getting an OPTIONS call is likely to break. The caller of this has to do a null check before responding with CORS headers. WDYT?

RenukaFernando: Thanks. Fixed with https://github.com/wso2/carbon-apimgt/pull/5095/commits/a4555fc63ff0b302cd029e49bc6910318cd21125
Shall we return an empty string? "No allowed methods".

PR #5134, opened 2018-02-28
fixing https://github.com/wso2/product-apim/issues/2815

tharindu1st:
fixing Search by exact value


tharindu1st:
fixing #https://github.com/wso2/product-apim/issues/2815

PR #5437, opened 2018-06-11
Fixing #2162

mushthaq33:
- Fixing #2162


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher/src/main/resources/publisher-api.yaml
```diff
@@ -2943,6 +2943,7 @@ definitions:
       - endpointConfig
       - visibility
       - type
+      - apiDefinition
```
Rajith90: Don't we have to generate the rest api after this change

mushthaq33: The rest API was generated but there is no change

PR #5504, opened 2018-06-18
Fix label attach issue in IE 11

Krishanx92:

Fix label attach issue in IE 11


Krishanx92:
Fix https://github.com/wso2/product-apim/issues/3453

PR #5534, opened 2018-07-05
Set exp in JWT in seconds 

isurulucky:

- Set the expiry time in JWT (exp) in seconds
- Handle the scenarios where validity period it not specified and specified as -1 (non-expiring token)


chamindias:
Fix for : https://github.com/wso2/product-apim/issues/3527

PR #5582, opened 2018-08-13
Migrate Material-UI beta version to stable 1.0.x

tmkasun:

- Upgrade Material-UI version to [latest stable](https://github.com/mui-org/material-ui/releases/tag/v1.5.0)

### Follow up actions

- Check for `Element type is invalid` errors and fix the imports accordingly

```Javascript
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

- To prevent manual testing for UI breaks , Integrate [jestjs.io](https://jestjs.io/docs/en/tutorial-react) and  [enzyme](http://airbnb.io/enzyme/)


Mariangela:
@chanaka3d , @tmkasun , please mark if this has a doc impact or not by selecting the appropriate label (Docs/Has Impact OR Docs/No Impact).

PR #5754, opened 2018-09-26
Fixing #3773-Handling POST request without Content-Type for Token API

rmsamitha:
## Purpose
To fix handling POST request without Content-Type for Token API to support backward compatibility of the product eventhough the Content-Type header is a required header as per the HTTP specs. (Issue #3773)

## Approach
Fix is done in SynapsePropertiesHandler which is used in Token API. So handler will check whether the Content-Type header is available in the request message context, and then if it is not available, and if the request is a POST or PUT type, it will add the Content-Type header with the value "application/x-www-form-urlencoded".

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/common/SynapsePropertiesHandler.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/common/SynapsePropertiesHandler.java
@@ -15,9 +15,13 @@
  */
 package org.wso2.carbon.apimgt.gateway.handlers.common;
 
+import org.apache.axis2.Constants;
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.rest.AbstractHandler;
 import org.wso2.carbon.apimgt.impl.APIConstants;
+
+import java.util.Map;
 
 public class SynapsePropertiesHandler extends AbstractHandler{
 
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/common/SynapsePropertiesHandler.java
```diff
@@ -30,6 +34,23 @@ public boolean handleRequest(MessageContext messageContext) {
...
+
+        String httpMethod = (String) ((Axis2MessageContext) messageContext).getAxis2MessageContext().
+                getProperty(Constants.Configuration.HTTP_METHOD);
+        org.apache.axis2.context.MessageContext axis2MsgContext =
+                ((Axis2MessageContext) messageContext).getAxis2MessageContext();
+        Map headers =
+                (Map) (axis2MsgContext).getProperty(org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS);
+        boolean isContentTypeNotSet = false;
+        if (headers != null) {
+            isContentTypeNotSet = headers.get("Content-Type") == null || headers.get("Content-Type").equals("");
```
ruks: Can we use StringUtils.isBlank() if possible?

rmsamitha: StringUtils.isBlank(..) or isEmpty(..) methods expect String arguments and here headers Map's get(..)  returns an Object. So if we want to use before mentioned methods, we will have to cast the get(..) return values to String and that is not a meaningful usage.

ruks: noted.

rmsamitha:
This fixes https://github.com/wso2/product-apim/issues/3773

PR #5841, opened 2018-10-29
Fixing #3899

mushthaq33:
## Purpose
Fixing #3899


gusto2:
The pull is fixing methods getApplicarionById and getApplicationByUUID, but not getApplicationByClientId, I will add a fix

PR #5856, opened 2018-11-10
avoid NullPointerExceptions

deki:
## Purpose
While using the API of API Manager 2.6.0, I got several NPEs as no proper validation was present.  

## Goals
Added null checks and additional exception messages.


bhathiya:
Thanks for your contribution @deki.

PR #5885, opened 2018-12-04
Fix jwt x5t

gusto2:
## Purpose
Fixing wso2#5884

## Goals
Generate validable JWT backend token

## Approach
Encodings of all JWT elements were done using basic Base64 (java.util.Base64) with URL encoding and without wrapping. That made the generated token and signature verfiable by 3rd party tools (usable by backend clients)

## Automation tests
Unfortunate none

## Security checks
"x5t" jwt header is considered obsolete (using sha-1), we shold use "x5t#256" or make it configurable

## Test environment
oracle jdk 1.8_181, Amazon Linux2 (CentOS)


gusto2:
Oh, wrong pull target, please disregard

PR #5970, opened 2019-02-06
fix for https://github.com/wso2/product-apim/issues/4302

isharac:
## Automation tests
 - Unit tests 
   > org.wso2.carbon.apimgt.impl.SAMLGroupIDExtractorImplTest


isharac:
DOC Impact: We can configure DefaultGroupExtractorClaimUri in api-manager.xml to use as the group extractor claim.

Issue #5985, opened 2019-02-14
Add Markdown support when adding docs in the publisher


Varuni-Punchihewa:
**Description:**
Currently, when you are adding a new document to the API in the publisher, you can only give Inline, URL and File as the source type. It doesn't have a mechanism to add markdown (md) docs. 

**Suggested Labels:**
Type/Improvement

**Affected Product Version:**
APIM 2.x   

**Suggested Approach:**
- Introduce a new source type as "Markdown" in the Docs tab of the Publisher
- Once the user selects the Markdown radio button, adds a new doc, and clicks the Edit Content button, it redirects the user to a Markdown editor.
- Use a library to convert the markdown to markup language when viewing it on the Store

npamudika:
Closing as this is available in APIM 3.0.0 now.

PR #6000, opened 2019-02-28
Add Markdown support for the APIM publisher

Varuni-Punchihewa:
## Purpose
Enable the user to add documents written in the markdown language 

## Approach
 - When the User goes to the API Publisher, selects an API, goes to the Docs tab, and press Add New Document. it would display "Markdown" option under the "Source" label
-  Once the user adds a Markdown document and clicks "Edit Content", the user will be directed into a new page where user is provided with a markdown text editor and preview

**For more details refer #5985**


Varuni-Punchihewa:
> I don't think we need to angular to support this "Showdown js" library. Can you please try to remove angular imports and see. Check this fiddle also.
> 
> https://jsfiddle.net/jacobjoaquin/em8ce73a/

Varuni-Punchihewa:
@chanaka3d but the editor uses the angular libraries
eg: var app = angular.module('showdown.editor', ['ng-showdown', 'pageslide-directive', 'ngAnimate', 'ngRoute', 'ngCookies', 'ngSanitize']);
  app.controller('editorCtrl', ['$scope', '$showdown', '$http', '$cookies', '$sanitize', function ($scope, $showdown, $http, $cookies, $sanitize) {
...
} as in editor.js file
So I think we need the angular libraries

PR #6073, opened 2019-03-28
Making eslint compatible

chanaka3d:
## Purpose
Restructure the code to make it eslint compatible and rename all files from js -> jsx. Deleted unused files.


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store/.eslintrc.js
```diff
@@ -43,17 +44,15 @@ module.exports = {
...
-            'error',
-            {
-                allowAfterThis: true,
-            },
-        ],
+        'no-underscore-dangle': 0,
         'no-restricted-syntax': ['off'],
+        'no-restricted-globals': ["off"],
         'no-plusplus': ['off'],
+        "no-param-reassign": 0,
```
tmkasun: I think this rule prevents following action, and IMHO it's better to keep this rule enabled? 

function (param) {
param = "My Val";
}

chanaka3d: Yes.. it prevents to do foo.bar = someothervar.param;

PR #6089, opened 2019-04-02
Fixing subscription related issues

chanaka3d:
## Purpose
API Details subscription listing page subscription delete feature and link to application was missing. This PR is fixing that.

The KeyType was maintained in the child components state which is an anti pattern causing unexpected behavior when generating tokens. This PR is fixing that as well.

React Context was passing a call back method to update the child components, but it was not been called from the child components after creating new apps and doing subscriptions etc. This PR is fixing that.


tmkasun:
@chanaka3d  , Can we close this PR now ?

PR #6130, opened 2019-04-10
fix for JWT generation fails when the username itself contains the te…

isharac:
This PR provides fix for https://github.com/wso2/product-apim/issues/4488


bhathiya:
@isharac please send the same to the master as well.

PR #6137, opened 2019-04-11
Revert "Fix invalid token cache initialization issue"

fazlan-nazeem:
Reverts wso2/carbon-apimgt#6107


bhathiya:
@ChamodDamitha Please check how we can fix this properly.

PR #6219, opened 2019-05-03
Fixes product-apim#4429

fazlan-nazeem:
- fixes wso2/product-apim#4429 

(for non-oauth APIs there will be no **subscription** tier. The existing logic was disabling the LC publish button based on tier availability. This fix does that check only for oauth APIs.)

- Changes default key validation client to WSClient from Thrift (separate PR will be sent to remove thrift dependencies and configs related to this)


bhathiya:
@fazlan-nazeem please make this change in all config files in integration tests too.

PR #6272, opened 2019-05-08
add endpointconfig support and improve scopes support

isharac:
add scopes support in create api REST API


dushansilva:
https://github.com/wso2/product-apim/issues/4736

PR #6305, opened 2019-05-14
Fixed JWT formatter issue when adding custom application attributes

HiranyaKavishani:
## Purpose
When adding custom attributes to an application, the JWT generator retrieves the JSON object values.

Fixed https://github.com/wso2-support/carbon-apimgt/pull/1450/files

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/token/AbstractJWTGenerator.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/token/AbstractJWTGenerator.java
@@ -18,6 +18,7 @@
 
 package org.wso2.carbon.apimgt.keymgt.token;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jwt.JWTClaimsSet;
 import org.apache.axiom.util.base64.Base64Utils;
 import org.apache.commons.codec.binary.Base64;
@@ -46,6 +47,7 @@ import org.wso2.carbon.user.core.UserStoreManager;
 import org.wso2.carbon.user.core.service.RealmService;
 import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.security.*;
 import java.security.cert.Certificate;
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/token/AbstractJWTGenerator.java
```diff
@@ -216,7 +218,18 @@ public String buildBody(TokenValidationContext validationContext) throws APIMana
...
+                        try {
+                            Map<String, String> map = mapper.readValue(claimVal, Map.class);
+                            jwtClaimsSetBuilder.claim(claimURI, map);
+                        } catch (IOException e) {
+                            // Exception isn't thrown in order to generate jwt without claim, even if an error is
+                            // occurred during the retrieving claims.
+                            log.error("Error while reading claim values", e);
+                        }
+                    } else if (userAttributeSeparator != null && claimVal != null &&
+                            claimVal.contains(userAttributeSeparator)) {
```
TanyaM: claimVal != null can be moved out to a if block outside since its checked in both if and else if

HiranyaKavishani: Yes possible. claimVal != null has been basically checked to avoid NPE comes with claimVal.contains function

TanyaM: Suggestion is not to remove it, but to move it to an outer block since its evaluated twice.

components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/token/AbstractJWTGenerator.java
```diff
@@ -216,7 +218,18 @@ public String buildBody(TokenValidationContext validationContext) throws APIMana
...
-                    if (userAttributeSeparator != null && claimVal != null && claimVal.contains(userAttributeSeparator)) {
+                    if (claimVal != null && claimVal.contains("{")) {
+                        ObjectMapper mapper = new ObjectMapper();
+                        try {
+                            Map<String, String> map = mapper.readValue(claimVal, Map.class);
+                            jwtClaimsSetBuilder.claim(claimURI, map);
+                        } catch (IOException e) {
+                            // Exception isn't thrown in order to generate jwt without claim, even if an error is
+                            // occurred during the retrieving claims.
+                            log.error("Error while reading claim values", e);
```
sanjeewa-malalgoda: We should add some more details to this error message, what claim value, URI failed, what user etc if possible.

HiranyaKavishani: Fixed with 4158a39f601c8feed0db799983f4f5bdf49a1f5d

components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/token/AbstractJWTGenerator.java
HiranyaKavishani: Yes. Got it. But if do so, this can affect other conditions.
Ex: "exp".equals(claimURI)" This time, it doesn't care about claimVal. It can be either null or not null.

PR #6362, opened 2019-05-21
add rest apis for application key functions and improve code quality

dushansilva:
Add rest apis to (https://github.com/wso2/product-apim/issues/4758)

- Update grant types and callback url of an application 
- Get key details of a given type
- Re-generate consumer secret

Add comments for several rest api resources
Improve error handling in UI


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Create/NewApp.jsx
```diff
@@ -106,8 +106,11 @@ class NewApp extends React.Component {
                 this.props.updateApps();
             })
             .catch((error) => {
-                if (error.message === 'Conflict') {
-                    Alert.error('Application already exists');
+                const { response } = error;
+                if (response && response.body) {
+                    const message = response.body.description ? response.body.description
```
tmkasun: I think we could simplify this as 

```javascript
const message = response.body.description || 'Error while creating the application';
```

dushansilva: done

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/package-lock.json
```diff
@@ -3087,9 +3087,9 @@
           "integrity": "sha1-hqdj9Y7k18L2sQLkdkBQ3n7ZDGs=",
           "dev": true,
           "requires": {
-            "regenerate": "^1.2.1",
-            "regjsgen": "^0.2.0",
-            "regjsparser": "^0.1.4"
+            "regenerate": "1.4.0",
```
TanyaM: IMO committing package-lock.json is wrong if there are no changes going to package.json file. For example say we have ^1.2.1 for given module in package.json. So it can take any 1.x.x version. But if a certain functionality is written to work with 1.2.1 version and if 1.4.0 get locked in the packge-lock.json after a npm install, it might break the functionality.
Open for discussion :)

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Create/NewApp.jsx
```diff
@@ -106,8 +106,10 @@ class NewApp extends React.Component {
                 this.props.updateApps();
             })
             .catch((error) => {
-                if (error.message === 'Conflict') {
-                    Alert.error('Application already exists');
+                const { response } = error;
+                if (response && response.body) {
+                    const message = response.body.description || 'Error while creating the application';
```
TanyaM: Isn't it better to have application name in the error message?

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/package-lock.json
dushansilva: According to what I have read online it is ok to commit the package-lock.json. We should actually lock the module versions in package.json file (by removing ^). I think this would be a better way to prevent from breaking functionality from npm install

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Create/NewApp.jsx
tmkasun: I think the `description` have it

dushansilva: message returned from description will include the application name as that is the error message returned from the backend. but however seems like name is not in the state of this component.

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/package-lock.json
tmkasun: @dushansilva ,

I think we should not remove the carets from the package.json file, which will make it implicitly a lock file. As I found out, NPM has released a new command to do the installation and update the version locks [npm ci](https://docs.npmjs.com/cli/ci.html) [This blog post](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable).
After searching about this issue , I too agree with the following comment :smile: 

> What's the fucking use of a lock file if you are just going to update it anyways because a package.json dependency can be updated to a newer version?
> 
> And please don't tell me I am suppose to remove the carets from my package.json otherwise I might as well consider that to be our new lock file.

https://github.com/npm/npm/issues/18103#issuecomment-370401935

source: [SO](https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json)

Until we figure this out shall we not to commit the package-lock.json file ?
I too agree with @taniamahanama  comment that, updating package versions unknowingly could lead to wired issues.

@taniamahanama , Thanks for bringing this up.
Shall we create a separate git  issue for this in product-apim and move this discussion there, and close this PR ?

dushansilva: +1 I will remove the package-lock.json file

TanyaM: > According to what I have read online it is ok to commit the package-lock.json. We should actually lock the module versions in package.json file (by removing ^). I think this would be a better way to prevent from breaking functionality from npm install
> 
> https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

@dushansilva : There are scenarios that we should commit package-lock.json. For example if you do a version bump to a particular module, then the package-lock.json will be automatically locked with the newer version. So my concern was about committing it when there are no changes to the package.json file.

PR #6501, opened 2019-06-13
Update .travis.yml

tmkasun:
Change travis cache directories


tharindu1st:
make jdk8 as jdk version

tmkasun:
Hi @tharindu1st 
> make jdk8 as jdk version

It is the pre-installed version in Travis `trusty` instances
We have an issue with updating to latest Travis instance (Xenial) because it does not support Oracle JDK 8 (Because it has reached EOL)
https://docs.travis-ci.com/user/reference/trusty/

PR #6502, opened 2019-06-13
fix issue in updating keys

dushansilva:
- This will allow to update application keys to update its grant type and callback url after its been generated.
- This will remove the token type selection option from key generation UI
- This will fix key generation issue in through api using wizard

https://github.com/wso2/product-apim/issues/5040


dushansilva:
don't merge this yet

dushansilva:
can merge now, all changes done

PR #6511, opened 2019-06-16
Old applications validate with default application validator

HiranyaKavishani:


malinthaprasan:
We need to do a proper fix without hardcoding "Role based scope validator"

malinthaprasan:
https://github.com/wso2/product-apim/issues/5015

PR #6575, opened 2019-06-22
Revert "Fix annotation"

tharindu1st:
Reverts wso2/carbon-apimgt#6530


tharindu1st:
This is reverting due to having new osgi services

Issue #4980, opened 2018-01-19
Build fail on Windows [APIM 2.2.0][6.x]


vlgunarathne:
**Description:**
The following error occurs when the 6.x branch is being built on a windows machine.

Failed tests:
  TenantManagerHostObjectTest.testGetStoreTenantThemesPath:23 expected:<repository[\deployment\server\jaggeryapps\store\site\tenant_themes\]> but was:<repository[/deployment/server/jaggeryapps/store/site/tenant_themes/]>

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environments details and versions:**    

**Steps to reproduce:**

**Related Issues:**

vlgunarathne:
This issue was fixed with the following PR:
https://github.com/wso2/carbon-apimgt/pull/4983

PR #6584, opened 2019-06-25
Fixing #4980

mushthaq33:
Fixes https://github.com/wso2/product-apim/issues/4980

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/impl/ApisApiServiceImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/impl/ApisApiServiceImpl.java
@@ -583,10 +583,21 @@ public class ApisApiServiceImpl extends ApisApiService {
             String tenantDomain = RestApiUtil.getLoggedInUserTenantDomain();
             apiIdentifier = APIMappingUtil.getAPIIdentifierFromApiIdOrUUID(apiId,
                     tenantDomain);
+            API api = APIMappingUtil.getAPIFromApiIdOrUUID(apiId, tenantDomain);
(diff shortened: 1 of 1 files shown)


pubudu538:
Fixing #4980 points to the 4980 issue in the carbon-apimgt. We should point to the product apim issue 4980.

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/impl/ApisApiServiceImpl.java
```diff
@@ -1883,4 +1897,41 @@ private boolean isAuthorizationFailure(Exception e) {
...
+     * @param mediation
+     * @return if the API is modified or not
+     */
+    private boolean isAPIModified(API api, Mediation mediation) {
+        if (mediation != null) {
+            String sequenceName;
+            if (APIConstants.API_CUSTOM_SEQUENCE_TYPE_IN.equalsIgnoreCase(mediation.getType())) {
+                sequenceName = api.getInSequence();
+                if (isSequenceExistsInAPI(sequenceName, mediation)) {
+                    api.setInSequence(null);
```
Rajith90: Why do we set null as the mediation and do an api update

mushthaq33: Is there any other way to remove the sequence which is already set? That's how we have done in other places as well I guess

rmsamitha: The same value of inSequence attribute of api object when an In sequence has not been set to an API when the API is created, should be set here too.

PR #6563, opened 2019-06-21
Fix wso2/product-apim#4895 #5098 #5023

dushaniw:
**Fixes https://github.com/wso2/product-apim/issues/4895**

This PR adds validation for application, API, and subscription policy add via Admin Portal UI. The REST API validation is already there. This PR validates and display an informative error message in the UI as follows.

image

image

image

**Fixes https://github.com/wso2/product-apim/issues/5098**
The PR adds correlation ID to the metaClientType field of even streams. So the request stream would be like follows.

`[2019-05-22 20:33:26,597]  INFO {org.wso2.siddhi.core.stream.output.sink.LogSink} - org.wso2.apimgt.statistics.request: : Event{timestamp=1558537406590, data=[{"correlationID":"cfdc9f37-a302-4a93-8e61-7356966aebfb","keyType":"PRODUCTION"}, MBO3Uka7VAeZgvw0dtfP4hwx4Hga, DefaultApplication, 1, admin, /pizzashack/1.0.0, PizzaShackAPI, 1.0.0, /menu, /menu, GET, admin, carbon.super, test, 10.10.10.2, admin@carbon.super, carbon.super, 10.10.10.2, curl/7.58.0, 1558537406582, false, 8, 1, 7, false, 0, https-8243, 200, https://localhost:9443/am/sample/pizzashack/v1/api/, 0, 0, 0, 0, 7, 0, SYNAPSE, Synapse], isExpired=false}`

Fixes https://github.com/wso2/product-apim/issues/5023

PR #6585, opened 2019-06-25
fixes #5023

hasuniea:
related to #5023


hasuniea:
The same fix was done from https://github.com/wso2/carbon-apimgt/pull/6563/

PR #6586, opened 2019-06-25
Fixing issue with classloading in velocity engine

harsha89:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/template/APITemplateBuilderImpl.java
```diff
@@ -326,4 +330,25 @@ public String getVelocityLogger() {
...
+    private void initVelocityEngine(VelocityEngine velocityengine) throws APITemplateException {
+        Thread thread = Thread.currentThread();
+        ClassLoader loader = thread.getContextClassLoader();
+        thread.setContextClassLoader(this.getClass().getClassLoader());
+
+        try {
+            velocityengine.init();
+        } catch (Exception e) {
+            String msg = "Error while initiating the Velocity engine";
+            log.error(msg, e);
```
hasuniea: If you catch the error and log it and rethrow it, then there's no way for the upstream code to know that you have already logged the exception, and so the same exception might get logged twice.

harsha89: It's a kind of anti pattern. We had several discussions on whether log and throw or simply throw. Having printing log twice isn't bad. If it go without logging from any where then it would be very hard to find what has happened.

PR #6590, opened 2019-06-26
 Fix annotation

tharindu1st:
Fix annotation


features/apimgt/org.wso2.carbon.apimgt.core.feature/src/main/resources/conf_templates/templates/repository/conf/jndi.properties.j2
```diff
@@ -0,0 +1,35 @@
+#
+        # Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
fazlan-nazeem: license year should be changed

PR #6598, opened 2019-06-27
Add validation for updating swagger defintion using Publisher REST API

vithu30:


components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
```diff
@@ -145,4 +145,13 @@
         return uriTemplateMap;
     }
 
+    /**
```
harsha89: do we have the same validation in the jaggery api basiclaly the UI? We can't have two behaviors. Also up to which level we validae the swagger?

vithu30: We do not have the same validation for jaggery API since REST APIs are used in new UIs. Here, we are using swagger third party library to validate swagger definitions. For instance, if there any mandatory attributes missed in the swagger definition or any invalid additional parameters has been sent as API definition, swagger validation will fail

malinthaprasan:
Can we merge the PR? Is there anything more to add?

vithu30:
We can merge it

PR #6635, opened 2019-07-04
Application Keys Import/Export

kasvith:
Added following changes to admin REST API

- Export keys with Application(using `withKeys` query parameter)
- Import Keys with Application(use `skipApplicationKeys` query parameter to skip keys while import)
- Encode client secrets with Base64 while exporting
- Updated tests


kasvith:
This PR enables using the feature from CLI side wso2/product-apim-tooling#78

PR #6650, opened 2019-07-09
 Add support to add encrypted passwords in sign-up and external stores registry config

vithu30:
With current implementation, password is added as plain text in sign-up-config.xml and external-api-stores.xml. This PR contains changes to give an extension point to user so that they can add encrypted passwords which can be decrypted using custom password resolver.

User can configure the custom password resolver (which should implement PasswordResolver Interface) class in api-manger.xml under <PasswordResolverImpl>.

The default implementation of PasswordResolver is DefaultPasswordResolverImpl.


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/PasswordResolverFactory.java
```diff
@@ -0,0 +1,78 @@
...
+     * This method will take hardcoded class name for password resolver from api-manager.xml file and will return
+     * an instance of PasswordResolver.
+     *
+     * @return PasswordResolver instance.
+     */
+    public static PasswordResolver getInstance() {
+
+        return passwordResolver;
+    }
+}
```
menakaj: Add a new line here

vithu30: Resolved

PR #6729, opened 2019-07-24
Token refresh, throttle, auth type selection support

ruks:
Token refresh on expired
Support selecting resource throttling policy and auth type

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Details/Resources/Resources.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Details/Resources/Resources.jsx
@@ -208,8 +208,8 @@ class Resources extends React.Component {
                 }
             });
 
-        const promised_resPolicies = Api.policies('api');
-        promised_resPolicies
+        const promisedResPolicies = Api.policies('api');
+        promisedResPolicies
             .then((policies) => {
                 this.setState({ apiPolicies: policies.obj.list });
             })
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Details/Resources/Resources.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Details/Resources/Resources.jsx
@@ -136,6 +136,7 @@ class Resources extends React.Component {
             notFound: false,
             showAddResource: false,
             showScopes: false,
+            apiPolicies: [],
         };
         this.api = new Api();
         this.api_uuid = props.api.id;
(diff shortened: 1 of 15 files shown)


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Details/Resources/Resources.jsx
```diff
@@ -207,6 +208,23 @@ class Resources extends React.Component {
                 }
             });
 
+        const promised_resPolicies = Api.policies('api');
```
tmkasun: Please use camel case names

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/data/APIClient.js
```diff
@@ -134,14 +134,44 @@ class APIClient {
...
-            AuthManager.refreshTokenOnExpire(request, this.environment);
+            // todo: sync token checking logic
+            const existingUser = AuthManager.getUser(this.environment.label);
+            if (!existingUser) {
+                console.log('User not found. Token refreshing failed.');
+                return request;
+            }
+            const existingToken = AuthManager.getUser(this.environment.label).getPartialToken();
+            const refToken = AuthManager.getUser(this.environment.label).getRefreshPartialToken();
+            if (existingToken && (existingToken !== 'undefined')) {
```
tmkasun: Is there any possibility of returning `'undefined'` string from `getPartialToken()` method ?

otherwise `&& (existingToken !== 'undefined')` part is redundent here

ruks: only if the value of the cookie gets empty. Just to avoid such an error scenario.

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/data/APIClient.js
```diff
@@ -134,14 +134,44 @@ class APIClient {
...
+                    console.log('Refresh token not found. Token refreshing failed.');
+                    return request;
+                }
+            }
+            const promise = new Promise((resolve, reject) => {
+                AuthManager.refresh(this.environment)
+                    .then(res => res.json())
+                    .then(() => {
+                        request.headers.authorization = 'Bearer '
+                            + AuthManager.getUser(this.environment.label).getPartialToken();
```
tmkasun: I think this will still return the old partial token , You might need to call `setPartialToken` in the current user before getting the partial token

ruks: Since it is a cookie, values get saved automatically once the response is received. Do we have to save it again?

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/data/APIClient.js
tmkasun: Yes I think that could happen only [if this](https://github.com/wso2/carbon-apimgt/blob/master/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/data/Utils.js#L39) `pair[1]` get undefined.

So , IMO it's better we handle it there and return null , if the value is `'undefined'` (note: string)

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/data/APIClient.js
tmkasun: Ahh we are setting the cookie using set-cookie headers here ?
But there is a catch here, If the UI is running in a different host to where jaggery services are running, this approach will break.

PR #6739, opened 2019-07-26
Adding user post self registration handler to impl from keymgt

CrowleyRajapakse:


pom.xml
```diff
@@ -1568,6 +1568,7 @@
 
         <!-- Carbon Identity import version ranges -->
         <carbon.identity.imp.pkg.version>[5.13.6, 6.0.0)</carbon.identity.imp.pkg.version>
+        <carbon.identity.recovery.pkg.version>[1.2.9, 1.3.0)</carbon.identity.recovery.pkg.version>
```
Rajith90: I think we can use the range starting from minot to major [1.2.9, 2.0.0), if there is no specific reason to limit upper range value until a minor version change

PR #6783, opened 2019-08-06
Fix test failures due to apiTypes constants

msm1992:


malinthaprasan:
Should we remove the old CONSTS?

PR #6787, opened 2019-08-07
add support for token type jwt for applications

dushansilva:
this PR will fix https://github.com/wso2/product-apim/issues/5024

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Apis/Listing/SubscriptionPolicySelect.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Apis/Listing/SubscriptionPolicySelect.jsx
@@ -77,38 +77,40 @@ class SubscriptionPolicySelect extends React.Component {
         const { selectedPolicy } = this.state;
 
         return (
-            policies &&
-            <div className={classes.root}>
-                <ScopeValidation
-                    resourcePath={resourcePaths.SUBSCRIPTIONS}
-                    resourceMethod={resourceMethods.POST}
-                >
-                    <Button
-                        variant='contained'
-                        size='small'
-                        color='primary'
-                        className={classes.buttonGap}
-                        onClick={() => {
-                            handleSubscribe(applicationId, apiId, selectedPolicy);
+            policies
+            && (
+                <div className={classes.root}>
+                    <ScopeValidation
+                        resourcePath={resourcePaths.SUBSCRIPTIONS}
+                        resourceMethod={resourceMethods.POST}
+                    >
+                        <Button
+                            variant='contained'
+                            size='small'
+                            color='primary'
+                            className={classes.buttonGap}
+                            onClick={() => {
+                                handleSubscribe(applicationId, apiId, selectedPolicy);
+                            }}
+                        >
+                            <FormattedMessage defaultMessage='Subscribe' id='Apis.Listing.SubscriptionPolicySelect.subscribe' />
+                        </Button>
+                    </ScopeValidation>
+                    <Select
+                        value={selectedPolicy}
+                        onChange={(e) => {
+                            this.setState({ selectedPolicy: e.target.value });
                         }}
                     >
-                        <FormattedMessage defaultMessage='Subscribe' id='Apis.Listing.SubscriptionPolicySelect.subscribe' />
-                    </Button>
-                </ScopeValidation>
-                <Select
-                    value={selectedPolicy}
-                    onChange={(e) => {
-                        this.setState({ selectedPolicy: e.target.value });
-                    }}
-                >
-                    {policies.map(policy => (
-                        <MenuItem value={policy}>
-                            {policy}
-                        </MenuItem>
-                    ))}
+                        {policies.map(policy => (
+                            <MenuItem value={policy}>
+                                {policy}
+                            </MenuItem>
+                        ))}
 
-                </Select>
-            </div>
+                    </Select>
+                </div>
+            )
         );
     }
 }
@@ -118,4 +120,3 @@ SubscriptionPolicySelect.propTypes = {
 };
 
 export default withStyles(styles, { withTheme: true })(SubscriptionPolicySelect);
-
--- a/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Applications.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Applications.jsx
@@ -32,7 +32,7 @@ const applications = () => {
     return (
         <Switch>
             <Route exact path='/applications' component={Listing} />
-            <Route path='/applications/:application_uuid/' render={props => <Details {...props} />} />
+            <Route path='/applications/:application_uuid/' component={Details} />
             <Route component={PageNotFound} />
         </Switch>
     );
--- a/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Applications.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Applications.jsx
@@ -16,7 +16,7 @@
  * under the License.
  */
 
-import React, { Component } from 'react';
+import React from 'react';
 import { Route, Switch } from 'react-router-dom';
 import Listing from './Listing/Listing';
 import Details from './Details/index';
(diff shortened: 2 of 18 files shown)


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Applications.jsx
```diff
@@ -27,16 +27,16 @@ import { PageNotFound } from '../Base/Errors';
...
-        );
-    }
-}
 
-export default Applications;
+const applications = () => {
+    return (
+        <Switch>
+            <Route exact path='/applications' component={Listing} />
+            <Route path='/applications/:application_uuid/' render={props => <Details {...props} />} />
```
tmkasun: This render method is redundant right ?
This is same as
```suggestion
            <Route path='/applications/:application_uuid/' component={Details} />
```
because when a component get rendered through `Route` it will get all the route related props passed by default

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/InfoBar.jsx
```diff
@@ -234,7 +233,9 @@ class InfoBar extends React.Component {
         if (typeof todo === 'boolean') {
             this.setState({ showOverview: todo });
         } else {
-            this.setState({ showOverview: !this.state.showOverview });
+            this.setState((prevState) => {
```
tmkasun: You could just omit the return keyword , if you unwrap the method body from curlybraces 
i:e
```javascript
            this.setState( prevState => ( { showOverview: !prevState.showOverview } ));
```

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/InfoBar.jsx
```diff
@@ -365,7 +366,11 @@ class InfoBar extends React.Component {
                     </Collapse>
                 )}
                 <div className={classes.infoContentBottom}>
-                    <div className={classes.contentWrapper} onClick={() => this.toggleOverview()}>
+                    <div
+                        className={classes.contentWrapper}
+                        onClick={() => this.toggleOverview()}
```
tmkasun: Why are we nesting the `this.toggleOverview` inside a lambda function?

if you need to preserve the `this` object, you could bind `this` to `this.toggleOverview` function in  the constructor

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/InfoBar.jsx
```diff
@@ -365,7 +366,11 @@ class InfoBar extends React.Component {
...
+                        onKeyDown={() => this.toggleOverview()}
```
tmkasun: Same here

```suggestion
                        onKeyDown={this.toggleOverview}
```

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Shared/AppsAndKeys/ViewKeys.jsx
```diff
@@ -136,12 +136,33 @@ class ViewKeys extends React.Component {
...
+     *
+     *
+     * @param {*} prevProps previous props
+     * @memberof ViewKeys
+     */
+    componentDidUpdate(prevProps) {
+        const { isKeyJWT } = this.props;
+        if (isKeyJWT && !prevProps.isKeyJWT) {
+            // eslint-disable-next-line react/no-did-update-set-state
+            this.setState({ isKeyJWT: true });
```
tmkasun: Why we need to do this here ?

I think we could have done this in `render()` method, We get the same prop in `render()` and could have done them if logic check there?
And please add an explanation if adding 

```javascript
// eslint-disable-next-line react/no-did-update-set-state
```
ignore rules 😃

dushansilva: Ran in to a small issue when setting in the render method. we can add it to the render method like this

`        if (jwtGenerated && !isKeyJWT) {
            this.setState({ isKeyJWT: true });
        }
`
where jwtGenerated is the what is passed from props and isKeyJWT is taken from state. But to close the dialog box i need set isKeyJWT false however if we set isKeyJWT to false it will always be set to true in the render method.
we can avoid that in componentDidUpdate with isKeyJWT && !prevProps.isKeyJWT consdition.

reason i ignored the rule is that checked with react docs and according to them its ok to setstate as long as we are checking a condition [1] 

[1] - https://reactjs.org/docs/react-component.html#componentdidupdate

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Applications.jsx
dushansilva: fixed  https://github.com/wso2/carbon-apimgt/pull/6787/commits/b09339ab96cfdcc35f13e351833f406efad26ead

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/InfoBar.jsx
dushansilva: fixed  https://github.com/wso2/carbon-apimgt/pull/6787/commits/b09339ab96cfdcc35f13e351833f406efad26ead

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/InfoBar.jsx
dushansilva: fixed  https://github.com/wso2/carbon-apimgt/pull/6787/commits/b09339ab96cfdcc35f13e351833f406efad26ead

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/InfoBar.jsx
dushansilva: fixed  https://github.com/wso2/carbon-apimgt/pull/6787/commits/b09339ab96cfdcc35f13e351833f406efad26ead

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Shared/AppsAndKeys/ViewKeys.jsx
dushansilva: added a comment to method explaining the reason  https://github.com/wso2/carbon-apimgt/pull/6787/commits/b09339ab96cfdcc35f13e351833f406efad26ead

tmkasun: Moving to render method in the sense , Not moving the logic with `this.setState`. Calling setState in `render()` method is prohibited as it could lead to infinite render loops. I think this is the case for `componentDidUpdate` method as well, But it's safe as long as we handle the infinite loop conditions there.

I think setting the state here is redundant? Do we need to update the state in every prop change ? If so , our state is not local to that component and has to be derived from props. If so [that is an anti-pattern](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html)[1], and we should avoid such scenarios.

[1] https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

dushansilva: created an issue based on discussion https://github.com/wso2/product-apim/issues/5293

features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Apis/Listing/SubscriptionPolicySelect.jsx
```diff
@@ -77,38 +77,40 @@ class SubscriptionPolicySelect extends React.Component {
...
-                        this.setState({ selectedPolicy: e.target.value });
-                    }}
-                >
-                    {policies.map(policy => (
-                        <MenuItem value={policy}>
-                            {policy}
-                        </MenuItem>
-                    ))}
+                        {policies.map(policy => (
+                            <MenuItem value={policy}>
```
tmkasun: Shall we add `key` prop here?

PR #6788, opened 2019-08-07
Changing the carbon-governance and carbon-registry version

selakap:
Changing the carbon-governance and carbon-registry version


tharindu1st:
Doing in next kernel update

PR #6888, opened 2019-08-27
Fix API Resource with dot (.) not displaying properly on API publisher issue

selakap:
Fixing https://github.com/wso2/product-apim/issues/5388


malinthaprasan:
We need to check and fix in React UI

malinthaprasan:
Not reproducible in React UI.

malinthaprasan:
Merged to the old jaggery app

PR #6932, opened 2019-08-30
Fix  backend rest api issues of search

TanyaM:
This PR provides the ability to search as "Bank API" vs Bank API
- "Bank API" --> search for exact string Bank API
- Bank API --> search for either Bank or API

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
@@ -180,7 +180,6 @@ import org.wso2.carbon.user.core.UserRealm;
 import org.wso2.carbon.user.core.common.AbstractUserStoreManager;
 import org.wso2.carbon.user.core.config.RealmConfigXMLProcessor;
 import org.wso2.carbon.user.core.service.RealmService;
-import org.wso2.carbon.user.core.tenant.TenantConstants;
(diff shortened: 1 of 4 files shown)


bhathiya:
Do we support "bank" and "api"?

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
```diff
@@ -7869,7 +7869,8 @@ public static String constructNewSearchQuery(String query) throws APIManagementE
 
         // sub context and doc content doesn't support AND search
         if (inputSearchQuery != null && inputSearchQuery.contains(" ") && !inputSearchQuery
-                .contains(APIConstants.TAG_SEARCH_TYPE_PREFIX4)) {
+                .contains(APIConstants.TAG_SEARCH_TYPE_PREFIX4) && (!inputSearchQuery
```
bhathiya: shall we rename this constant to `TAG_COLON_SEARCH_TYPE_PREFIX`? 
we can rename other similar constants too similar way. 
TAGS_SEARCH_TYPE_PREFIX
TAG_EQ_SEARCH_TYPE_PREFIX etc.

TanyaM: Done

TanyaM:
> Do we support "bank" and "api"?

No we don't support. AND search is not supported.

PR #6937, opened 2019-09-02
Add client certificate api implementation.

menakaj:


malinthaprasan:
@menakaj Do we add client certificates per API or global?
Just wondering this API should be as a base resource or as part of /apis/{apiId}/..

menakaj:
@malinthaprasan Client certificate is added per api and subscription tier and access is being done with the certificate alias. It is also possible to use the api/{api_id} as the base resource.

Then it should be like: /api/{api_id}/certificates/{alias}

malinthaprasan:
Since that is a cleaner approach. Shall we proceed with that?

menakaj:
@malinthaprasan Will do.

PR #7034, opened 2019-09-16
Fix overview page UI

erangatl:
Fix: https://github.com/wso2/product-apim/issues/5572

image


bhathiya:
@erangatl could you please make all fonts in the UI consistent? 

For example, see transport vs Created time vs Business owner.

bhathiya:
and do we need parentheses for tags?

bhathiya:
and I also think the positioning of fields can be optimized. (for example, gateway related and store related fields can be put together. )

PR #7036, opened 2019-09-16
Add Publisher Store UI for publishing APIs for external stores

dushaniw:
This PR adds Publisher UI for publishing APIs for external stores. The capability of publishing to external stores is given only for  API is "PUBLISHED" state. 

Screenshot from 2019-09-17 01-55-09

This PR updates the Store UI for displaying advertised APIs (published as external store APIs).  Only API Documents and API resource views are available for the advertised APIs. 

Screenshot from 2019-09-17 01-52-10
Screenshot from 2019-09-17 01-52-55

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIExportUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIExportUtil.java
@@ -64,13 +64,11 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.AbstractMap;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamException;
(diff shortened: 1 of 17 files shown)


bhathiya:
In the external stores table, column name alignment can be improved?

dushaniw:
> In the external stores table, column name alignment can be improved?

The leftmost column, "Name" is left-aligned while others are right-aligned. Any suggestions on how it could improve? :)

bhathiya:
> > In the external stores table, column name alignment can be improved?
> 
> The leftmost column, "Name" is left-aligned while others are right-aligned. Any suggestions on how it could improve? :)

if we center all, won't that look nice?

dushaniw:
> > > In the external stores table, column name alignment can be improved?
> > 
> > 
> > The leftmost column, "Name" is left-aligned while others are right-aligned. Any suggestions on how it could improve? :)
> 
> if we center all, won't that look nice?

If we center, it would look like this. Isn't the gap between the checkbox and the store name looks too wide?

Screenshot from 2019-09-17 16-08-16

bhathiya:
Can't we center the column names only? while keeping the values left aligned?

dushaniw:
> The gap is ok for me.
> 
> However, can't we center the column names only? while keeping the values left aligned?
Tried that option as well, it wouldn't work. Because we are keeping a min-width for the table and normally the values aren't that long.
Screenshot from 2019-09-17 16-18-13
 
> 
> Anyway, we should decide one way and use it for all tables in the entire app. Maybe we can discuss with all.

Checked the other places where tables are used (API listing, API history table etc). In all of them, all the columns names and values are left-aligned.

Screenshot from 2019-09-17 16-20-26

+1 to discuss with all and come into a common alignment pattern.

bhathiya:
The last one looks ok to me. 

@erangatl @tmkasun @chanaka3d @TanyaM WDYT?

PR #7054, opened 2019-09-17
Adding Mutual SSL UI support

mushthaq33:
Adding the Mutual SSL support for Publisher UI.

Screen Shot 2019-09-17 at 11 25 15 AM
Screen Shot 2019-09-17 at 11 24 58 AM


bhathiya:
Isn't the color of the delete button and tick box too light?

mushthaq33:
> Isn't the color of the delete button and tick box too light? and isn't it better if the delete button is reddish?
> 
> @erangatl @tmkasun

This is the theme color used across publisher. May be we can change the color in the theme so that it will reflect everywhere.

bhathiya:
> > Isn't the color of the delete button and tick box too light? and isn't it better if the delete button is reddish?
> > @erangatl @tmkasun
> 
> This is the theme color used across publisher. May be we can change the color in the theme so that it will reflect everywhere.

yes, we should, I think. I mentioned the same in https://github.com/wso2/product-apim/issues/5541 too.

PR #7061, opened 2019-09-18
Changing delete icon color

mushthaq33:
Changing the delete icon color. Note: need to change the global error color upon agreeing on a color.

Screen Shot 2019-09-18 at 12 01 03 PM


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Details/Endpoints/GeneralConfiguration/Certificates.jsx
```diff
@@ -235,9 +235,8 @@ function Certificates(props) {
                                         </IconButton>
                                         <IconButton
                                             onClick={() => setCertificateToDelete({ open: true, alias: cert.alias })}
-                                            color='secondary'
                                         >
-                                            <Icon>delete</Icon>
+                                            <Icon color='error'>delete</Icon>
```
isharac: Can't we use the color as red instead?

erangatl: IMHO we should not use a different color for delete icon, what we need to have is either a conformation box or a way to revert the action for some pre defined time.

PR #7070, opened 2019-09-18
Add swagger download to store UI

Krishanx92:
Fix : https://github.com/wso2/product-apim/issues/5604

previous


bhathiya:
@Krishanx92 why do we have ":Bearer" inside access token?

PR #7092, opened 2019-09-19
Remove unused API Product Store REST API

uvindra:
The Store REST API now serves both APIs and API Prodcts via the same '/api' resources.
Hence this commit removes the redundant '/api-product' resources, their implementation and
their usage.


uvindra:
Local build with tests passes. Travis CI may have failed due to nexus slowness

PR #7108, opened 2019-09-19
Improve configuration page UX

erangatl:
Improve configuration page UX
image

TODO: Move design section to a subpage, Add Mediation policies


bhathiya:
1. Shall we make "Configuration" plural in all cases? (i.e. menu, top level, CORS etc.)
2. 2nd word of Schema validation and response caching should be upper case?

bhathiya:
image

1. is >> are
2. the 2 lines should have different fonts/sizes
3. uppercase in the middle of a sentence
4. Need to be consistant with other field names. 
2.

PR #7128, opened 2019-09-20
Add a button to view invoice in store.

chamindias:


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/store-new/source/src/app/components/Applications/Details/SubscriptionTableData.jsx
```diff
@@ -86,6 +87,18 @@ class SubscriptionTableData extends React.Component {
         }
     }
 
+    /**
+     * Handle onclick for view invoice
+     * @param {*} subscriptionId subscription id
+     * @memberof SubscriptionTableData
+     */
+    handleRequestInvoice(subscriptionId) {
```
tmkasun: I think this method definition is redundant? Why can't we use the prop in the Button `onClick` event directly? , We could use defaultProps to handle the undefine issues

PR #7129, opened 2019-09-20
Fix Rest api resources are not shown in store when created using a WSDL

AmaliMatharaarachchi:
Fix https://github.com/wso2/product-apim/issues/5606


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher-new/source/src/app/components/Apis/Create/WSDL/ApiCreateWSDL.jsx
```diff
@@ -126,7 +126,8 @@ export default function ApiCreateWSDL(props) {
                 },
             };
         }
-
+        additionalProperties.gatewayEnvironments = ['Production and Sandbox'];
+
```
chamilaadhi: lets put a constant and refer it here

AmaliMatharaarachchi: Done. Thanks.

chamilaadhi:
I don't think this fix is correct. Hardcoding the environment name is not correct. we need to get the values from the api-manager.xml file and add it here. Please check what happens when there are multiple gateway environements as well.

PR #7146, opened 2019-09-22
Fix issues in basic auth flow caches

nalakas:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/caching/CacheProvider.java
```diff
@@ -166,6 +185,51 @@ public static Cache createInvalidTokenCache() {
...
 
+    /**
+     * Create and return basic authenticated resource request cache
+     */
+    public static Cache createGatewayBasicAuthResourceCache() {
+        String apimGWCacheExpiry = getApiManagerConfiguration().getFirstProperty(APIConstants.TOKEN_CACHE_EXPIRY);
+        if (apimGWCacheExpiry != null) {
+            return getCache(APIConstants.API_MANAGER_CACHE_MANAGER, APIConstants.GATEWAY_BASIC_AUTH_RESOURCE_CACHE_NAME,
+                    Long.parseLong(apimGWCacheExpiry), Long.parseLong(apimGWCacheExpiry));
+        } else {
```
praminda: FYI,
`else` is not required here. Simply having the,
```java 
long defaultCacheTimeout = getDefaultCacheTimeout();
...
```
below if statement will do the job. There's no harm in keeping the else :relieved:

nalakas: yes. :)

PR #7149, opened 2019-09-23
Removing subscription creation approval requests upon retired API deletion

tgtshanika:
Fixing https://github.com/wso2/product-apim/issues/4453


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIProviderImpl.java
```diff
@@ -1689,6 +1689,7 @@ public boolean updateAPIforStateChange(APIIdentifier identifier, String newStatu
                         } else { // API Status : RETIRED or CREATED
                             Map<String, String> failedToRemoveEnvironments = failedGatewaysMap;
                             if(!APIConstants.CREATED.equals(newStatus)) {
+                                cleanUpPendingSubscriptionCreationProcessesByAPI(api.getId());
```
bhathiya: This handles only 'retired' state. Other states are already handled?

PR #7184, opened 2019-09-25
Restyling Mediation Policies Page

chanaka3d:
Screen Shot 2019-09-25 at 11 36 02 AM


bhathiya:
shouldn't we replace the gateway with the input boxes? The flow is not reflected in this right?

chanaka3d:
> shouldn't we replace the gateway with the input boxes? The flow is not reflected in this right?

This UI is for the custom mediation selection page ( for In, Out and Fault ). I am not clear about the input boxes you are referring here for gateway.

PR #7186, opened 2019-09-25
Add review changes to HIG

erangatl:


tmkasun:
@erangatl Are we going to maintain the HIG in docs directory?

erangatl:
> @erangatl Are we going to maintain the HIG in docs directory?
We can decide on that, I'm +1 for html view as it's easy to refer when the guide get lengthy.

PR #7261, opened 2019-10-01
Fixing store related issues

chanaka3d:
https://github.com/wso2/product-apim/issues/5336
https://github.com/wso2/product-apim/issues/5726

screencapture-localhost-9443-devportal-apis-d2c718eb-8f33-4907-ac49-89f1b39b1d6a-credentials-2019-10-01-12_10_43

screencapture-localhost-9443-devportal-apis-d2c718eb-8f33-4907-ac49-89f1b39b1d6a-overview-2019-10-01-12_08_14


chanaka3d:
> Access token should be two words. May be letters should be captialize right after the ->. Can't we have some meaningful name for WIZARD?

The access token word is splitted to two words.

PR #7274, opened 2019-10-01
Fixed operation updated issue

ruks:
Fixed operation updated issue

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Resources/Resources.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Resources/Resources.jsx
@@ -166,9 +166,21 @@ export default function Resources(props) {
                 }
                 const regEx = /(?<=\{)(?!\s*\{)[^{}]+/g;
                 const params = data.target.match(regEx);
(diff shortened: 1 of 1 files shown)


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Resources/Resources.jsx
```diff
@@ -166,17 +166,33 @@ export default function Resources(props) {
                 }
                 const regEx = /(?<=\{)(?!\s*\{)[^{}]+/g;
```
tmkasun: @ruks , I think this regex contains a [lookbehind](https://v8.dev/blog/regexp-lookbehind-assertions) pattern which is [not supported](https://v8.dev/blog/regexp-lookbehind-assertions) in FF, Shall we use an alternative pattern ?

ruks: yes. seems it is not supported in FF. Will use an alternative patterns.

ruks: Fixed

PR #7310, opened 2019-10-03
Adding Wizard for APIProduct Creation flow

HiranyaKavishani:
Step1 :
Screen Shot 2019-10-03 at 4 19 36 PM

Step2 : 
When APIs are available
Screen Shot 2019-10-03 at 4 38 37 PM

When APIs are not available
The following message has been modified to "No REST APIs are created yet"

Screen Shot 2019-10-03 at 4 17 33 PM


bhathiya:
image

A few suggestions:

1) The height of the input box seems too much compared to the width. Can we reduce the height a bit?  
2) We might need to simplify the wording. Something like this?

`Resources of the {$API_Name} API are shown below. ...` API in `api product should be in upper case`.

3) HTTP verbs need to be uppercase

4) shall we add a similar API product name + description to the RHS panel as well? Then we can get the 2 resources list to the same level. 

5) 2 points seem too far from each other. can we make them a bit closer?

bhathiya:
image

1) I don't think we need to give validation guidelines by default. do we do that anywhere else? 

2) and 4)  "API Product"?

3) one or "more" throttling "policies"?

HiranyaKavishani:
@bhathiya, Thank you very much for the above suggestions.

I will modify the resource step as you suggested.

> 2 points seem too far from each other. can we make them a bit closer?
But here, If we do so, consistency with other Wizards will be dropped. Because in everywhere,
(Length of the Wizard = length of the form) 

Regarding API Product Creation  Page,
This form is the same one (DefaultAPIForm), which has been used for the Rest API Creation also, Therefore instead of 2) and 4), the issues have been there.

bhathiya:
> > 2 points seem too far from each other. can we make them a bit closer?
> 
> But here, If we do so, consistency with other Wizards will be dropped. Because in everywhere,
> (Length of the Wizard = length of the form)

losing consistency a bot over UX is fine for me. let's see others' opinion as well.  

> 
> Regarding API Product Creation Page,
> This form is the same one (DefaultAPIForm), which has been used for the Rest API Creation also, 

I think we need a new one for API product.

HiranyaKavishani:
I have seen the same issue that you pointed here(1 and 3) in API creation

> 1. I don't think we need to give validation guidelines by default. do we do that anywhere else?
> 3. one or "more" throttling "policies"?


Screen Shot 2019-10-03 at 10 43 22 PM

tmkasun:
Hi @bhathiya 


I'm -1 to the following change,

> 5. 2 points seem too far from each other. can we make them a bit closer?

Currently, we are using the default stepper labels styling coming from the [material-ui library](https://material-ui.com/components/steppers/#steppers) which has designed the stepper according to the [material.io design](https://material.io/archive/guidelines/components/steppers.html#steppers-usage) principles.
image

 So I think the current look and feel is more appropriate to the stepper label than the suggested changes.

On the other hand, If we shrink the wizard labels section less than the width of the content, It will give an impression, the content is something out of the wizard.

bhathiya:
> Hi @bhathiya
> 
> I'm -1 to the following change,
> 
> > 1. 2 points seem too far from each other. can we make them a bit closer?
> 
> Currently, we are using the default stepper labels styling coming from the [material-ui library](https://material-ui.com/components/steppers/#steppers) which has designed the stepper according to the [material.io design](https://material.io/archive/guidelines/components/steppers.html#steppers-usage) principles.
> image
> 
> So I think the current look and feel is more appropriate to the stepper label than the suggested changes.
> 
> On the other hand, If we shrink the wizard labels section less than the width of the content, It will give an impression, the content is something out of the wizard.

when there are more than 2 steps, it looks ok. but when there are just 2 steps it doesn't look like a wizard. anyway it's just what I feel. if it's just me then let's go ahead with above.

HiranyaKavishani:
What gives a better look for stepper?

Form 1. 
 Screen Shot 2019-10-03 at 4 19 36 PM

Form 2.
Screen Shot 2019-10-04 at 2 57 43 PM

Krishanx92:
> What gives a better look for stepper?
> 
> Form 1.
> Screen Shot 2019-10-03 at 4 19 36 PM
> 
> Form 2.
> Screen Shot 2019-10-04 at 2 57 43 PM

+1 with form 1.
Also for clarification why next and cancel buttons placed as such 2 corners, is it a UX guide for wizards?

kavishkafernando:
+1 for form 2.

tmkasun:
+1 with form 1 as mentioned earlier

tmkasun:
And the action buttons should come next to each other with major action button to the right 

image

HiranyaKavishani:
Form 1:
Screen Shot 2019-10-06 at 11 12 33 PM

Form 2:
Screen Shot 2019-10-06 at 11 25 36 PM

HiranyaKavishani:
Form 1:
Screen Shot 2019-10-06 at 11 12 33 PM

Form 2:
Screen Shot 2019-10-06 at 11 46 23 PM

TanyaM:
> What gives a better look for stepper?
> 
> Form 1.
> Screen Shot 2019-10-03 at 4 19 36 PM
> 
> Form 2.
> Screen Shot 2019-10-04 at 2 57 43 PM

+1 for the Form 2 with the button changes suggested. But I feel we need to have the horizontal line below the wizard stepper. If not it looks more like a title of the selected step.

screenshot

erangatl:
Shall we change the Cancel and Back buttons to text buttons 
image

mushthaq33:
+1 for form 2. And one more suggestion. Isn't it better to bring the labels below the number icon in the stepper?

shaniR:
A few suggestions:
**General**
1) The heights of the boxes are not the same.
2) Once we add a resource to the API product, should it still be displayed under the product?

**Stepper**
1) the first step says "provide details". How about changing it to "Define API Product" ? 
2) The description below the main heading "Create API Product" 

**Select API**
1) What does visible mean? visible in the list? 
2) is the filtering done based on name? if so,  Should the text below the text box, say "Filter the APIs by API name"

**API Resources**
1) Do we need the description, and the instructions? It is quite intuitive right? The description should only say that these are the resources of the selected API.
2) If we do need the description,  "areshown" , missing space, and we need to reword the instruction.
3) The arrow in front of each resource is a single arrow head, where as the action buttons on the  top, which also adds items to the right hand side box, has doble arrow heads. Should we stick to one design of arrow heads?
4) The heading for this section should be an action ("Select API Resource") as well. Since the first box heading is an action, let's keep it uniform across the boxes. 
5) The font of the authorization and tier, can it be one size smaller? As it is shaded to be additional info, it should be less prominent than the main text.


**API Product**
1) Do we need the description?
2) There's no heading for the box ("Design API Product")? - 
3) Can the resources under an API be collapsible? - An improvement
4) "APIProduct" space is missing.

HiranyaKavishani:
Final Outcome according to the above suggestions.
Screen Shot 2019-10-07 at 12 24 01 PM

Screen Shot 2019-10-07 at 12 23 53 PM

https://github.com/wso2/carbon-apimgt/issues/7343 This issue will track the related issues in the API creation forms

shaniR:
1)Was it decided not to change the arrow head for an individual resource?
2) the third box, still does not have a heading.

HiranyaKavishani:
Discussed with @erangatl, @chanaka3d Regarding the 1) and 2)
1) Yes
2) We don't need heading as all two columns represent API Resources

PR #7348, opened 2019-10-07
Add CORS support for store/publisher REST APIs

ruwiniwj:
Add CORS support for store/publisher REST APIs


ruwiniwj:
Fix https://github.com/wso2/product-apim/issues/5274

PR #7381, opened 2019-10-08
Fix sidebar menu items order

erangatl:
Fixes: https://github.com/wso2/product-apim/issues/6150

image


bhathiya:
> Did we agree to move the Subscription tab to a higher level?

we tried to get the list of business plans to runtime and then keep the subscriptions at a lower level. but the runtime config page doesn't have any more space left. so we can't move the business plans there. so we had to take the subscriptions up.

PR #7385, opened 2019-10-09
Fixed https://github.com/wso2/product-apim/issues/6237

ruks:
Fixed https://github.com/wso2/product-apim/issues/6237


tmkasun:
@taniamahanama  and @ruks I too have sent a fix for this particular issue in [here](https://github.com/wso2/carbon-apimgt/pull/7359/files#diff-b482d1f71c70e6990d68abcab270327aR42), 

I think the correct approach would be, to define the indexes in the theme file and referring them in required places, otherwise, if we put random values in an ad-hoc manner, it's not scalable. Developers could come up with different values like -1 -2 100 10000 to fix overlay issues and ultimately will roun out of zindex values.
[MUI doc](https://material-ui.com/customization/z-index/)

If adding a zindex shall we define it [in the theme](https://material-ui.com/customization/default-theme/?expend-path=$.zIndex) first and refer it in the required place

ruks:
@tmkasun is this already fixed or need additional improvements?

PR #7393, opened 2019-10-09
Fix store comment page issues

nalakas:
fixes for:https://github.com/wso2/product-apim/issues/5786

Screen Shot 2019-10-09 at 11 49 23 AM


kavishkafernando:
The number of characters allowed is too much to the corner. Try reducing the size of the text box or add a margin right.
comments

nalakas:
fixed

Screen Shot 2019-10-09 at 12 24 38 PM

PR #7443, opened 2019-10-10
Fix server startup error log print

Krishanx92:


Krishanx92:
Fix: https://github.com/wso2/product-apim/issues/5973

PR #7558, opened 2019-10-15
Fix properties page style and typo

erangatl:
Fixes: https://github.com/wso2/product-apim/issues/6385

image


harsha89:
We can improve the long text in the display. "is case senstivie" is not required. We need to combine with and. Shall we have a quick view on the long detail messsage and fix it as well.

bhathiya:
don't need here right?

image

erangatl:
> don't need here right?
> 
This is same UI when user is editing the table inline, so we should not hide it, but better if we can disable it.

PR #7563, opened 2019-10-16
Add headings to Subscription elements

kavishkafernando:
This PR has the following changes,
- Fixes issue https://github.com/wso2/product-apim/issues/6424
- Remove extra space when selecting specific tenants
- Remove subscription availability label

Screen Shot 2019-10-16 at 10 42 19 AM


bhathiya:
about the tenant list input, should that be a standard input that we use everywhere else?

PR #7574, opened 2019-10-16
[Developer Portal] Fix issue with getComments for Anonymous User

sanjulamadurapperuma:
Fixes [wso2/product-apim#6484](https://github.com/wso2/product-apim/issues/6484)

image


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/UserAwareAPIConsumer.java
```diff
@@ -127,18 +127,6 @@ public Comment getComment(Identifier identifier, String commentId) throws APIMan
...
-        checkSubscribePermission();
-        return super.getComments(apiTypeWrapper);
-    }
-
-    @Override
-    public void deleteComment(APIIdentifier identifier, String commentId) throws APIManagementException {
-        checkSubscribePermission();
-        super.deleteComment(identifier,commentId);
-    }
-
```
chamilaadhi: I don't think comment deletion should be allowed for an anonymous user

sanjulamadurapperuma: Sorry, error on my end. Fixed now. Only getComment and getComments methods should be removed. Thanks for pointing out.

PR #7617, opened 2019-10-17
Condition group loading

ChamodDamitha:
This will add the functionalities mentioned below. 

- Retrieving condition groups of all tenants at the startup and storing it in a Map at the Gateway. 

- When an Advanced Throttle Policy is added/updated/deleted, the map will be updated through JMS.

- When verbInfoDTO is populated, the condition groups retrieved from the map will be used.

Fixes https://github.com/wso2/product-apim/issues/6432


ChamodDamitha:
Could be fixed by this https://github.com/wso2/carbon-apimgt/pull/7620

PR #7661, opened 2019-10-19
Reviewing UI text

ruthryi:
- Reviewed UI text in the locale json files for Publisher and Dev Portal
- Checked spelling and grammar


erangatl:
Seems like json files are directly edited. These changes need to be done on respective files and generate.

bhathiya:
@ruthryi Fixes are fine, but as @erangatl has mentioned we may have to change the fix locations.

PR #7764, opened 2019-10-23
Fix product-apim/issues/6620

AmaliMatharaarachchi:
Fix https://github.com/wso2/product-apim/issues/6620 - When tenant updates an API resource, they are not visible to others

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AbstractAPIManager.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AbstractAPIManager.java
@@ -1213,7 +1213,7 @@ public abstract class AbstractAPIManager implements APIManager {
             Registry registryType;
             //Tenant store anonymous mode if current tenant and the required tenant is not matching
             if (this.tenantDomain == null || isTenantDomainNotMatching(apiTenantDomain)) {
-                if (apiTenantDomain != null){
+                if (apiTenantDomain != null) {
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AbstractAPIManager.java
```diff
@@ -1208,10 +1208,13 @@ public String getGraphqlSchemaDefinition(APIIdentifier apiId) throws APIManageme
     public String getOpenAPIDefinition(Identifier apiId) throws APIManagementException {
         String apiTenantDomain = getTenantDomain(apiId);
         String swaggerDoc = null;
+        boolean tenantFlowStarted = false;
         try {
             Registry registryType;
             //Tenant store anonymous mode if current tenant and the required tenant is not matching
             if (this.tenantDomain == null || isTenantDomainNotMatching(apiTenantDomain)) {
+                startTenantFlow(apiTenantDomain);
```
mushthaq33: This will not work for cross tenants with logged in users.

Rajith90: Yes, I also thinks still registry would be the wrong one, unless we explicitly load the registry

AmaliMatharaarachchi: We tested it. This works as expected.

PR #7767, opened 2019-10-23
Show confirmation dialog when changing endpoint type

menakaj:
Fixes https://github.com/wso2/product-apim/issues/6618
Screenshot (23)


bhathiya:
@menakaj both buttons should have the primary color.

PR #7815, opened 2019-10-24
Fix issues with reverse proxy with custom contexts

ibrohimislam:
latest attempt to fix issues with reverse proxy with custom contexts from @bhathiya makes reverse proxy unusable. This pull request try to fix it by including custom context to reverse proxy config.

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/services/login/idp.jag
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/services/login/idp.jag
@@ -23,19 +23,20 @@
     include("/services/constants.jag");
     include("/services/jagg.jag");
 
-    var serverUrl = '';
+    var serverUrl = utils.getServerURL();
+    var serverUrlWithContext = serverUrl + app.context;
     if (app.reverseProxy.enabled === true) {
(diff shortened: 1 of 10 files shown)


bhathiya:
Hi @ibrohimislam,

Thank you very much for your contribution. Really appreciate it. 

Please note that the current implementation was done as per a design decision that was taken to let the reverse proxy make the context change so that APIM itself does not have to worry about the new context.

You can find the related NGinx configurations in https://github.com/wso2/product-apim/issues/6467 

To get the custom context working you have to change the redirect rules such as

`proxy_redirect  https://localhost:9443/publisher  https://localhost/publisher;`

to

`proxy_redirect  https://localhost:9443/publisher  https://localhost/newpublishercontext;`

However, there was just one case where we thought that APIM should know about the context and the reverse proxy base URL, that is to use for the callback URL which is used in the /authorize call to the IDP. The reason is this URL goes in the URL as a query parameter and user seeing a localhost URL with the original context does not very good.

ibrohimislam:
ok, thank you for the clarification.  I run into a problem when using `utils.getServerURL()` because it uses  HostName instead of MgtHostName. But, maybe it is another issue.

bhathiya:
Yes, ideally it should read from the `MgtHostnName`, but in publisher and dev portal, the values for `HostName` and `MgtHostName` are the same, typically.

PR #7825, opened 2019-10-29
adding additional properties to service provider

nipun-sachintha:
adding several features like userAccessExpiry time,application expiry time,renew refresh token enable,set federated identity provider to service provider.
In tenant-conf json variable isAdvance is added to view these configurations in advance view.


bhathiya:
Closing the PR until we rewrite this feature.

PR #7858, opened 2019-11-13
fix for product-apim#6817

isharac:
According to this implementation, internal calls route through "caron.server.host:${mgt.transport.https.port}
Hence, when you deploy in Kubernetes NodePort you need to use node port as the proxyPort in deployment.toml


isharac:
move configuration to settings.js
remved reverseProxy config and added custonUrl to support custom URLs.

    app: {
        context: '/devportal',
        customUrl: {
            enabled: false,
            forwardedHeader: 'X-Forwarded-For',
        },
        origin: {
            host: 'localhost',
        },
    },

dushaniw:
Fix https://github.com/wso2/product-apim/issues/6817

PR #8039, opened 2019-12-13
Fix publisher restriction overriding visible to my domain accessibility

1akshitha:
Fixing wso2/product-apim#7106
If publisher access control is set to restrict to roles and store visibility to visible to my domain, store visibility is set only for the restricted roles of the publisher. With this PR, even though publisher access is restricted, store visibility to visible to my domain will set null as the role in registry.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
@@ -3578,7 +3578,8 @@ public final class APIUtil {
                     publisherAccessRoles = new StringBuilder(APIConstants.NULL_USER_ROLE_LIST);
                 }
 
-                if (visibility.equalsIgnoreCase(APIConstants.API_GLOBAL_VISIBILITY)) {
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
```diff
@@ -3578,7 +3578,8 @@ public static void setResourcePermissions(String username, String visibility, St
                     publisherAccessRoles = new StringBuilder(APIConstants.NULL_USER_ROLE_LIST);
                 }
 
-                if (visibility.equalsIgnoreCase(APIConstants.API_GLOBAL_VISIBILITY)) {
+                if (visibility.equalsIgnoreCase(APIConstants.API_GLOBAL_VISIBILITY)
+                        || visibility.equalsIgnoreCase(APIConstants.API_PRIVATE_VISIBILITY)) {
```
Rajith90: Its better to use the equalsIgnoreCase, with the constant which is guaranteed not to be null

1akshitha: Fixed with https://github.com/wso2/carbon-apimgt/pull/8152

PR #8070, opened 2020-01-09
Update APIUtilTest.java to make it work anywhere on earth

mmuesly:
The test initialization code uses new Date().toString(). Parsing this string with a SimpleDateFormat instance that is initialized with a non US locale does not work.

This change stabilizes the test results for people who run their laptop in another default language than us englisch. It makes the `man clean install` result more reliable.


malinthaprasan:
@mmuesly Thank you very much for the contribution.

PR #8151, opened 2020-02-05
fix for registry resource is added as the mediation seuence file name

isharac:
fix for  wso2/product-apim#7003

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIExportUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIExportUtil.java
@@ -568,7 +568,7 @@ public class APIExportUtil {
             String sequenceFileName = sequenceDetails.getKey();
             OMElement sequenceConfig = sequenceDetails.getValue();
             CommonUtil.createDirectory(pathToExportedSequence);
-            String exportedSequenceFile = pathToExportedSequence + sequenceFileName;
+            String exportedSequenceFile = pathToExportedSequence + sequenceFileName + APIConstants.XML_EXTENSION;
             try (OutputStream outputStream = new FileOutputStream(exportedSequenceFile)) {
                 sequenceConfig.serialize(outputStream);
                 if (log.isDebugEnabled()) {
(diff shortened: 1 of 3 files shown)


dushaniw:
@isharac 
I further checked, and it seems the exported sequence file name is the exact file name in the registry. 

The fix in [1] will export the default sequences in `/apimgt/customsequences` with `.xml` extension appended twice. Hence instead of fix in [1], better if we fix it in [2]  with following,

` String sequenceFileName = sequenceName + APIConstants.XML_EXTENSION`

and  [3] with following,
`pathToExportedSequence += APIImportExportConstants.CUSTOM_TYPE + File.separator;`

So `.xml` will be always appended to the sequence name, not to the file name in the registry.

To keep the same behavior as in 2.x versions, in which to keep the file name in registry equivalent to the sequence name (without the extension) and when importing we need to remove the extension from the file before putting it to the registry. For that need to fix here [4]

```
String inSequenceFileName = importedApi.getInSequence();
String inSequenceFileLocation = pathToArchive + 
           APIImportExportConstants.IN_SEQUENCE_LOCATION   + 
           APIImportExportConstants.CUSTOM_TYPE + File.separator + inSequenceFileName       //Adding in-sequence, if any
 if (CommonUtil.checkFileExistence(inSequenceFileLocation + APIConstants.XML_EXTENSION)) {
            String inSequencePath = APIConstants.API_CUSTOM_SEQUENCE_TYPE_IN + RegistryConstants.PATH_SEPARATOR + inSequenceFileName;
            addSequenceToRegistry(true, registry, inSequenceFileLocation + 
                      APIConstants.XML_EXTENSION, regResourcePath + inSequencePath);
   }
```

[1] https://github.com/wso2/carbon-apimgt/pull/8151/files#diff-e71d8b3a68ee4d93c3baf6932f4c162dR571
[2] https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIExportUtil.java#L534
[3]https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIExportUtil.java#L454
[4] https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java#L679

PR #8154, opened 2020-02-06
Single Logout with OIDC Session Management methodology

CrowleyRajapakse:
## Issue
product-apim: https://github.com/wso2/product-apim/issues/7410
also resolves logout doesn't revoke user access token issue

## Methodology
Implement Single Logout feature with respect to OIDC Session Management.
- Implemented RP Iframe.
- Stored Client ID and session_state in browser cookies in login phase.
- Invoke the OIDC checksession endpoint to load the OP Iframe using the stored client ID and session state.
- After the session state returns from the response either 'changed' or 'unchanged', handle the response accordingly.


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/package.json
```diff
@@ -32,6 +32,7 @@
         "draftjs-to-html": "^0.8.4",
         "file-loader": "^0.11.2",
         "html-to-draftjs": "^1.4.0",
+        "js-cookie": "^2.2.1",
```
tmkasun: Why we need this library? can't we use `document.cookies` directly

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/package.json
```diff
@@ -52,6 +53,7 @@
         "react-dom": "^16.10.2",
         "react-draft-wysiwyg": "^1.13.2",
         "react-dropzone": "^v10.1.7",
+        "react-iframe": "^1.8.0",
```
tmkasun: Why we need a library for adding an iframe to react app?

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/ProtectedApp.jsx
```diff
@@ -63,8 +67,13 @@ export default class Protected extends Component {
         super(props);
         this.state = {
             settings: null,
+            clientId: Cookies.get(User.CONST.PUBLISHER_CLIENT_ID),
```
tmkasun: I think these should be accessed via the User interface

tmkasun: for example `User.getClientID` otherwise managing, especially clearing these cookie stores would be hard,

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/package.json
CrowleyRajapakse: There is a method in ./Utils getCookie() method which return the relevant cookie by using document.cookies, but since it append the environment name to the given name. I'll create a separate method which returns the cookie and use that instead of above library

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/package.json
CrowleyRajapakse: Otherwise we have to write a separate component to handle the iframe logic. This library seems to handle all that.

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/ProtectedApp.jsx
CrowleyRajapakse: These cookies are created in the login_callback.jag and stored indefinitely in the browser, CLIENT_ID cookie value doesn't change often but session_state cookie will be changed after every logout/login. So these cookies are managed by the login and logout process. I've now changed the retrieval of cookies as follows,

` 
clientId: Utils.getCookieWithoutEnvironment(User.CONST.PUBLISHER_CLIENT_ID),
sessionStateCookie:Utils.getCookieWithoutEnvironment(User.CONST.PUBLISHER_SESSION_STATE),`

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/package.json
CrowleyRajapakse: removed the library and resolved the issue

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/package.json
CrowleyRajapakse: removed the react-iframe library

PR #8199, opened 2020-02-17
Add response cache extension

AmaliMatharaarachchi:
Fix https://github.com/wso2/product-apim/issues/7468
Add following extension to the oas definition
```
  x-wso2-response-cache: 
    enabled: true
    cacheTimeoutInSeconds: 300
```


ruks:
@AmaliMatharaarachchi, is it necessary to have 'enabled' attribute here? can we just use the time out value?
```
ex:
x-wso2-response-cache-timeout: 300
```

AmaliMatharaarachchi:
@ruks yes. the enabled is required as we need it to check if the response caching is disabled

ruks:
> @ruks yes. the enabled is required as we need it to check if the response caching is disabled

What if we consider cache is enabled, if x-wso2-response-cache-timeout is defined. Otherwise, we consider as disabled.

AmaliMatharaarachchi:
> > @ruks yes. the enabled is required as we need it to check if the response caching is disabled
> 
> What if we consider cache is enabled, if x-wso2-response-cache-timeout is defined. Otherwise, we consider as disabled.

IMO we cannot consider that response caching is disabled when the extension is not defined in the API definition. Actually, in microgateway scenario default behaviour (without the extension) should be response caching enabled with default values. If we want to configure/disable response caching,  the extension should provide this support.

PR #8200, opened 2020-02-17
Mandated HTTPS transport if mutual SSL is in use.

isuruh15:
Added an info message which appears when disabling HTTPS with mutual SSL enabled. 

Fixes wso2/product-apim#7398


menakaj:
Add ' Fixes https://github.com/wso2/product-apim/issues/7398 ' in the PR description.

PR #8232, opened 2020-02-21
Improvements to the theming in devportal

chanaka3d:
Addressed the following issue as a workaround.

https://github.com/wso2/product-apim/issues/6733
Without the use of the Tomcat filter, we can now only put the content that's bee overridden by the theme into the defaultTheme.js. 

The above method will prevent from replacing the defaultTheme.js changes that a customer has done via an update in the product original source.

This PR is also adding the Arabic language to the list of default languages. It also adds an additional attribute for each language of language switcher to set the "LTR" or "RTL" for each language when it's changed via the language switcher. The default will be LTR. Only if we specify specifically the direction, the page layout will change.

Language can be different for each tenant with the help of the theme. For example, if the supper admin has language French the tenant can have English.

screencapture-localhost-9443-devportal-apis-2020-02-22-07_46_09

Screen Shot 2020-02-22 at 7 47 04 AM


tmkasun:
@shaniR @Mariangela , Please note I think we need to document this change for APIM 3.1.0

PR #8274, opened 2020-03-01
Add x-wso2-mutual-ssl extension and change transport extension

AmaliMatharaarachchi:
The following format[1] of x-wso2-transport extension was later changed to [2] to support mutual ssl. But after this change, format [1] gives parsing issues.

```
[1]x-wso2-transports:
  - "http"
  - "https"
[2]x-wso2-transports:
  http: true
  https: true
  mutualssl: "mandatory"
```
This causes backword compatibility issues. Hence the extension was changed to below

```
x-wso2-transports:
- "http"
- "https"
x-wso2-mutual-ssl: "mandatory"
```

Fixing https://github.com/wso2/product-microgateway/issues/1083
Related to https://github.com/wso2/carbon-apimgt/pull/8153

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
@@ -1096,27 +1096,16 @@ public class OASParserUtil {
         if (extensions == null) {
             return;
         }
-        if (extensions.containsKey(APIConstants.X_WSO2_AUTH_HEADER)) {
-            extensions.remove(APIConstants.X_WSO2_AUTH_HEADER);
-        }
-        if (extensions.containsKey(APIConstants.X_THROTTLING_TIER)) {
-            extensions.remove(APIConstants.X_THROTTLING_TIER);
-        }
-        if (extensions.containsKey(APIConstants.X_WSO2_CORS)) {
-            extensions.remove(APIConstants.X_WSO2_CORS);
-        }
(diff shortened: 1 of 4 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
```diff
@@ -1096,27 +1096,16 @@ public static void removePublisherSpecificInfo(Map<String, Object> extensions) {
...
-        if (extensions.containsKey(APIConstants.X_WSO2_SANDBOX_ENDPOINTS)) {
-            extensions.remove(APIConstants.X_WSO2_SANDBOX_ENDPOINTS);
-        }
-        if (extensions.containsKey(APIConstants.X_WSO2_BASEPATH)) {
-            extensions.remove(APIConstants.X_WSO2_BASEPATH);
-        }
-        if (extensions.containsKey(APIConstants.X_WSO2_TRANSPORTS)) {
-            extensions.remove(APIConstants.X_WSO2_TRANSPORTS);
-        }
+        extensions.remove(APIConstants.X_WSO2_AUTH_HEADER);
```
ruks: can you please confirm extensions.remove() not throw NPE if the key is null. Looks like it does not throw NPE.

AmaliMatharaarachchi: if the element is not contained inside the array, it will not do anything. Refer https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#remove(java.lang.Object)

PR #8335, opened 2020-03-11
Fixing subscriptions are auto assigning issue to new versions

HiranyaKavishani:
Description:
If the API is moving across the environments, it has no idea whether this is a new API or API that created from an existing version. In the next environment, it might not have the same applications and etc. Therefore, subscriptions shouldn't be migrated to next environments.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java
@@ -434,7 +434,8 @@ public final class APIImportUtil {
             // Change API lifecycle if state transition is required
             if (StringUtils.isNotEmpty(lifecycleAction)) {
                 log.info("Changing lifecycle from " + currentStatus + " to " + targetStatus);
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java
```diff
@@ -434,6 +434,7 @@ public static void importAPI(String pathToArchive, String currentUser, boolean i
             // Change API lifecycle if state transition is required
             if (StringUtils.isNotEmpty(lifecycleAction)) {
                 log.info("Changing lifecycle from " + currentStatus + " to " + targetStatus);
+                apiProvider.changeAPILCCheckListItems(importedApi.getId(), 1, true);
```
malinthaprasan: what is that magic number 1 ? :P

HiranyaKavishani: It was used to refer the first attribute  of the "CheckListItems"

HiranyaKavishani: Done with 1291d974f0ea49f2eb61114428f473fb3811d844

PR #8378, opened 2020-03-20
Fix for #7696

TanyaM:
This PR resolves the issue of resource cache invalidation cluster message not being sent to the worker nodes when updating an API resource.


TanyaM:
Please don't merge until 3.1.0-GA

TanyaM:
This fix is not required in 3.1.0 as we no longer use cluster message for resource cache invalidation.

PR #8387, opened 2020-03-28
Modify exceptions when importing/exporting API thumbnail icon from UI

wasuradananjith:
## Purpose
Change the exception handling when importing/exporting API thumbnail icon from UI. 
Fixes: https://github.com/wso2/product-apim-tooling/issues/216

## Goals
- Change the generic exception to specific ones.
- Justify the need more using comments

## Approach
- String mimeType = URLConnection.guessContentTypeFromName(fileName) decides the mimetype for image extensions such as .jpg, .jpeg, .png, .gif, .bmp. But if the file type is .json URLConnection returns null. 
- So, used JsonParser which will identify whether it is a valid .json file or not.
- If it is not an image or not a JSON file then the exception should be handled without breaking the flow. (In here the thumbnail will not get imported, but the API will get successfully imported without the thumbnail) 

## Related PRs
- https://github.com/wso2/carbon-apimgt/pull/8384
- https://github.com/wso2-support/carbon-apimgt/pull/1929

## Test environment
-  JDK 1.8.0_241
- Ubuntu 18.04.4 LTS

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java
@@ -19,7 +19,12 @@
 package org.wso2.carbon.apimgt.impl.importexport.utils;
 
 import com.google.common.collect.Sets;
-import com.google.gson.*;
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.JsonParseException;
 import com.google.gson.reflect.TypeToken;
 
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/importexport/utils/APIImportUtil.java
```diff
@@ -531,8 +532,13 @@ private static void updateAPIWithThumbnail(File imageFile, API importedApi, APIP
                 // Check whether the icon is in .json format (UI icons are stored as .json)
                 new JsonParser().parse(new FileReader(imageFile));
                 mimeType = APIConstants.APPLICATION_JSON_MEDIA_TYPE;
-            } catch (Exception e) {
-                log.error("Failed to read the file. ", e);
+            } catch (JsonParseException e) {
```
malinthaprasan: Shall we use `} catch (JsonParseException | FileNotFoundException e) {`

wasuradananjith: I have used two exceptions to display two different logs.

PR #8388, opened 2020-03-30
Re-arranging Try-out console and adding generate token

kavishkafernando:
Fixes https://github.com/wso2/product-apim/issues/7581

Screen Shot 2020-03-30 at 5 33 27 PM


erangatl:
"Generate Test Key" is not the primary action of this UI, so shall we use a secondary button?
@tmkasun @chanaka3d WDYT?

Also, you can use an existing key as well, it's not intuitive in the UI. Shall we at least give a tooltip?

https://xd.adobe.com/view/4566d6eb-af2f-49bf-522d-4bc7a3e9a541-3357/?fullscreen

Screen Shot 2020-03-30 at 4 34 48 PM

kavishkafernando:
+1 @erangatl. I have updated the PR with the new screenshot and added the tooltip as well.

PR #8391, opened 2020-03-31
Errorous cast of APIs & APIProducts into Set<API>

slawomirkisiel:
apiProvider.searchPaginatedAPIs returns APIs and APIProducts in single Set. Casting this set to Set of API cause error in org.wso2.carbon.apimgt.rest.api.publisher.utils.mappings.APIMappingUtil.fromAPIListToDTO method because APIProduct is not extended on API class. 
Exception than occurs: java.lang.ClassCastException: org.wso2.carbon.apimgt.api.model.APIProduct cannot be cast to org.wso2.carbon.apimgt.api.model.API


slawomirkisiel:
I suppose it would be much better to change behaviour of apiProvider.searchPaginatedAPIs method but did not have time to search for this solution.

malinthaprasan:
Hi @slawomirkisiel .. Thank you very much for your contribution. However, I believe @isharac fixed this by avoiding returning APIProducts from searchPaginatedAPIs method..

isharac:
This is fixed in APIM 3.0.0 as a wum update and the fix is available in APIM 3.1.0

malinthaprasan:
Thanks, closed the PR since the issue is fixed.

malinthaprasan:
Fixed commit https://github.com/wso2/carbon-apimgt/commit/1feca1937de1a29156eb8e8709e22eac09c3b428

PR #8302, opened 2020-03-03
Adding passive mode for DevPortal

Senthuran100:
Git issue : https://github.com/wso2/product-apim/issues/4171

PR #8415, opened 2020-04-17
Handle OIDC Error log when ClientId is null

hisanhunais:
## Purpose
Fixes: https://github.com/wso2/product-apim/issues/7874

## Approach
Render the session checking iframe only when clientId is not null.

## Test environment
JDK 1.8.0_241
Ubuntu 18.04.4 LTS


tmkasun:
I think we need to do the check session in ananymous mode as well, As discussed offiline with @malinthaprasan  This check session action might required to do the automatic login in developer portal (passive login).

 @senthuran100 Has worked on this passive login feature 
https://github.com/wso2/carbon-apimgt/pull/8302

@senthuran100 Can you give some input on this?

tmkasun:
Update: Had an offline chat with @Senthuran100 , @hisanhunais and decide to park this change until we find a better alternative method to do the session check when the user hasn't logged in

bhathiya:
I'm closing the PR based on the last comment from @tmkasun

PR #8449, opened 2020-04-30
Fix token generated from application not working in tryout console

kavishkafernando:
Fix token generated from application not working in tryout console


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/devportal/source/src/app/components/Apis/Details/ApiConsole/TryOutController.jsx
```diff
@@ -332,6 +332,17 @@ function TryOutController(props) {
             case 'password':
                 setPassword(value);
                 break;
+            case 'accessToken':
+                if (securitySchemeType === 'API-KEY' && selectedKeyType === 'PRODUCTION') {
```
tmkasun, line 336: If these are constants and used in some other paces, Better to move them to a constant variable and refer to the variable when need the value

PR #8452, opened 2020-05-01
Listing base component improvements

chanaka3d:
Updated Listing page as a base component, making it possible to reuse over different components.

Handled empty response from backend case.

Allows a call back to trigger when the edit or create dialog opens.


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/admin_new/source/src/app/components/AdminPages/Microgateways/AddEdit.jsx
```diff
@@ -112,7 +112,13 @@ function AddEdit({
         }
         return errorText;
     };
-
+    const dialogOpenCallback = () => {
+        // We can do an API call when we are in the editing mode
```
chanaka3d, line 116: This is the place to do a backend call and fetch data @kavishkafernando

PR #8455, opened 2020-05-03
Publish analytics events for throttling scenario

VirajSalaka:
With the following PR,

- Throttle events would be published to the analytics.
- Bug Fix: Closed WebSocket frame will not increase the throttle counter. As well as it will not be published to the analytics under API Usage.
- Request events will be published to the analytics for both inbound and outbound frames (for the connection between client and the gateway)

public issue: wso2/product-apim#7879


Arshardh:
We are holding off this PR till the end of this week as there are sone further tests to be carried out

VirajSalaka:
> We are holding off this PR till the end of this week as there are some further tests to be carried out
Testing is completed. The PR can be merged.

PR #8456, opened 2020-05-03
Rename throttle data service 

isharac:
throttle data service is renamed to internal data service.


isharac:
Re created https://github.com/wso2/carbon-apimgt/pull/8457

PR #8428, opened 2020-04-23
Flexibility to specify any scope scheme apart from "default" scheme when importing an OpenAPI definition

Chamindu36:
## Purpose
Providing Flexibility to specify any scope scheme apart from "default" scheme when importing an OpenAPI definition through API publisher or through API Controller and assign those scopes into resources correctly.

## Goals
Currently, the API publisher does not support multiple OAuth2 scope schemes and support only 'default' scope schemes. But when a customer brings their own API definition as a swagger file there can be OAuth2 scope schemes that are not written in the 'default' scope scheme. In that case, currently, those scopes do not get exposed or used in API publisher to generate tokens or perform any other operations. Since Microgateway supports these types of different OAuth2 scope schemes the APIM product should provide support for this particular scenario. 
Fixes https://github.com/wso2/product-apim-tooling/issues/124 and https://github.com/wso2/product-apim/issues/7648

## Approach
Change the swagger file when importing into APIM publisher - 

In this approach what we are going to do is copy or replace the Oauth2 scope definitions that are not default scheme into default schemes. In this approach, the swagger definition can be replaced or merged with the changes. And that changed swagger definition will be the subject to OAS3 parser or OAS2 parser. Then the API exposing process will go on as in the current procedure. 

 

_Disadvantage_ -  In this approach since we make a change to the swagger file when the support of multiple scope schemes implemented properly this fix will be troublesome.
_Advantage_ - For the moment this approach will provide a fix to the use-case that we discussed above without breaking or interrupting the existing chain of processes. 

> _Eg: External Swagger Scheme_

7QXViHC9_73xS3OCtF-X1pbP3l19eFTWEyBUsmFToa43Q8yO1OSVcRTlr4uAs368dz35aMvQDQWxPLjFYVGxIt9Xww2uPvxfneUfedBa7Av_vRXlbaGdx5zBKyXdvimqkOTAhkRB

> _The way resources use these security schemes originally._

9O0GVP37C-bTfx2hPdncy7WK7ICqgq5DB-d0NBRPoqufklclhXLMNcZK3K8BYzRODf4OpDeuNbpBjXJVDIRpesap1h-fvIj5R2ZgwEWaChSiw2yIiYAeIc5gC2GNq7OsT7X6jzG2

>  _After the change._

[...]

PR #8461, opened 2020-05-04
Flexibility to specify any scope scheme apart from "default" scheme when importing an OpenAPI definition

Chamindu36:
## Purpose
Providing Flexibility to specify any scope scheme apart from "default" scheme when importing an OpenAPI definition through API publisher or through API Controller and assign those scopes into resources correctly.

## Goals
Currently, the API publisher does not support multiple OAuth2 scope schemes and support only 'default' scope schemes. But when a customer brings their own API definition as a swagger file there can be OAuth2 scope schemes that are not written in the 'default' scope scheme. In that case, currently, those scopes do not get exposed or used in API publisher to generate tokens or perform any other operations. Since Microgateway supports these types of different OAuth2 scope schemes the APIM product should provide support for this particular scenario. 
Fixes https://github.com/wso2/product-apim-tooling/issues/124 and https://github.com/wso2/product-apim/issues/7648

## Approach
Change the swagger file when importing into APIM publisher - 

In this approach what we are going to do is copy or replace the Oauth2 scope definitions that are not default scheme into default schemes. In this approach, the swagger definition can be replaced or merged with the changes. And that changed swagger definition will be the subject to OAS3 parser or OAS2 parser. Then the API exposing process will go on as in the current procedure. 

 

_Disadvantage_ -  In this approach since we make a change to the swagger file when the support of multiple scope schemes implemented properly this fix will be troublesome.
_Advantage_ - For the moment this approach will provide a fix to the use-case that we discussed above without breaking or interrupting the existing chain of processes. 

> _Eg: External Swagger Scheme_

7QXViHC9_73xS3OCtF-X1pbP3l19eFTWEyBUsmFToa43Q8yO1OSVcRTlr4uAs368dz35aMvQDQWxPLjFYVGxIt9Xww2uPvxfneUfedBa7Av_vRXlbaGdx5zBKyXdvimqkOTAhkRB

> _The way resources use these security schemes originally._

9O0GVP37C-bTfx2hPdncy7WK7ICqgq5DB-d0NBRPoqufklclhXLMNcZK3K8BYzRODf4OpDeuNbpBjXJVDIRpesap1h-fvIj5R2ZgwEWaChSiw2yIiYAeIc5gC2GNq7OsT7X6jzG2

>  _After the change._

mLHyIozkBiIjO2dvuuEvr9hh_E8_R_2YC0mHMmjw5l8XRb1Tf2EZ6Pcm8NIXdetaoeT54qAubusCKKJYCmFsLoEKcyLYhVxqEdSogTwusRR_crss7equnZm1gF8EwvN4QYsRIc8I

> _The way resources use these security schemes after the fix of the issue._

image

## User stories
* Importing external swagger 2 definitions through API CTL.
* Importing external swagger 3 definitions through API CTL.
* Importing external swagger 2 definitions through API publisher UI or using curl commands.
* Importing external swagger 3 definitions through API  publisher UI or using curl commands.
* Importing Swagger definitions with additional properties.

## Documentation
Doc has Impact
https://github.com/wso2/docs-apim/pull/1075
https://github.com/wso2/docs-apim/pull/1074

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
@@ -197,4 +197,13 @@ public abstract class APIDefinition {
     public abstract String getOASDefinitionWithTierContentAwareProperty(String oasDefinition,
             List<String> contentAwareTiersList, String apiLevelTier) throws APIManagementException;
 
+    /**
+     * This method changes the URI templates from the API definition as it support different schemes
+     * @param resourceConfigsJSON json String of oasDefinition
+     * @throws APIManagementException throws if an error occurred
+     * @return URI templates
+     */
+    public abstract String processOtherSchemeScopes(String resourceConfigsJSON)
+            throws APIManagementException;
+
 }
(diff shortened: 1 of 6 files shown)


malinthaprasan:
Reviews were done in https://github.com/wso2/carbon-apimgt/pull/8428

PR #8469, opened 2020-05-06
Add roles and scopes to support Endpoint Registry functionalities

ChamodDamitha:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/resources/tenant/tenant-conf.json
```diff
@@ -196,6 +196,22 @@
       {
         "Name": "apim:admin_operations",
         "Roles": "admin"
+      },
+      {
+        "Name": "registry:view",
+        "Roles": "admin,Internal/registry_admin,Internal/registry_user,Internal/registry_browser"
```
fazlan-nazeem, line 202: Shall we add Internal/creator,Internal/publisher roles as well? Those users should be able to view the registries created because it will be used in the publisher portal.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/resources/tenant/tenant-conf.json
```diff
@@ -196,6 +196,22 @@
...
+      },
+      {
+        "Name": "registry:write",
+        "Roles": "admin,Internal/registry_admin"
+      },
+      {
+        "Name": "registry:entry_view",
```
fazlan-nazeem, line 209: The same roles mentioned in the previous comment should be applied here as well.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/resources/tenant/tenant-conf.json
ChamodDamitha: fixed

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/resources/tenant/tenant-conf.json
ChamodDamitha: fixed

PR #8470, opened 2020-05-06
Adding generated files that missed in the PR #8300

isuruh15:
generated files were missed in the PR #8300. This PR is to add that file.


ChamodDamitha:
Check if the changes to below doc file should be committed as well

docs/apidocs/publisher/v1/models/OpenAPIDefinitionValidationResponse_info.html

PR #8481, opened 2020-05-11
[Admin Portal Revamp] Breadcrumbs implementation

chathuranga95:
Breadcrumbs component is now implemented to work with the RouteMenuMapping and behave dynamically with page navigation.

breadcrumbs screenshot

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/admin_new/source/src/app/components/Base/Header/Breadcrumbs.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/admin_new/source/src/app/components/Base/Header/Breadcrumbs.jsx
@@ -33,7 +33,7 @@ function getPageIDs(routeMenuMapping, currentPath, pageIDs = []) {
 
 /**
  * Render breadcrumb component
- * * @param {JSON} props .
+ * @param {JSON} props .
  * @returns {JSX} breadcrumbs.
  */
 function Breadcrumbs(props) {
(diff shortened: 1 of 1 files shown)


chanaka3d:
I don't see history.listen callback registration. How does this component update every time a route change?

chathuranga95:
> I don't see history.listen callback registration. How does this component update every time a route change?

It uses the browser's `location`. Therefore, the component updates correctly.

However, I have now changed it to history.location. So the code is much cleaner.

chathuranga95:
Please note that with commit: https://github.com/wso2/carbon-apimgt/pull/8481/commits/c1eb8f4ee28840b4aef453632ed7e992162e8290, Add/Edit url patterns will be supported.

On `RouteMenuMapping` component, please add children for identifying the pages. Only `id` and `path` are needed. Set `path` as a regex when uuid's are present.

Following is an example for a scenario where there is a separate editing page for Blacklist Policies.

```
            {
                id: intl.formatMessage({
                    id: 'Base.RouteMenuMapping.blacklisted.items',
                    defaultMessage: 'Blacklist Policies',
                }),
                path: '/throttling/blacklisted',
                component: () => <DemoTable />,
                icon: <PhonelinkSetupIcon />,
                children: [
                    {
                        id: intl.formatMessage({
                            id: 'Base.RouteMenuMapping.blacklisted.items.Editing',
                            defaultMessage: 'Edit Blacklist Policy',
                        }),
                        path: '/throttling/blacklisted/(.*?)/(edit$)',
                    },
                ],
            },
```

PR #8491, opened 2020-05-12
Add deleteEndpointRegistry resource

ChamodDamitha:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/EndpointRegistryImpl.java
```diff
@@ -78,6 +78,21 @@ public EndpointRegistryInfo getEndpointRegistryByUUID(String registryId) throws
...
     }
 
+    /**
+     * Deletes an Endpoint Registry
+     *
+     * @param registryId Registry Identifier(UUID)
+     * @throws APIManagementException if failed to delete the Endpoint Registry
+     */
+    public void deleteEndpointRegistry(String registryId) throws APIManagementException {
+        EndpointRegistryInfo endpointRegistryInfo = apiMgtDAO.getEndpointRegistryByUUID(registryId);
```
fazlan-nazeem, line 88: Shall we do this in the REST API level, so that we can throw the exception from there itself? I felt the readability is a bit reduced when we throw it from here and catch it from the parent and then handle the exception again.

fazlan-nazeem: if you agree with this approach, you may change the same in other places as well.

ChamodDamitha: Fixed with the commit **dd74437**

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
```diff
@@ -14789,6 +14792,32 @@ public EndpointRegistryInfo getEndpointRegistryByUUID(String registryId) throws
...
+    /**
+     * Deletes an Endpoint Registry
+     *
+     * @param registryUUID Registry Identifier(UUID)
+     * @param registryId Registry Identifier
+     * @throws APIManagementException if failed to delete the Endpoint Registry
+     */
+    public void deleteEndpointRegistry(String registryUUID, int registryId) throws APIManagementException {
+        String deleteRegQuery = SQLConstants.DELETE_ENDPOINT_REGISTRY_SQL;
+        String deleteRegEntryQuery = SQLConstants.DELETE_ENDPOINT_REGISTRY_ENTRY_BY_REGISTRY_ID_SQL;
```
fazlan-nazeem, line 14804: We don't need this right? since we have 'delete on cascade' ?

ChamodDamitha: Ah yes. I ll update it

ChamodDamitha: Fixed in **027f192**

PR #8502, opened 2020-05-13
Add UpdateEndpointRegistryEntry and DeleteEndpointRegistryEntry resources

ChamodDamitha:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/EndpointRegistryImpl.java
```diff
@@ -155,12 +155,39 @@ public EndpointRegistryEntry getEndpointRegistryEntryByUUID(String registryId, S
...
+    /**
+     * Updates Registry Entry
+     *
+     * @param registryEntry EndpointRegistryEntry
+     * @throws APIManagementException if failed to update EndpointRegistryEntry
+     */
+    public void updateEndpointRegistryEntry(EndpointRegistryEntry registryEntry) throws APIManagementException {
+        EndpointRegistryEntry endpointRegistryEntry =
+                apiMgtDAO.getEndpointRegistryEntryByUUID(registryEntry.getEntryId());
+        if (!endpointRegistryEntry.getName().equals(registryEntry.getName()) &&
```
fazlan-nazeem, line 173: I think we decided to allow to edit the entry name as well.

ChamodDamitha: that is right. this check is only to avoid an existing entry name

components/apimgt/org.wso2.carbon.apimgt.rest.api.endpoint.registry/src/main/java/org/wso2/carbon/apimgt/rest/api/endpoint/registry/impl/RegistriesApiServiceImpl.java
```diff
@@ -254,13 +254,68 @@ public Response getRegistryEntryByUuid(String registryId, String entryId, Messag
...
+            }
+            EndpointRegistryEntry endpointRegistryEntry =
+                    registryProvider.getEndpointRegistryEntryByUUID(registryId, entryId);
+            if (endpointRegistryEntry == null) {
+                RestApiUtil.handleResourceNotFoundError("Endpoint registry entry with the id: " + entryId +
+                        " is not found", log);
+            }
 
-        return Response.ok().entity(registryEntry).build();
+            ResourceFile definitionFile = new ResourceFile(definitionFileInputStream, definitionFileDetail
```
fazlan-nazeem, line 276: can any null pointers occur here when we don't upload a file?

ChamodDamitha: yeah it occurs. I will fix it. If there is no file attached, still we should update the other details right?

fazlan-nazeem: yeah, let's keep the existing blob in the DB as it is and update the rest in that case.

fazlan-nazeem: we should also think of the use case where the user wants to remove the blob from the DB and update. in that sense, we should set an empty blob if no file is attached.

fazlan-nazeem: since there would be no way to achieve the above if we don't update the blob, let's go ahead and set an empty blob if no file is attached. wdyt

ChamodDamitha: Fixed

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
```diff
@@ -15012,6 +15012,53 @@ public String addEndpointRegistryEntry(EndpointRegistryEntry registryEntry) thro
...
+            connection.setAutoCommit(false);
+            ps.setString(1, registryEntry.getName());
+            ps.setString(2, registryEntry.getServiceURL());
+            ps.setString(3, registryEntry.getDefinitionType());
+            ps.setString(4, registryEntry.getDefinitionURL());
+            ps.setString(5, registryEntry.getMetaData());
+            ps.setString(6, registryEntry.getServiceType());
+            ps.setBlob(7, registryEntry.getEndpointDefinition().getContent());
+            ps.setString(8, registryEntry.getEntryId());
+            ps.executeUpdate();
```
dushaniw, line 15034: shouldn't we catch the exception in executeUpdate and rollback the connection explicitly. 

See https://stackoverflow.com/questions/3160756/in-jdbc-when-autocommit-is-false-and-no-explicit-savepoints-have-been-set-is-i

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
```diff
@@ -15012,6 +15012,53 @@ public String addEndpointRegistryEntry(EndpointRegistryEntry registryEntry) thro
...
+     */
+    public void deleteEndpointRegistryEntry(String entryId) throws APIManagementException {
+        String query = SQLConstants.DELETE_ENDPOINT_REGISTRY_ENTRY_SQL;
+
+        try (Connection connection = APIMgtDBUtil.getConnection();
+             PreparedStatement statement = connection.prepareStatement(query)
+        ) {
+            connection.setAutoCommit(false);
+            statement.setString(1, entryId);
+            statement.execute();
```
dushaniw, line 15055: same here, IMO we should rollback the connection if statement.execute fails

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
ChamodDamitha: Fixed in https://github.com/wso2-enterprise/service-catalog/pull/73

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
ChamodDamitha: Fixed in https://github.com/wso2-enterprise/service-catalog/pull/73

PR #8503, opened 2020-05-14
Get tenant aware username when the EnableEmailUserName

amalkasubasinghe:


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/devportal/services/login/introspect.jag
```diff
@@ -46,6 +46,10 @@
     // For more info about response.xhr object refer http://jaggeryjs.org/documentation.jag?api=xhr
     
     if (introspectResult.xhr.status === 200) {
+        var isEnableEmailUserName = appUtils.isEnableEmailUserName();
+        if (isEnableEmailUserName || introspectResult.data.username.endsWith('@carbon.super')) {
+            introspectResult.data.username = introspectResult.data.username.replace('@carbon.super', '');
+        }
```
malinthaprasan, line 52: We shouldn't remove "carbon.super" if EnableEmailUserName settings is disabled but username is an email.

Ex1:
say username is amalka@gmail.com@carbon.super, then 
1. if isEnableEmailUserName is true => returned username is amalka@gmail.com
2. if isEnableEmailUserName is false => return amalka@gmail.com@carbon.super

Ex2: 
say username is malintha@carbon.super, then 
1. if isEnableEmailUserName is true => returned username is malintha
2. if isEnableEmailUserName is false => return malintha (same)

Basically you need to move the same logic in AuthManager.jsx with additional check for EnableEmailUserName. 
```
const count = (data.username.match(/@/g) || []).length;
if (isEnableEmailUserName || (data.username.endsWith('@carbon.super') && count <= 1)) {
    introspectResult.data.username = introspectResult.data.username.replace('@carbon.super', '');
}
```

PR #8510, opened 2020-05-15
Add validation for existing registry name

fazlan-nazeem:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/EndpointRegistryImpl.java
```diff
@@ -191,10 +191,29 @@ public void deleteEndpointRegistryEntry(String entryId) throws APIManagementExce
...
+            endpointRegistryInfo) throws APIManagementException {
+
+        String tenantDomain = MultitenantUtils
+                .getTenantDomain(APIUtil.replaceEmailDomainBack(endpointRegistryInfo.getOwner()));
+        int tenantId;
+        try {
+            tenantId = ServiceReferenceHolder.getInstance().getRealmService().getTenantManager()
+                    .getTenantId(tenantDomain);
+            // if another registry with the updated name already exists, fail the operation.
+            if (!registryName.equals(endpointRegistryInfo.getName()) &&
```
vithu30, line 204: I think when the registry name already exists in DB, it will throw 500, internal server error response. If so we need to change it.

PR #8525, opened 2020-05-18
Fix schema validation when property "json-eval($.)" is used

renuka-fernando:
#### Purpose
Replace Jackson with GSON without changing the logic of SchemaValidator.

#### Related Changes (updated synapse)
https://github.com/wso2-support/wso2-synapse/pull/787: Set GsonJsonProvider as the default config for Jayway JsonPath when init the server.

fix wso2/product-apim#7949
fix https://github.com/wso2/product-apim/issues/8056

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1751,7 +1751,7 @@
         <imp.package.version.osgi.framework>[1.6.0, 2.0.0)</imp.package.version.osgi.framework>
 
         <!-- Misc Versions -->
-	    <synapse.version>v2.1.7-wso2v165</synapse.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1751,7 +1751,7 @@
         <imp.package.version.osgi.framework>[1.6.0, 2.0.0)</imp.package.version.osgi.framework>
 
         <!-- Misc Versions -->
-	<synapse.version>2.1.7-wso2v147</synapse.version>
+	    <synapse.version>2.1.7-wso2v165</synapse.version>
```
dinusha92, line 1754: Let's remove this additional space.

PR #8527, opened 2020-05-18
Feature opa

SasankaMallikarachchi:


Arshardh:
Do we need files like WSO2AM_DB.mv.db to be pushed?

bhathiya:
@SasankaMallikarachchi can you please fix the review comments soon?

PR #8616, opened 2020-06-01
Moving Blacklist policies to a new page

kavishkafernando:
This PR includes the following changes

- Moving Blacklist policies to a new page
- Put the "The following sample siddhi query ..." text into an infobox
- Add key template helper text
- Validate Key Templates from UI

**Define policy:**
Screen Shot 2020-06-01 at 2 33 58 PM

**Edit policy:**
Screen Shot 2020-06-01 at 2 34 24 PM

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin_new/source/src/app/components/Throttling/Custom/AddEdit.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin_new/source/src/app/components/Throttling/Custom/AddEdit.jsx
@@ -81,8 +81,7 @@ const formattedSampleSiddhiQuery = sqlFormatter.format(sampleSiddhiQuery);
  * @param {JSON} state The second number.
  * @returns {Promise}
  */
-function reducer(state, newValue) {
-    const { field, value } = newValue;
+function reducer(state, { field, value }) {
     switch (field) {
         case 'policyName':
             return { ...state, [field]: value };
@@ -92,8 +91,10 @@ function reducer(state, newValue) {
             return { ...state, [field]: value };
         case 'siddhiQuery':
             return { ...state, [field]: value };
+        case 'editDetails':
+            return value;
         default:
-            return newValue;
+            return state;
     }
 }
 
@@ -132,7 +133,7 @@ function AddEdit(props) {
                     keyTemplate: result.body.keyTemplate,
                     siddhiQuery: formattedSiddhiQuery,
                 };
-                dispatch(editState);
+                dispatch({ field: 'editDetails', value: editState });
             });
         }
         setInitialState({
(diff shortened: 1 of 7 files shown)


features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin_new/source/src/app/components/Throttling/Custom/AddEdit.jsx
```diff
@@ -100,8 +120,22 @@ function AddEdit(props) {
     const [validationError, setValidationError] = useState([]);
     const [editMode, setIsEditMode] = useState(false);
     const restApi = new API();
+    const { policyId } = props.match.params;
 
     useEffect(() => {
+        if (policyId !== 'create') {
+            setIsEditMode(true);
```
tmkasun, line 127: I think we can set this state at the state initialization time 
i:e
```javascript
const [editMode, setIsEditMode] = useState(policyId !== 'create');
```

kavishkafernando: Fixed with `6317b1b`

features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin_new/source/src/app/components/Throttling/Custom/AddEdit.jsx
```diff
@@ -100,8 +120,22 @@ function AddEdit(props) {
...
+            restApi.customPolicyGet(policyId).then((result) => {
+                const formattedSiddhiQuery = sqlFormatter.format(result.body.siddhiQuery);
+                const editState = {
+                    policyName: result.body.policyName,
+                    description: result.body.description,
+                    keyTemplate: result.body.keyTemplate,
+                    siddhiQuery: formattedSiddhiQuery,
+                };
+                dispatch(editState);
```
tmkasun, line 135:
```javascript
                dispatch(editState);
```
Though this `dispatch` call does the job here, It's better to follow the action dispatching methodology which is described in [Flux pattern](https://facebook.github.io/flux/docs/in-depth-overview#actions).
In the Flux pattern Object that we pass into the `dispatcher` called an action, Typically there are two elements in an action type and payload, and we derive the next state by switching the executing by the action type and incoming event's payload.

This pattern can be seen in [React documentation examples](https://reactjs.org/docs/hooks-reference.html#lazy-initialization) as well.

Shall we follow a similar pattern here as well?

kavishkafernando: Fixed with `4e14418`

PR #8620, opened 2020-06-01
Refractor service catalog scope names and role names

ChamodDamitha:
Scopes

> service_catalog:view
> service_catalog:write
> service_catalog:entry_view
> service_catalog:entry_write

Roles

> Internal/service_catalog_admin
> Internal/service_catalog_user
> Internal/service_catalog_viewer


fazlan-nazeem:
difference between service_catalog_browser and service_catalog_user is not that intuitive, I guess. Any better name suggestions?

ChamodDamitha:
> difference between service_catalog_browser and service_catalog_user is not that intuitive, I guess. Any better name suggestions?

How about **service_catalog_writer** and **service_catalog_viewer/service_catalog_browser** ?

fazlan-nazeem:
service_catalog_user and service_catalog_viewer seem to be better since the viewer suffix will indicate it's a read-only role.

ChamodDamitha:
> service_catalog_user and service_catalog_viewer seem to be better since the viewer suffix will indicate it's a read-only role.

Updated

PR #8624, opened 2020-06-02
Add mapping between APIs and Servcie catalog entries

vithu30:
When API is created by selecting an endpoint from service catalog, an entry will be added to table called `AM_API_ENDPOINT_REG_ENTRY_MAPPING`.


features/apimgt/org.wso2.carbon.apimgt.core.feature/src/main/resources/sql/h2.sql
```diff
@@ -1863,4 +1863,12 @@ CREATE TABLE IF NOT EXISTS ENDPOINT_REG_ENTRY (
   FOREIGN KEY (REG_ID) REFERENCES ENDPOINT_REG (ID) ON DELETE CASCADE
 );
 
+CREATE TABLE IF NOT EXISTS AM_API_ENDPOINT_REG_ENTRY_MAPPING (
+    API_ID INTEGER NOT NULL,
+    REG_NAME VARCHAR(255) NOT NULL,
+    ENTRY_NAME VARCHAR(255) NOT NULL,
+    ENTRY_VERSION VARCHAR(30),
+    UNIQUE (API_ID, REG_NAME, ENTRY_NAME, ENTRY_VERSION),
```
fazlan-nazeem, line 1871: Going forward, production endpoint and sandbox endpoint can be attached to two different registry entries. So we cannot add this unique constraint, as there can be two rows with the same combination of the same entry is used for both endoints.

vithu30: Yes, will introduce the support for adding separate entries for production and sandbox endpoints in next PR.

PR #8644, opened 2020-06-03
Add mapping between Registry Entries and associated APIs

vithu30:
Following changes are added:
- Production and sandbox URL fields removed from Registry Entry swagger model
- Renamed Create Registry Entry  resource path 
- Renamed getEndpointDefinition resource
- Response of Delete operation is changed


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
```diff
@@ -10687,4 +10687,32 @@ public static String getClaimDisplayName(String claimURI, String subscriber) thr
         }
         return scopeToKeyMap;
     }
+
+    /**
+     * Extract Endpoint Registry Entry information
```
fazlan-nazeem, line 10692: shall we add this to an endpoint registry util class. if non exists, create one.

fazlan-nazeem: ignore above

PR #8645, opened 2020-06-03
Fix service-catalog/entry displayName update issue and Refractor RegistriesApiServiceImpl

ChamodDamitha:
- Refractor **RegistriesApiServiceImpl** class
  - Introduced private methods to check the existence of Registry/Entry
  - Moved endpoint definition retrieving and validation to a new method

- Fix service-catalog/entry **displayName** update issue
  - With new implementation, if a user does not send displayName in the payload(during an update resource call), the current displayName will not be updated

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/endpoint/registry/impl/EndpointRegistryImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/endpoint/registry/impl/EndpointRegistryImpl.java
@@ -66,6 +66,12 @@ public class EndpointRegistryImpl implements EndpointRegistry {
         try {
             tenantId = ServiceReferenceHolder.getInstance().getRealmService().getTenantManager()
                     .getTenantId(tenantDomain);
+
+            // If the displayName is empty, set the registryName as the displayName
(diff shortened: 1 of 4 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/endpoint/registry/impl/EndpointRegistryImpl.java
```diff
@@ -190,6 +196,11 @@ public EndpointRegistryEntry getEndpointRegistryEntryByUUID(String registryId, S
      */
     public String addEndpointRegistryEntry(EndpointRegistryEntry registryEntry) throws EndpointRegistryException {
 
+        // If the displayName is empty, set the entryName as the displayName
+        if (StringUtils.isEmpty(registryEntry.getDisplayName())) {
```
fazlan-nazeem, line 200: since we have done the same logic in multiple places, shall we move this logic to some common place, like mappingUtils?

ChamodDamitha: Fixed in https://github.com/wso2/carbon-apimgt/pull/8645/commits/56072355e9225b17e47a17ba49e7d6fbf0c6539c

PR #8563, opened 2020-05-25
Support for In Memory Subscription validation

isharac:
This PR provides the support for In Memory Subscription Validation in the gateway.
New REST APIs are introduced to load meta data from the db.

Ex:
**/apis**
curl --location  --request GET 'https://localhost:9443/internal/data/v1/apis' -kv -header 'X-WSO2-Tenant: abc.com'
{"count":1,"list":[{"apiId":1,"provider":"adminfoo@abc.com","name":null,"version":"1.0.0","context":"/t/abc.com/pizzashack/1.0.0","policy":null,"urlMaping":[]}],"pagination":null}

curl --location  --request GET 'https://localhost:9443/internal/data/v1/apis?vesion=1.0.0&context=/t/abc.com/pizzashack/1.0.0' -kv
{"count":1,"list":[{"apiId":1,"provider":"adminfoo@abc.com","name":null,"version":"1.0.0","context":"/t/abc.com/pizzashack/1.0.0","policy":null,"urlMaping":[]}],"pagination":null}

curl --location  --request GET 'https://localhost:9443/internal/data/v1/apis’
{"count":2,"list":[{"id":3,"name":null,"subName":"adminfoo@abc.com","policy":"Unlimited","tokenType":"JWT","groupIds":[],"attributes":[]},{"id":4,"name":null,"subName":"adminfoo@abc.com","policy":"10PerMin","tokenType":"JWT","groupIds":[],"attributes":[]}],"pagination":null}

**/subscriptions**
curl --location  --request GET 'https://localhost:9443/internal/data/v1/subscriptions'
{"count":1,"list":[{"subscriptionId":1,"policyId":"Unlimited","apiId":1,"appId":4,"subscriptionState":"UNBLOCKED"}],"pagination":null}

curl --location  --request GET 'https://localhost:9443/internal/data/v1/subscriptions?apiId=1&appId=4' -kv
{"count":1,"list":[{"subscriptionId":1,"policyId":"Unlimited","apiId":1,"appId":4,"subscriptionState":"UNBLOCKED"}],"pagination":null}

curl --location  --request GET 'https://localhost:9443/internal/data/v1/subscriptions' -kv --header 'X-WSO2-Tenant: abc.com'
{"count":1,"list":[{"subscriptionId":1,"policyId":"Unlimited","apiId":1,"appId":4,"subscriptionState":"UNBLOCKED"}],"pagination":null}

**/applictaions**
curl --location  --request GET 'https://localhost:9443/internal/data/v1/applications’
[...]

PR #8673, opened 2020-06-08
Maintain a state to track whether the initial subscription validation data loading is done

isharac:
https://github.com/wso2/carbon-apimgt/pull/8563


components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/impl/SubscriptionDataStoreImpl.java
```diff
@@ -250,4 +273,64 @@ public void run() {
...
+
+    public boolean isSubscriptionPoliciesInitialized() {
+
+        return subscriptionPoliciesInitialized;
+    }
+
+    public void setSubscriptionPoliciesInitialized(boolean subscriptionPoliciesInitialized) {
+
+        this.subscriptionPoliciesInitialized = subscriptionPoliciesInitialized;
+    }
```
chamilaadhi, line 335: I think we only need getter methods for this case. Also we could add a single method to check whether everything is initialized (result of all the boolean values) instead of having separate getter methods.

isharac: +1

isharac: fixed with 41eb399b03e35d34e67d40d8e42ad37f1d3e9dd6

PR #8698, opened 2020-06-10
Added REST API to validate the scope.

Meruja:
Already existing API has been modified with different resource path.


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIAdminImpl.java
```diff
@@ -585,4 +586,69 @@ public Workflow getworkflowReferenceByExternalWorkflowReferenceID(String externe
         }
         return workflow;
     }
+
+    /**
+     * TThis method used to check the existence of the scope name for the particular user
+     *
```
tgtshanika, line 592: Please correct the typo here

Meruja: Fixed with 7a3f728

PR #8738, opened 2020-06-12
Add a config to restrict token type when creating a new application

RAVEENSR:
## Goal
**This is a front port of a fix done to APIM 3.1.0 version.**

This PR fixes the issue https://github.com/wso2/product-apim/issues/8187.

This PR also fixes https://github.com/wso2/product-apim/issues/8325.

## Related PR
https://github.com/wso2/product-apim/pull/8326

## Description
This PR introduces a new config `allowed_application_token_types` which is defined under `[apim.devportal]` element to restrict the token type for creating a new application. 
example config in deployment.toml:
```
[apim.devportal]
...
allowed_application_token_types = ["JWT"]
```

The applicable values for the above config are `["JWT", "OAUTH"]` or `["OAUTH"]` or `["JWT"]`.

The relevant config entry in the api-manager.xml file would be as follows(if both are selected):
```
<APIStore>
...
    <AllowedApplicationTokenTypes>
        <TokenType>JWT</TokenType>
        <TokenType>OAUTH</TokenType>
    </AllowedApplicationTokenTypes>
</APIStore>
```

## Approach
By adding the above-mentioned config in the `deployment.toml` file the user can restrict the token type used to create the application. **This restriction is applicable to both UI level and dev portal REST API level(Dev Portal v1).** If the admin given value for the above config is `["OAUTH"]` then the `JWT` token type application creation is restricted and vice versa. The default created application will also adhere to this restriction. If `["JWT", "OAUTH"]` is given then both token types are allowed when creating the application. In a scenario where one user has created an application which uses the `JWT` token type and then the server configs are given to restrict it and allow only `OAUTH` application creation, the token type(JWT) of the already created application will not be changed. However, if the user needs to edit the application and change the token type to `OAUTH` it is possible. If the user is doing this change from the UI, a warning message is displayed in the helper text which is associated with the token type.
warn

In a scenario where an attempt to create an application using the dev portal REST API(v1) using a restricted token type, `"Specified token type is not allowed"` bad request error will be thrown. 

In a scenario where the allowed token type is `OAUTH` and using the dev portal v1 REST API an application creation request is made without mentioning the token type in the payload, by default[1] the  REST API will set the token type to `JWT`. However, in the validation, it will fail and will result in `"Specified token type is not allowed"` bad request error. 

[1] https://github.com/wso2/carbon-apimgt/blob/15aa104c0b29fda40c00825cb883e7b08b9a8e57/components/apimgt/org.wso2.carbon.apimgt.rest.api.store.v1/src/main/resources/store-api.yaml#L4363


Arshardh:
Please get this pr merged soon

RAVEENSR:
Closing this PR as from 3.2.0 onwards the only application token type would be JWT.

PR #8741, opened 2020-06-12
Publisher UI performance Optimization

kavishkafernando:
Publisher UI performance Optimization

Fixes issue https://github.com/wso2/product-apim/issues/8100


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/data/APIClient.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
```
tmkasun, line 1: Why do we need to disable the rule for the entire file?

kavishkafernando: Fixed with `a25e819`

PR #8780, opened 2020-06-18
[UI] Devportal password change

chathuranga95:
### Description
The UI for the developer portal password change feature. Note that the functionality is not implemented since the REST API resources are not available at the moment.

### Screenshot
Screenshot from 2020-06-18 19-32-06


rmsamitha:
Shall we change the phrase 'oldPassword' to 'currentPassword'. It is the better wording. The REST API is also using the 'currentPassword'

chathuranga95:
The phrase replace: old password--> current password is done now

PR #8781, opened 2020-06-18
Fixing JWTConfig read issue.

dinusha92:
The issue is related to read JWT Config -> ConvertClaimsToConsumerDialect.
This was read using the first property. Hence this should be read as JWTConfiguration.ConvertClaimsToConsumerDialect.

But it was there without the "."(dot) separation. This PR fixes that issue.


dinusha92:
Fixes https://github.com/wso2/product-apim/issues/8505

PR #8795, opened 2020-06-20
fix keymanager not retrieve from rest api

tharindu1st:


components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/entity/ApplicationKeyMapping.java
```diff
@@ -71,14 +72,30 @@ public void setApplicationId(int applicationId) {
     }
 
     @Override
-    public String getCacheKey() {
+    public ApplicationKeyMappingCacheKey getCacheKey() {
 
-        return getConsumerKey();
+        return new ApplicationKeyMappingCacheKey(getConsumerKey(), getKeyManager());
```
isharac, line 77: Can't we use the String concatenation here rather than creating an object?

tharindu1st: since these two strings can have the concat character its good to have a cache key object

PR #8845, opened 2020-06-25
Update wso2 dependencies

uvindra:


tharindu1st:
need to update same in product-apim

uvindra:
> need to update same in product-apim

Yes , that is done by https://github.com/wso2/product-apim/pull/8465

PR #8892, opened 2020-07-01
Fixing issue Devportal Application page issues

chanaka3d:
### Update Application keygen pages to align with new multiple KMs

Screen Shot 2020-07-11 at 8 43 18 PM

### When there is only one KM, the tabs are hidden.

Screen Shot 2020-07-12 at 8 58 04 AM

### Wizard
For the wizard disable the next button when there is no KM with Client Credentials grant is not selected. And let user go with the default KM, even there are multiple KMs available. 
And select all the grant types, 
Make the sandbox environment selected by default.
Select all grant types available for the KM.
If the implicit or Application code grants available with the KM, set http://localhost as the default callback.
Tell the user that these configs can be changed with the applications page.
Screen Shot 2020-07-11 at 8 42 02 PM

### Removed the "Token Generation" button when the Client Credentials grant is not selected.
captured (2)

### Remove the Token Validity time from the token gen screen.

Screen Shot 2020-07-12 at 9 08 15 AM

### Remove the Key type field from the application creation page
Screen Shot 2020-07-12 at 8 58 22 AM

Fixing https://github.com/wso2/product-apim/issues/7968


chanaka3d:
We can't merge this. need feather modifications.

PR #8894, opened 2020-07-01
Remove port value from IP address in X-Forwarded-For header

Jayanie:
Fix the issue related to X-Forwarded-For header containing a IP address with a port value.


RAVEENSR:
Related issue: https://github.com/wso2/product-apim/issues/8518

PR #8923, opened 2020-07-03
fix query complexity analysis in jwt

tharindu1st:


Arshardh:
Fixes https://github.com/wso2/product-apim/issues/8558

PR #8930, opened 2020-07-05
Fix DB script errors for Subscription validation

Arshardh:


Arshardh:
Fixes https://github.com/wso2/product-apim/issues/8522

PR #8969, opened 2020-07-13
Block creation of applications of OAUTH token type

ChamodDamitha:


tharindu1st:
this will thorwing exception to the client about token type.
Since we need to handle it in silently

PR #8978, opened 2020-07-14
Avoid appending the version twice to the context when updating an exported-imported API using APICTL (Fix for product-apim-tooling/issues/404) 

Chamindu36:
## Purpose
After exporting an API using the APICTL, when we reimport the API to APIM and edit it in publisher the store will display the context by appending the version twice. This PR will solve this issue.

## Goals
Fixes https://github.com/wso2/product-apim-tooling/issues/404

## Approach

Removing process of extracting up x-wso2-basepath extension from a swagger or open API definition and set that value as the context of API when importing an API through the API Controller tool. 

## User stories
Importing an API using the API Controller tool.

## Test environment
OS - Ubuntu 20.04 LTS
Java - JDK 1.8_252
APIM - 3.2.0
GO - 1.14
APICTL- 3.2.0 Beta

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS2Parser.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS2Parser.java
@@ -1402,13 +1402,12 @@ public class OAS2Parser extends APIDefinition {
     /**
      * This method returns api that is attached with api extensions related to micro-gw
      *
-     * @param apiDefinition                  String
-     * @param api                            API
-     * @param isBasepathExtractedFromSwagger boolean
+     * @param apiDefinition String
+     * @param api           API
      * @return API
      */
     @Override
-    public API setExtensionsToAPI(String apiDefinition, API api, boolean isBasepathExtractedFromSwagger) throws APIManagementException {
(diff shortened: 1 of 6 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS2Parser.java
```diff
@@ -1459,13 +1458,7 @@ public API setExtensionsToAPI(String apiDefinition, API api, boolean isBasepathE
...
             api.setApiLevelPolicy(throttleTier);
         }
-        //Setup Basepath
-        String basePath = OASParserUtil.getBasePathFromSwagger(extensions);
-        if (StringUtils.isNotBlank(basePath) && isBasepathExtractedFromSwagger) {
-            basePath = basePath.replace("{version}", api.getId().getVersion());
-            api.setContextTemplate(basePath);
-            api.setContext(basePath);
-        }
+
```
Shehanir, line 1461: Do we need this extra line here

Chamindu36: Fixed with c1f846270025eb67fcd8e92c102a8539e1e699a3

PR #8988, opened 2020-07-16
Resolve errors in Swagger definitions.

Meruja:
Structural error at definitions:
RoleAlias and Scope
Definitions should NOT have an additional property: schema


components/apimgt/org.wso2.carbon.apimgt.rest.api.admin.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/v1/utils/mappings/SystemScopesMappingUtil.java
```diff
@@ -99,13 +103,16 @@ public static JSONObject createJsonObjectOfScopeMapping(ScopeListDTO body)
         JSONArray scopeJson = new JSONArray();
         for (ScopeDTO scope : body.getList()) {
             JSONObject scopeRoleJson = new JSONObject();
+            String roles = scope.getRoles().toString().replaceAll("\\[", "").
```
tgtshanika, line 106: Isn't a null check required here?

Meruja: We don't need to check null here because when we define scopes in tenant-conf we should give role name as well.

PR #9012, opened 2020-07-21
Fix migration issues with swagger security schemes

msm1992:
Fixes https://github.com/wso2/product-apim/issues/8724 and https://github.com/wso2/product-apim/issues/8748

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
@@ -205,4 +205,13 @@ public abstract class APIDefinition {
     public abstract API setExtensionsToAPI(String swaggerContent, API api)
             throws APIManagementException;
 
+    /**
+     * This method will extract scopes from legacy x-wso2-security and add them to default scheme
+     * @param swaggerContent swagger definition
+     * @return
+     * @throws APIManagementException
+     */
+    public abstract String processLegacyScopes(String swaggerContent)
+            throws APIManagementException;
+
 }
--- a/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
@@ -204,14 +204,4 @@ public abstract class APIDefinition {
      */
     public abstract API setExtensionsToAPI(String swaggerContent, API api)
             throws APIManagementException;
-
-    /**
-     * This method will extract scopes from legacy x-wso2-security and add them to default scheme
-     * @param swaggerContent swagger definition
-     * @return
-     * @throws APIManagementException
-     */
-    public abstract String processLegacyScopes(String swaggerContent)
-            throws APIManagementException;
-
 }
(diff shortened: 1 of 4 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
```diff
@@ -1325,7 +1325,8 @@ static JsonNode getResponseCacheConfig(String responseCache, int cacheTimeout) {
     public static String preProcess(String swaggerContent) throws APIManagementException {
         //Load required properties from swagger to the API
         APIDefinition apiDefinition = getOASParser(swaggerContent);
-        return apiDefinition.processOtherSchemeScopes(swaggerContent);
+        swaggerContent = apiDefinition.processOtherSchemeScopes(swaggerContent);
+        return apiDefinition.processLegacyScopes(swaggerContent);
```
ruks, line 1329: shouldn't processLegacyScopes execute before the processOtherSchemeScopes?

components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
```diff
@@ -205,4 +205,13 @@ public abstract String processOtherSchemeScopes(String resourceConfigsJSON)
     public abstract API setExtensionsToAPI(String swaggerContent, API api)
             throws APIManagementException;
 
+    /**
+     * This method will extract scopes from legacy x-wso2-security and add them to default scheme
+     * @param swaggerContent swagger definition
+     * @return
+     * @throws APIManagementException
+     */
+    public abstract String processLegacyScopes(String swaggerContent)
```
ruks, line 214: I think we should not expose processLegacyScopes via the interface because I think it is not required to execute directly. 
I think we can define same method in respective parsers as private and call inside the processOtherSchemeScopes method. Because processLegacyScopes is used by only processOtherSchemeScopes.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
msm1992: processLegacyScopes will add the default schema to SecuritySchemes and since we now skip the processOtherSchemeScopes if the default schema already exists in the definition we have to call processLegacyScopes after calling processOtherSchemeScopes.
If we are going to call processLegacyScopes before calling processOtherSchemeScopes we have to copy the legacy scopes to some oauth2 scheme other than 'default' and let them be copied to 'default' scheme when processOtherSchemeScopes is executed.

components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIDefinition.java
msm1992: fixed with https://github.com/wso2/carbon-apimgt/pull/9012/commits/b188ce5404a9c6d4634d3c3ad7d44eba254e4437

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
msm1992: fixed with https://github.com/wso2/carbon-apimgt/pull/9012/commits/b188ce5404a9c6d4634d3c3ad7d44eba254e4437

ruks: Yes. Got it.

PR #9041, opened 2020-07-23
Disable the buttons and tet boxes when a KM is disabled

chanaka3d:
If the KM is disabled, added the following to disable all the UI elements and show a warning.

Screen Shot 2020-07-23 at 12 26 42 PM


chanaka3d:
Updated as requested 
Screen Shot 2020-07-23 at 4 34 31 PM

PR #9103, opened 2020-07-31
Changing the key manager uis to check the type

chanaka3d:


features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/KeyManagers/AddEditKeyManager.jsx
```diff
@@ -578,7 +578,7 @@ function AddEditKeyManager(props) {
                                 >
                                     {settings.keyManagerConfiguration.map((keymanager) => (
                                         <MenuItem key={keymanager.type} value={keymanager.type}>
-                                            {keymanager.type}
+                                            {keymanager.displayName || keymanager.type}
```
chanaka3d, line 581: We don't need this "OR" check
In your case you can put {keymanager.displayName}

PR #9109, opened 2020-07-31
[Private Jet Mode] Fix deselecting, selected environment in dev portal try-out page

renuka-fernando:
- Fix the issue of Deselecting environment when loading the try out page of dev portal

image


renuka-fernando:
Fix https://github.com/wso2/product-apim/issues/8873

PR #9127, opened 2020-08-05
Fix intermittent test failures for GatewayArtifactsMgt DAO

Sarangan0219:
This PR fixes intermittent test failure issues in GatewayArtifactsMgt DAO


Arshardh:
will it be sufficient only to have a sleep in the setup method?

Sarangan0219:
> will it be sufficient only to have a sleep in the setup method?

We don't need to delay the addition test methods. We only need to delay the getter test methods as the database update is not get reflected in retrieving them.

Arshardh:
got it

PR #9147, opened 2020-08-07
Adding permission validation for workflow UIs

chanaka3d:
Screen Shot 2020-08-07 at 10 04 27 AM

Fixing https://github.com/wso2/product-apim/issues/8953


chamilaadhi:
@chanaka3d I think you missed the admin-api.yaml related changes

chamilaadhi:
@chanaka3d Additionally, I think we might hv to add the permission check for the permission changing section as well.

tmkasun:
@chamilaa , In this change, we are allowing a subset of special admin users to login to the admin portal and manage(only) the workflow related tasks, hence we will not need to change the `admin-api.yaml` nor permission selection UI

PR #9154, opened 2020-08-10
[Admin Portal][UI]Update error message on application ownership change

chathuranga95:
### Description
When changing the ownership of an application using the admin portal UI, 
when the target application owner is a valid `subscriber` and hasn't logged into the devportal at least once,
the REST API responses with `500`. But since this is a special troubleshooting step which is included in the [docs](http://localhost:8000/learn/consume-api/manage-application/advanced-topics/changing-the-owner-of-an-application/) too, we need to display a meaningful Error alert.

> Please note that the separate API call which is used for checking the target owner is a subscriber or not is returning with true even if that user hasn't logged into the devportal at least once. (which is expected behavior IMO). Therefore, the error message update is needed.

This PR updates the error message to `Error while updating ownership to <owner>`

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/ApplicationSettings/EditApplication.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/ApplicationSettings/EditApplication.jsx
@@ -76,7 +76,7 @@ function Edit(props) {
     };
 
     const validateOwner = () => {
-        let validationError = '';
+        let validationError = 'Something went wrong when validating user';
 
         const applicationsWithSameName = applicationList.filter(
             (app) => app.name === name && app.owner === owner,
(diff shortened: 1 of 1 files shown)


features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/ApplicationSettings/EditApplication.jsx
```diff
@@ -124,8 +124,9 @@ function Edit(props) {
                 })
                 .catch((error) => {
                     const { response } = error;
-                    if (response.body) {
-                        throw response.body.description;
+                    if (response.body.code === 500) {
```
tmkasun, line 127: catch clause could catch none-HTTP exceptions, in that case, error object will not have a body or code , so we need to do a sanity check here, We can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here

```javascript
if (response?.body?.code === 500)
```

chathuranga95: fixed with: https://github.com/wso2/carbon-apimgt/pull/9154/commits/e218dec7a0576687cb375829d06a5a5bb2b90708

chathuranga95:
Added fix for [this](https://github.com/wso2/product-apim/issues/9068) issue as well.

PR #9169, opened 2020-08-12
Add a description to the default application

rmsamitha:


tharindu1st:
need to fix in both ends  from Rest API need to send empty description

rmsamitha:
> need to fix in both ends from Rest API need to send empty description

@tharindu1st Null return value will be handled by the UI. Returning empty string for null descriptions can be considered as a separate effort with other devportal/publisher REST APIs. There are other places we return null for empty values.

PR #9175, opened 2020-08-14
Produce Alerts when trying to generate key from a disabled key manager

chathuranga95:
When a key manager is disabled and we are trying to generate Access Tokens from it, the UI doesn't produce a proper error message.

This PR produces an error Alert stating that Key Manager is not enabled.

Fixes: https://github.com/wso2/product-apim/issues/8930


tharindu1st:
@chathuranga95  please send fixes to 6.7.x

PR #9176, opened 2020-08-14
Produce Alerts when trying to generate key from a disabled key manager

chathuranga95:
When a key manager is disabled and we are trying to generate Access Tokens from it, the UI doesn't produce a proper error message.

This PR produces an error Alert stating that Key Manager is not enabled.

Fixes: https://github.com/wso2/product-apim/issues/8930

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/devportal/source/src/app/components/Shared/AppsAndKeys/ViewKeys.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/devportal/source/src/app/components/Shared/AppsAndKeys/ViewKeys.jsx
@@ -293,7 +293,7 @@ class ViewKeys extends React.Component {
                 }
                 this.setState({ isUpdating: false });
                 const { response } = error;
-                if (response.body) {
+                if (response && response.body) {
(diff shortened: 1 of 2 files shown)


features/apimgt/org.wso2.carbon.apimgt.store.feature/src/main/resources/devportal/source/src/app/components/Shared/AppsAndKeys/ViewKeys.jsx
```diff
@@ -291,6 +292,10 @@ class ViewKeys extends React.Component {
                     this.setState({ notFound: true });
                 }
                 this.setState({ isUpdating: false });
+                const { response } = error;
+                if (response.body) {
```
tmkasun, line 296: shall we check the response value as well
```
 if (response && response.body) {
```

otherwise, if `response` is undefined this line will fail

chathuranga95: Fixed with: https://github.com/wso2/carbon-apimgt/pull/9176/commits/3d152a9f1b35a4528529dfe91a884c75eebf1202

PR #9178, opened 2020-08-14
fix for https://github.com/wso2/product-apim/issues/8925

isharac:


isharac:
fixes okta token generation issue with scopes

PR #9190, opened 2020-08-18
Provide the capability to update policyID in db2

Sarangan0219:
### Problem
Fixes https://github.com/wso2/product-apim/issues/9086
An identity column defined as GENERATED ALWAYS cannot be allowed to provide an explicit value for db2.

### Solution

If you define the column as GENERATED BY DEFAULT, we can insert a value, and Db2 provides a default value if we do not supply one.


uvindra:
As discussed, lets try to come up with a suitable update query instead of changing the table create statement.

Sarangan0219:
Iam Closing this PR as the Review comments are addressed in this PR https://github.com/wso2/carbon-apimgt/pull/9211

PR #9193, opened 2020-08-18
Fix Issue when updating API categories via a publisher user

msm1992:
Fixes https://github.com/wso2/product-apim/issues/9101


rmsamitha:
The change should be added to the test resources in product-apim repo too

PR #9209, opened 2020-08-19
[Improvement] Add support for "x-wso2-application-security" mgw extension in API level

Chamindu36:
### Purpose
x-wso2-application-security is not supported when defined in API definition when creating an API with existing API definition.
This application security values should be applied under the runtime configurations in API. 

When the Following extension is defined in the API definition this should reflect it's changed configs in publisher UI which is not supported currently.
```
 x-wso2-application-security: 
        security-types: 
          - "api_key"
        optional: false
```
This PR will support this task.

### Goals
Improve the extension support which is used by the micro gateway.
Fixes https://github.com/wso2/product-apim/issues/9133
and https://github.com/wso2/product-apim-tooling/issues/469

### Approach
In OASParsers of APIM, there are methods to preprocess swagger definitions before actually processing and creating API objects. On that level, the OAS parsers can map x-wso2-application-security extension values into API object in OAS parsers when creating API objects using these OpenAPI definitions. 

### User stories
Import API via API Controller.
Import API via Publisher Portal

### Documentation
No doc changes required.

### Test environment
Ubuntu 20.04 LTS
Java JDK 1.8_241
APIM 3.2.0
APICTL 3.2.0

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
@@ -1467,4 +1467,16 @@ public class OASParserUtil {
         return authorizationHeader == null ? null : authorizationHeader.toString();
     }
 
+    /**
+     * This method returns extension of mutualSSL related to micro-gw
+     *
+     * @param extensions Map<String, Object>
+     * @return String
+     * @throws APIManagementException throws if an error occurred
+     */
+    public static String getApplicationSecurity(Map<String, Object> extensions) throws APIManagementException {
+        Object applicationSecurity = extensions.get(APIConstants.X_WSO2_APP_SECURITY);
+        return applicationSecurity == null ? null : applicationSecurity.toString();
+    }
+
 }
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
@@ -1325,6 +1325,7 @@ public class OASParserUtil {
     public static String preProcess(String swaggerContent) throws APIManagementException {
         //Load required properties from swagger to the API
         APIDefinition apiDefinition = getOASParser(swaggerContent);
+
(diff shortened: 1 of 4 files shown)


malinthaprasan (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
```diff
@@ -1467,4 +1467,40 @@ public static String getAuthorizationHeaderFromSwagger(Map<String, Object> exten
         return authorizationHeader == null ? null : authorizationHeader.toString();
     }
 
+    /**
+     * This method returns extension of application security types related to micro-gw
+     *
+     * @param extensions Map<String, Object>
+     * @return String
```
rmsamitha, line 1474: @return type wrong. And there is no use of this doc comment if you don't add a description on what it returns.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
```diff
@@ -1467,4 +1467,40 @@ public static String getAuthorizationHeaderFromSwagger(Map<String, Object> exten
...
+            appSecurityTypes = mapper.convertValue(appSecurityTypesNode.get("security-types"), ArrayList.class);
+        }
+        return appSecurityTypes;
+    }
+
+    /**
+     * This method returns extension of application security types state related to micro-gw
+     *
+     * @param extensions Map<String, Object>
+     * @return boolean
```
rmsamitha, line 1492: here is no use of this doc comment if you don't add a description on what it returns.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
Chamindu36: Fixed with c1f846270025eb67fcd8e92c102a8539e1e699a3

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OASParserUtil.java
Chamindu36: Fixed with c1f846270025eb67fcd8e92c102a8539e1e699a3

PR #9214, opened 2020-08-20
Fix for Jaeger tracing issue

ShalkiWenushika:
This PR will fix the Jaeger tracing issue when using a hostname different than localhost.


tharindu1st:
This was already fixed in 6.7.x

PR #9235, opened 2020-08-21
Policy name Limit to 30 characters

chathuranga95:
### Description

This PR set the policy name length validation limit to 30 characters. (considering the database storing limits and consistency)


Arshardh:
Let's merge to master instead

PR #9237, opened 2020-08-21
Fixing Issue with IPv6 address in X-Forwarded-For header

Jayanie:
- This Fixes the issue with IPV6 address in X-Forwarded-For header


RAVEENSR:
Related issue: https://github.com/wso2/product-apim/issues/9149

PR #9257, opened 2020-09-03
Enable Mutual SSL behind Proxy/LB

Meruja:
Fix for wso2/carbon-apimgt#9172 
There are some use cases where the Proxy/LB send the client certificate unencoded as a header. But Mutual SSL is failing in this use-case since it is always trying to decode the certificate present in the header by the source code.

In this fix, we introduced a new configuration to determine whether the certificate has to be decoded or not.
[apimgt.mutual_ssl]
certificate_header = "Header Name"
#This property needs to be true if MutualSSL connection established between load balancer and gateway.
enable_client_validation = false
#This property needs to be false if the certificate shouldn’t be decoded when it is passed from load balancer to the gateway.
client_validation_encode = false


tharindu1st:
@Meruja  please include integration/ unit tests and update or with master

dushaniw:
@Meruja Can we get the conflicts resolved and relevant tests update?

dushaniw (approved):

PR #9289, opened 2020-09-29
Fix HTML rendering issue in MarkdownEditor.jsx

menakaj:
Fixes https://github.com/wso2/product-apim/issues/9312


chanaka3d:
I think the same  need to be done in the devportal side to show the markup properly.

chanaka3d (approved):

PR #9319, opened 2020-10-15
[Do not merge] Add DCR app creation for portal web apps to server startup

tharikaGitHub:
This PR adds DCR web app creation of portal apps (Admin, Publisher and Devportal) to the time the server starts up.


tmkasun:
Shall we close this PR ? since this is no longer required ?

PR #9361, opened 2020-11-03
Moving the Storage/Retrieval of API UUID to the Database.

shilmyhasan:
Currently the API UUID is stored and retrived from the Registry. The PR changes this behavior where the UUID will be stored and retrieved from AM_API table in the WSO2AM DB.

The PR addresses the first part of the following issue
- [https://github.com/wso2/product-apim/issues/9407](https://github.com/wso2/product-apim/issues/9407)


malinthaprasan (dismissed):

malinthaprasan (changes requested):
If we are persisting the UUID in AM_API, we have to persist the UUID of the Registry artifact. Otherwise there are few issues:

- As per this PR, the flow of getting an API is we get the API Path from UUID from AM_API and passing that path to fetch the resource. But, it is not a good approach to depend on the usual paths because we are again going to further bind with provider-APIName-version based path. We are already getting issues due to this when API Name having some uncommon charactors and those are getting into registry paths giving issues. 
- We are also developing an abstraction for the registry/mongoDB based registry implementation. With the above PR, the entry point for the registry will be getAPI(some_reg_path) which is not going to be a valid method for the interface.
- Earlier we had retrieve api gov artifact using the UUID directly. With this approach, we have to first get build a registry path using AM_API table -> then get the reg resource -> get UUID -> then retrieve gov artifact. Won't this be just increasing the amount of DB calls?

Hence shall we do below changes before merging this:
- If we are bringing a UUID to the AM_API bring the original UUID of the API artifact
- Do not change the entry point of getting artifacts from registry other than providing the UUID of the API - like getAPI(UUID) because we can't have a method like getAPI(some reg path) in the persistence interface.

shilmyhasan:
@tharindu1st How can we proceed with this ? Should we revert the changes to persist the ID of the artifact itself in DB and let the flow of retrieving an artifact through the UUID as it is.

dushaniw:
Fix added with https://github.com/wso2/carbon-apimgt/pull/9397

PR #9389, opened 2020-11-14
Added a system property to stop updating service provider name.

prasa7:
To stop overriding the Service provider name we need to set override.sp.name system property to false
-Doverride.sp.name=false


1akshitha:
Fixes https://github.com/wso2/product-apim/issues/9455

mushthaq33 (approved):

bhathiya:
@prasa7 why is this till open?

PR #9403, opened 2020-11-19
Fix scope validation issue when invoking default API with Oauth token

YasasRangika:
## Purpose
Fixes: https://github.com/wso2/product-apim/issues/9386

## Goals
To solve the failure of scope validation when invoking default version of an API using Oauth as token type with no version in the uri.

## Approach
Defined new variable inside the DefaultKeyValidationHandler.java to store pure api version since default api versions are prefixed with `_default_` and it causes to fail the api search using `getApiByContextAndVersion()`.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/handlers/DefaultKeyValidationHandler.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/handlers/DefaultKeyValidationHandler.java
@@ -158,10 +158,17 @@ public class DefaultKeyValidationHandler extends AbstractKeyValidationHandler {
         } else {
             resourceArray = new ArrayList<>(Arrays.asList(resourceList));
         }
+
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/handlers/DefaultKeyValidationHandler.java
```diff
@@ -158,10 +158,17 @@ public boolean validateScopes(TokenValidationContext validationContext) throws A
         } else {
             resourceArray = new ArrayList<>(Arrays.asList(resourceList));
         }
+
+        String actualVersion = validationContext.getVersion();
```
bhathiya, line 162: it's wrong to call this actual version right?

YasasRangika: Let's make it as apiVersion and then assign the actual version by splitting.

bhathiya (approved):

PR #9485, opened 2020-12-10
Provides the ability to customize the apikey generation and signing

Ketharan:
**Purpose**
It provides the ability to customize the API key generation and signing.
Fixes public issue :  https://github.com/wso2/product-apim/issues/8448

**Approach**
Adds a configuration option to provide a separate Keystore to provide the private key to sign the API key.
Also an extendable interface is given to add any custom implementation for API key generation.

If a user needs to use a custom Keystore to sign the JWT of  APIKey then that custom Keystore file should be added to the Keystore directory (`APIM_Home/repository/resources/security`) and `deployment.toml` via the config

```
[custom_keystore.KeySignKeyStore]
file_name =  "apikeysigner.jks"
type =  "JKS"
password =  "wso2carbon"
alias =  "apikeysigner"
key_password =  "wso2carbon"
```

and the name of the custom Keystore (that is the later part divided by the full stop in the config header, i.e `KeySignKeyStore`) should be given as `api_key_keystore` and the `api_key_alias` should point to the alias of the public certificate of the corresponding private key, which already added to the client trust store. 

```
[apim.devportal]
api_key_alias = "apikeysigner"
api_key_keystore = "KeySignKeyStore"
```

If the `api_key_keystore` is not configured then Primary Keystore will be used for signing and the `api_key_alias` should point to the public certificate alias of the corresponding private key of Primary Keystore.

Further the APIKey generation can be extended either implementing the interface
`org.wso2.carbon.apimgt.impl.token.ApiKeyGenerator`
or extending the class 
`org.wso2.carbon.apimgt.impl.token.DefaultApiKeyGenerator` 
and mentioning the fully qualified class name
in `deployment.toml` 
eg.

```
[apim.devportal]
api_key_generator_impl = org.wso2.carbon.apimgt.impl.token.CustomApiKeyGenerator
```

If that configuration is not given then the default implementation 
`org.wso2.carbon.apimgt.impl.token.DefaultApiKeyGenerator`
will be used.


Ketharan:
Closing this hence this is already added to the master

PR #9550, opened 2021-01-07
Fixing product-apim/issues/9586

mushthaq33:
This PR fixes https://github.com/wso2/product-apim/issues/9586

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/data/AuthManager.js
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/data/AuthManager.js
@@ -19,7 +19,6 @@
 import qs from 'qs';
 import CONSTS from 'AppData/Constants';
 import Configurations from 'Config';
-import API from 'AppData/api';
 import Utils from './Utils';
 import User from './User';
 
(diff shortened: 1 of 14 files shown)


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/data/AuthManager.js
```diff
@@ -19,6 +19,7 @@
 import qs from 'qs';
 import CONSTS from 'AppData/Constants';
 import Configurations from 'Config';
+import API from 'AppData/api';
```
tmkasun, line 22: We have a small problem here, Importing `AppData/api` to Authmanager will import all of `AppData/api`'s dependencies(swagger-js, async-mutex ect) here. Which will increase the initial page load timing.

Ideally, we should have decare the constant outside the API class as a separate entity and export it separately in the `AppData/api` and when importing also , We can use named import to get only the constant values.

tmkasun: `AuthManager` is in the initial page load dependency tree, that's why the concern
We have the following options here
- Remove the import and use the string value 
- Change the way constant has defined and export

tmkasun (approved):

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/data/AuthManager.js
mushthaq33: Fixed with fce66217cc5536f61353527c8bfa4e52ccf40ac8

PR #9570, opened 2021-01-11
Adding relevant error messages for Revision REST APIs

CrowleyRajapakse:
## Issues

wso2/product-apim#9406

## Implementation

Adding relevant error messages for REVISION REST APIs
Removing API revision api artifacts and documents from Indexing
Adding check to verify update API calls are only for APIs not for Revisions.


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
```diff
@@ -1572,6 +1580,14 @@ paths:
         200:
           description: |
             OK.
+        201:
```
vithu30, line 1583: I believe, the resource returns 201 response when successfully restoring the revision. If so, shall we remove the 200 response from API definition?

vithu30: Also, REST API impl also need to return the 201 using `Response.created()` instead of `Response.ok()`

CrowleyRajapakse: ack, yes, will remove

CrowleyRajapakse: resolved

PR #9630, opened 2021-01-25
Read the api provider from Authentication handler properties

HeshanSudarshana:
This PR fixes the issue API publisher resolve as Prod when using an unsecured resource in API product.

Fixes https://github.com/wso2/product-apim/issues/9760


HeshanSudarshana:
Closing since the issue is no longer reproduced in master

PR #9631, opened 2021-01-25
Adding initial event publisher client for APIM

ruks:
- Adding initial event publisher client for APIM

To engage the analytics handler, add the following code to deployment.toml
```
enabled_global_handlers=["external_call_logger", "open_tracing", "analytics"]

[synapse_handlers.analytics]
name="analytics"
class="org.wso2.carbon.apimgt.gateway.handlers.analytics.AnalyticsMetricsHandler"
```

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/PublisherUtils.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/PublisherUtils.java
@@ -1,0 +1,69 @@
+/*
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package org.wso2.carbon.apimgt.gateway.handlers.analytics;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.wso2.am.analytics.publisher.exception.MetricCreationException;
+import org.wso2.am.analytics.publisher.exception.MetricReportingException;
+import org.wso2.am.analytics.publisher.reporter.CounterMetric;
+import org.wso2.am.analytics.publisher.reporter.MetricReporter;
+import org.wso2.am.analytics.publisher.reporter.MetricReporterFactory;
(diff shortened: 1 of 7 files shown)


components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/PublisherUtils.java
```diff
@@ -0,0 +1,69 @@
...
+    }};
+    public static final ObjectMapper mapper = new ObjectMapper();
+    public static final TypeReference<Map<String, String>> typeRef = new TypeReference<Map<String, String>>() {
+    };
+
+    public static void doPublish(String metricName, ResponseEvent responseEvent) {
+        MetricReporter metricReporter;
+        try {
+            metricReporter = MetricReporterFactory.getInstance().createMetricReporter(null, configs);
+            CounterMetric counterMetric = metricReporter.createCounterMetric(metricName, MetricSchema.RESPONSE);
```
jaadds, line 60: Doesn't initialising like this make multiple EventHub Clients be built? Can't we keep countermetric as an instance variable and call incrementCount with each event?

ruks: As per the @tishan89 implementation, we don't need to initialize a reporter. It is a singleton publisher and is guaranteed to return the same reporter for the given metric. Will get @tishan89's recommendation as well.

tishan89: metric reporter is singleton and any Metric we create using that reporter will be unique by its name. If you create another metric with same name you will simply get the earlier Metric instance. It is designed this way so that we can easily instrument at multiple points in the request path

jaadds: Got it. But is this logic something we need to run with each invocation?

tishan89: We dont have to create a metric reporter for each event if thats what u are asking. We should create is inside something like init() method and then use the incrementCount() in the critical path for performance concerns

jaadds (approved):

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/PublisherUtils.java
ruks: if so shall we make `metricReporter` as class variable?
```
metricReporter = MetricReporterFactory.getInstance().createMetricReporter(null, configs);
```

mushthaq33 (approved):

ruks:
config updated as follows

```
[synapse_handlers.analytics]
enabled = true
class = "org.wso2.carbon.apimgt.gateway.handlers.analytics.AnalyticsMetricsHandler"
```

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/PublisherUtils.java
jaadds: Metric reporter and CounterMetric both.

PR #9636, opened 2021-01-26
Add Service Catalog Feature Back-end Implementation

YasasRangika:
### Objective
The main purpose of this Service Catalog feature is to create a connection between API Management and Integration layers by making integration services discoverable to API Management layer.

### Approach
Service catalog portal will consist of service catalog web user interface which is available in the Publisher web portal, REST API, and data persistent components. Each tenant will have a separate service catalog space, so that only the users from the same tenant space can read from and write to the catalog.
For this release it will available four main endpoints that can use to work with Service Catalog.

**POST	/service-entries/import**

- Import a service by providing an archived service
- This endpoint currently support with overwrite parameter = true
- If the client send request with payload zip that contains directories relevant to each service with overwrite=true, then this endpoint will overwrite all the entries.

**GET		/service-entries**

- Retrieve or search services in the service catalog of the user's organization or tenant.
- Client should send the request with shrink=true

**DELETE	​/service-entries​/{serviceKey}**

- Delete a service by providing the service key

**GET		/services/export**

- Export a service as an archived zip file.


malinthaprasan:
Closing due to other opened PR (9638)

PR #9655, opened 2021-01-28
Fixing https://github.com/wso2/product-apim/issues/9749

chanaka3d:
download (1)


CrowleyRajapakse (approved):

menakaj:
I just have a concern about the top panel. To me, it's like cluttered. Need to scan all the way from left to right to see the info. Is the orientation configurable? 

Screenshot - 2021-01-29T091156 515

chanaka3d:
> I just have a concern about the top panel. To me, it's like cluttered. Need to scan all the way from left to right to see the info. Is the orientation configurable? @chanaka3d @tmkasun
> 
> Screenshot - 2021-01-29T091156 515

This is not configurable. The design is base on a discussion we had earlier.

PR #9666, opened 2021-02-01
Update prototypes

erangatl:


tmkasun (approved):

tmkasun:
We could say this as `API Revisions` 
image

image


There are changes and designs related to publisher portal as well , I think done in Figma , shall we link that too ?

PR #9679, opened 2021-02-02
Updating the start page

chanaka3d:
Updated the onboarding page as follows 

Screen Shot 2021-02-02 at 2 36 35 PM

Responsive behavior

Screen Shot 2021-02-02 at 2 36 57 PM


menakaj:
Suggestions:
1. Shall we use the Roboto font for the API Types as well? 
2. IMO if we can add a vertical line in between, it would help the separation of API types.
@chanaka3d wdyt?

tmkasun:
Hi,

> 1. Shall we use the Roboto font for the API Types?
Yes +1 to [Roboto](https://fonts.google.com/specimen/Roboto), And also we need to keep uniform serif san-serif type faces.

> 2. IMO if we can add a vertical line in between, it would help the separation of API types.
Adding lines would clutter to make it looks more compact i think, My suggestion is to add more space to show a clear separation
> 3. Instead of links, shall we use buttons for the actions? (since links change color when clicked)
-1 , These are actual links so all the link behaviors are expected here, I:e Ctr+link open in a new tab, Right-click and copy link eect, Button will screw that. For the staying issue, we can use a class or globally suppress the colors of the visited links (prefer a local class)
>    @chanaka3d wdyt?

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Listing/SampleAPI/SampleAPI.jsx
```diff
@@ -17,43 +17,38 @@
...
+    links: {
+        cursor: 'pointer',
     },
-    buttonLeft: {
-        marginRight: theme.spacing(1),
+    root: {
+        '& a': {
+            color: '#34679D',
+        },
+        '& a:visited': {
```
tmkasun, line 49: @menakaj , @chanaka3d 
I think this will keep the visited color same as the link color

image

tmkasun:
For me, The low-resolution layout is a bit cluttered ,

@chanaka3d Shall we remove the description in low res ( < sm) ?


image

tmkasun (approved):

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Listing/SampleAPI/SampleAPI.jsx
chanaka3d: Yes. That was the plan. Otherwise, the UI will have two different colors.

menakaj: +1

PR #9764, opened 2021-02-11
Adding additional data to analytics events

ruks:
- Adding apiType and platform to analytics event
- Include applicationUUID to message context
- Use API information from the internal SubscriptionDataStore
- Collect method not allowed, resource not found, and other faulty event types.  
- Calculated and publish request path and response path latency

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/basicauth/BasicAuthAuthenticator.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/basicauth/BasicAuthAuthenticator.java
@@ -203,6 +203,7 @@ public class BasicAuthAuthenticator implements Authenticator {
             authContext.setCallerToken(null);
             authContext.setApplicationName(null);
             authContext.setApplicationId(clientIP); //Set clientIp as application ID in unauthenticated scenario
+            authContext.setApplicationUUID(clientIP); //Set clientIp as application ID in unauthenticated scenario
             authContext.setConsumerKey(null);
             APISecurityUtils.setAuthenticationContext(synCtx, authContext, null);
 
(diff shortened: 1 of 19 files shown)


components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/basicauth/BasicAuthAuthenticator.java
```diff
@@ -203,6 +203,7 @@ public AuthenticationResponse authenticate(MessageContext synCtx) {
             authContext.setCallerToken(null);
             authContext.setApplicationName(null);
             authContext.setApplicationId(clientIP); //Set clientIp as application ID in unauthenticated scenario
+            authContext.setApplicationUUID(clientIP); //Set clientIp as application ID in unauthenticated scenario
```
fazlan-nazeem, line 206: This could result in a large combination of groups in the streaming layer. So I prefer to have a predefined string for this.

ruks: set application UUID same as ID because this goes to auth context. But when publishing to analytics, we check the username and set all attribute to anonymous
```
if (APIConstants.END_USER_ANONYMOUS.equalsIgnoreCase(authContext.getUsername())) {
       this.setAnonymousApp(responseEvent);
 }
```

fazlan-nazeem: ack

pom.xml
```diff
@@ -2087,7 +2087,7 @@
         <version.checkstyle>8.34</version.checkstyle>
         <org.wso2.orbit.org.mapstruct.version>1.4.1.wso2v1</org.wso2.orbit.org.mapstruct.version>
         <org.mapstruct.version>1.4.1.Final</org.mapstruct.version>
-        <analytics.event.publisher.client.version>0.3.0</analytics.event.publisher.client.version>
+        <analytics.event.publisher.client.version>0.3.1-SNAPSHOT</analytics.event.publisher.client.version>
```
fazlan-nazeem, line 2090: update to release version?

ruks: updated to released version

fazlan-nazeem (approved):

PR #9791, opened 2021-02-14
Move analytics codes to gateway common

ruks:
- Move analytics codes to gateway common
- Provide shared analytics event identification


tishan89 (approved):

components/apimgt/org.wso2.carbon.apimgt.gateway.common/src/main/java/org/wso2/carbon/apimgt/gateway/common/analytics/collectors/AnalyticsDataProvider.java
```diff
@@ -0,0 +1,74 @@
...
+import org.wso2.carbon.apimgt.gateway.common.analytics.publishers.dto.Operation;
+import org.wso2.carbon.apimgt.gateway.common.analytics.publishers.dto.Target;
+
+/**
+ * Data provider interface to extract request data
+ */
+public interface AnalyticsDataProvider {
+    boolean isSuccessRequest();
+
+    boolean isFaultRequest();
```
jaadds, line 34: I think rather than returning a boolean we should return the error type enum with each request. But we can do the refactoring later.

jaadds (approved):

PR #9804, opened 2021-02-15
Adding Role Name Validation when Creating/Editing Subscription Policy in Admin Portal

hisanhunais:
## Purpose
- This PR adds role name validation when creating/editing subscription policy in admin portal
- Related Issue: https://github.com/wso2/product-apim/issues/9108

## Goal
- Prevent creation/modification of subscription policies with invalid role names as allowed/denied permissions.

## Approach
- Introduced support for role names to be added as Chip Inputs. A role name has to be added and "Enter" should be pressed. (This is mentioned as a helper text). Then the role name will be added as a Chip Input. 
- When a valid role name is entered, it will be displayed as follows.
roles-chip-inputs

- When an invalid role name is entered, it will be displayed as follows. In this case, the "Save" button also will be disabled.
invalid-roles

- Introduced a new resource(**/roles/{roleId}**) to the admin REST API to validate a given role. This resource is bound with the "admin" scope as well as a newly introduced "role_manage" scope.

## Test Environment
- JDK 1.8.0_241
- Ubuntu 20.04.1 LTS


components/apimgt/org.wso2.carbon.apimgt.rest.api.admin.v1/src/main/resources/admin-api.yaml
```diff
@@ -3548,6 +3548,34 @@ paths:
...
+      operationId: validateSystemRole
+      parameters:
+        - $ref: '#/components/parameters/roleId'
+      responses:
+        200:
+          description: OK. Requested role name exists.
+          content: {}
+        404:
+          $ref: '#/components/responses/NotFound'
+      security:
```
vithu30, line 3570: Shall we define `500` status code also here, for documentation purpose.. In doc space, we need to display the all the possible response codes for this API call.

hisanhunais: Fixed with https://github.com/wso2/carbon-apimgt/pull/9804/commits/926300b8438fa22c702bc5b8334f65f20277b17f

vithu30 (approved):

features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/Throttling/Subscription/AddEdit.jsx
```diff
@@ -215,6 +224,7 @@ function AddEdit(props) {
                     unitTimeEdit = result.body.defaultLimit.bandwidth.unitTime;
                     typeEdit = result.body.defaultLimit.type;
                 }
+                setValidRoles(result.body.permissions.roles);
```
chanaka3d, line 227: We can use the same dispatcher to handle the roles without introducing the new state variables ( validRoles, invalidRoles and roleValidity). Use the same error handling method to show errors or disable the submit button.  Inside the formSaveCallback method, you can format the payload required to send to the backend.

chanaka3d: Let's go with the current implementation. It seems like getting rid of the extra state variables is too much work and doesn't add value except removing an extra rerender at the initial component mount.

chanaka3d (approved):

PR #9839, opened 2021-02-17
Adding initial faulty sub categories

ruks:
- Adding initial faulty sub-categories
- Read region ID from config or system property
- Identify user IP
- Fixed https://github.com/wso2/product-apim/issues/9961
- Fixed https://github.com/wso2/product-apim/issues/9962


components/apimgt/org.wso2.carbon.apimgt.common.gateway/src/main/java/org/wso2/carbon/apimgt/common/gateway/analytics/collectors/impl/SuccessRequestDataCollector.java
```diff
@@ -68,6 +70,11 @@ public void collectData() {
         Latencies latencies = provider.getLatencies();
         MetaInfo metaInfo = provider.getMetaInfo();
         String userAgent = provider.getUserAgentHeader();
+        String userIp = provider.getEndUserIP();
+        if (userIp == null) {
```
fazlan-nazeem, line 74: Can it be empty?

components/apimgt/org.wso2.carbon.apimgt.common.gateway/src/main/java/org/wso2/carbon/apimgt/common/gateway/analytics/collectors/impl/SuccessRequestDataCollector.java
```diff
@@ -48,7 +50,7 @@ public SuccessRequestDataCollector(AnalyticsDataProvider provider) {
         this(provider, new SuccessRequestDataPublisher());
     }
 
-    public void collectData() {
+    public void collectData() throws AnalyticsException {
```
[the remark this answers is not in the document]
ruks: DataNotFoundException or InvalidCategoryException exception can be thrown and AnalyticsException is the parent exception. Known cases are handled not to throw such exceptions. If it throws we skip the event.

components/apimgt/org.wso2.carbon.apimgt.common.gateway/src/main/java/org/wso2/carbon/apimgt/common/gateway/analytics/collectors/impl/SuccessRequestDataCollector.java
ruks: This is just to make sure send a default value of the userIP is not found. Because without userIP we can proceed not failed schema validation we set a default value. Base on the userIP identification and the way it provides this can be null. In the synapse, I don't think this will be null.

mushthaq33 (approved):

PR #9853, opened 2021-02-18
Add throttling support for webhooks

chanikag:
This PR is to add throttling support for webhooks APIs.

1. Connections count based throttling
2. Event based throttling


bhathiya:
@chanikag can you please add unit tests?

sachithKay (approved):

PR #9927, opened 2021-03-01
Add missing mandatory elements to OAS definitions

YasasRangika:
### Purpose
To solve below issues associated with support tickets.
**For OAS3:**
When importing the API via apictl or publisher the security and securitySchemes have been introduced. The implicit flow is added in this scenario. According to the OAS3 specification, the security or securitySchemes are not mandatory. But if we defined for code or implicit grants then we need to provide authorizationURL as it is a mandatory field.
**For OAS2:**
To solve critical error shows when validating auto updated swagger 2.0 definition after uploading the original definition to APIM-3.X.X.

### Goal
To prevent from https://github.com/wso2/product-apim/issues/9889 and https://github.com/wso2/product-apim/issues/9912 issues in future

### Approach
Add mock authorizatoinURL (https://test.com) to the authorizationUrl if it is not present and add empty scopes object if there is no scopes provided on OAS3 definitions. For OAS2 definitions, add an empty object "scopes: {}" when generating definition in APIM if there is no any scopes defined in uploaded definition.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS3Parser.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS3Parser.java
@@ -691,6 +691,37 @@ public class OAS3Parser extends APIDefinition {
     }
 
     /**
+     * Include authorizationUrl and scopes, mandatory elements to the OpenAPI definition if not present under
+     * securitySchemes in provided definition.
+     *
+     * @param swaggerContent OpenAPI Definition content
+     */
+    private String processSecuritySchemes(String swaggerContent) {
+        OpenAPI openAPI = getOpenAPI(swaggerContent);
+        Components components = openAPI.getComponents();
+        if (components != null) {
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS3Parser.java
```diff
@@ -690,6 +690,37 @@ private void populatePathParameters(Operation operation, String pathName) {
         }
     }
 
+    /**
+     * Include authorizationUrl and scopes, mandatory elements to the OpenAPI definition if not present under
+     * securitySchemes in provided definition.
+     *
+     * @param swaggerContent OpenAPI Definition content
```
wasuradananjith, line 697: Shall we add the @return as well to the java doc?

YasasRangika: Added in https://github.com/wso2/carbon-apimgt/pull/9927/commits/378729ec7b9401b4d317907185d27b13e98a52da

mushthaq33 (approved):

PR #9958, opened 2021-03-03
updating workflows with parallel builds

malinthaprasan:


.github/workflows/build-carbon.yml
```diff
@@ -0,0 +1,36 @@
+name: Build carbon-apimgt
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
bhathiya, line 9: workflow_dispatch:
    branches:
      - master

.github/workflows/build-product.yml
```diff
@@ -0,0 +1,62 @@
...
+
+on:
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+    branches:
+    - master
+  schedule:
+    - cron:  '0 20 * * *'
```
bhathiya, line 11: can't  set the branch here?

.github/workflows/build-product.yml
```diff
@@ -0,0 +1,62 @@
...
+        restore-keys: ${{ runner.os }}-m2
+    - name: Build carbon-apimgt without tests, skip AspectJ
+      run: mvn clean install --file pom.xml -Dmaven.test.skip=true -Dskip.aspectj=true
+    - name: Get carbon.apimgt.version
+      run: mvn help:evaluate -Dexpression=project.version -q -DforceStdout > ../CARBON_APIMGT_VERSION_FILE
+    - name: Print carbon.apimgt.version
+      run: |
+        echo $(cat ../CARBON_APIMGT_VERSION_FILE)
+    - name: Checkout Product-APIM
+      uses: actions/checkout@v1
```
bhathiya, line 50:
```suggestion
      uses: actions/checkout@v2
```

.github/workflows/build-carbon.yml
malinthaprasan: added

.github/workflows/build-product.yml
malinthaprasan: As per the doc "Scheduled workflows run on the latest commit on the default or base branch." which means the master branch.
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

.github/workflows/build-product.yml
malinthaprasan: fixed

.github/workflows/main.yml
```diff
@@ -1,32 +1,108 @@
...
-      with: 
-        repository: wso2/product-apim
-        ref: refs/heads/master
-    - name: Build Product-Apim with Maven
-      run: mvn clean install -Dcarbon.apimgt.version=$(cat ../CARBON_APIMGT_VERSION_FILE) -fae --file ../product-apim/pom.xml
+      - uses: actions/download-artifact@v2
+        with:
+          path: reports
+      - name: Publish Test Report
+        if: ${{ always() }}
```
bhathiya, line 104: can use just `if: always()`

.github/workflows/main.yml
```diff
@@ -1,32 +1,108 @@
...
+      fail-fast: false
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: '10'
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
```
bhathiya, line 55: can we run with JDK 11?

.github/workflows/main.yml
```diff
@@ -1,32 +1,108 @@
...
-        repository: wso2/product-apim
-        ref: refs/heads/master
-    - name: Build Product-Apim with Maven
-      run: mvn clean install -Dcarbon.apimgt.version=$(cat ../CARBON_APIMGT_VERSION_FILE) -fae --file ../product-apim/pom.xml
+      - uses: actions/download-artifact@v2
+        with:
+          path: reports
+      - name: Publish Test Report
+        if: ${{ always() }}
+        uses: malinthaprasan/action-surefire-report@v1
```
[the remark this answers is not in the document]
malinthaprasan: Yes the original version of the plugin (https://github.com/ScaCap/action-surefire-report) has some issues as it tries to post tests results from surefire plugin to the git action itself but since the token inside github actions doesn't have necessary privileges it fails. Because of this the reporting is failing. Therefore this plugin had to be extended to disable that and just add "echo" logs with the failing test cases.
The change is added here https://github.com/ScaCap/action-surefire-report/compare/master...malinthaprasan:master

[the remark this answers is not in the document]
bhathiya: shall we contribute it back with a config?

.github/workflows/main.yml
tharindu1st: but we can run integration tests

bhathiya: This  is a JDK 11 supported unofficial fork. https://github.com/awslabs/aws-lambda-powertools-java/issues/50 

But not sure if we should use it for prod.

PR #9965, opened 2021-03-05
Fix DefaultVersion 

tharindu1st:

PR #9977, opened 2021-03-08
Update the correct default API version when importing an API via APICTL

wasuradananjith:
## Purpose
Fixes: https://github.com/wso2/product-apim-tooling/issues/533

## Goals
Fixing the behaviour of the default API version when there is an API already inside the APIM which the default API version is true.

## Approach
- The default API version and the published default API version will be updated in the AM_API_DEFAULT_VERSION table.
- The registry artifacts will be updated accordingly as well.
- The new default API version will be added to the gateway by removing the old one.

## User stories
- If there is not any, default API in the APIM, when you import an API with the default API version, it will be assigned as the default one.
- If there is already an API with the default API version inside APIM, when you are importing an API with the default API version, it will be assigned as the default one by making the old API default version false.

## Test environment
- JDK 1.8.0_251
- Ubuntu 20.04 LTS

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIProvider.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/APIProvider.java
@@ -443,6 +443,33 @@ public interface APIProvider extends APIManager {
             throws DuplicateAPIException, APIManagementException;
 
     /**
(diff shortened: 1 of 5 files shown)


1akshitha (approved):

uvindra (approved):

wasuradananjith:
The issue which has been addressed from this PR has been fixed via https://github.com/wso2/carbon-apimgt/pull/9965. Hence, closing the PR

PR #10001, opened 2021-03-11
Fix subscription test failures

Sarangan0219:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AbstractAPIManager.java
```diff
@@ -453,22 +453,16 @@ protected API getApi(GovernanceArtifact artifact) throws APIManagementException
...
-            } else {
-                api.setSwaggerDefinition(getOpenAPIDefinition(identifier, organizationId));
-            }
-            if (api.getType() != null && APIConstants.APITransportType.GRAPHQL.toString().equals(api.getType())){
-                api.setGraphQLSchema(getGraphqlSchema(api.getId()));
-            }
-            //check for API visibility
-            if (APIConstants.API_GLOBAL_VISIBILITY.equals(api.getVisibility())) { //global api
-                return api;
-            }
```
malinthaprasan, line 471: is it needed to remove those?

Sarangan0219: When checking the usages of swagger definition from this returned API there are none. That's why I removed it. But there may be cases where I missed so added it again https://github.com/wso2/carbon-apimgt/pull/10001/commits/c63181d28305d4b74edc9cc8bb186fe25117b703

```
            if (APIConstants.API_GLOBAL_VISIBILITY.equals(api.getVisibility())) { //global api
                return api;
            }
```

With this check, we are not achieving anything. Thus removed it.

malinthaprasan (approved):

PR #10027, opened 2021-03-16
Handle event handling failure when hub secret is not defined.

chanikag:
This is to handle the scenarion when the hub secret is not defined


chanikag:
This is related to https://github.com/wso2/product-apim/pull/10288

tharindu1st:
@chanikag please include unit test for this scenario since this can test from unit test level.

tharindu1st (approved):

PR #10048, opened 2021-03-19
Fix application update issue with null string

RakhithaRR:
Fixes https://github.com/wso2/product-apim/issues/10282


malinthaprasan:
Let's fix from the application key retrieval REST API level to avoid a "null" coming to the UI's Callback URL text box.

malinthaprasan:
As discussed let's fix from the REST API GET application keys level to avoid getting a "null" to the UI's callback URL textbox. That should be the reason for submitting a null value to the "update keys" REST API.

malinthaprasan (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AbstractAPIManager.java
```diff
@@ -3239,6 +3239,9 @@ protected String getSearchQuery(String searchQuery) throws APIManagementExceptio
                                 if (StringUtils.isEmpty(oAuthApplicationInfo.getCallBackURL())) {
                                     oAuthApplicationInfo.setCallBackURL(storedOAuthApplicationInfo.getCallBackURL());
                                 }
+                                if (oAuthApplicationInfo.getCallBackURL().equalsIgnoreCase("null")) {
```
malinthaprasan, line 3242:
```suggestion
                                if ("null".equalsIgnoreCase(oAuthApplicationInfo.getCallBackURL())) {
```

PR #10128, opened 2021-03-26
check for null gatewayEnvironments API field

natechols:
I ran into an NPE here using generated Scala client code with API Manager 3.2.0 (which I guess runs v6.7.206 of this repo, but this part of the code is the same there and in master).  There's an easy client-side workaround for this of course but since it took me an hour to figure out I thought I'd post what I believe is the proper fix - feel free to do as you wish with it.


malinthaprasan:
Hi @natechols. thank you very much for your contribution!!

malinthaprasan (approved):

PR #10140, opened 2021-03-29
Set Vhost when importing API and API product

renuka-fernando:
### Purpose
$subject

#### Sample deployment_environments.yaml
Sample deployment_environments.yaml in the APICTL project

```yaml
type: deployment_environments
version: v4.0.0
data:
 -
  deploymentEnvironment: Production and Sandbox
  deploymentVhost: localhost   # OPTIONAL
  displayOnDevportal: true
 -
  deploymentEnvironment: us-region
  deploymentVhost: us.wso2.com # OPTIONAL
  displayOnDevportal: true
```

when `deploymentVhost` is not present in the file, the default (first vhost) is used to deploy the revision.

When exporting API no vhost set in the file deployment_environments.yaml.
```yaml
type: deployment_environments
version: v4.0.0
data:
 -
  deploymentEnvironment: Production and Sandbox
  displayOnDevportal: true
 -
  deploymentEnvironment: us-region
  displayOnDevportal: true
```

Do not set vhost in deployment environment file when export API (or API Project). So when importing the exported API, the default vhost of the new environment is selected.


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/ImportUtils.java
```diff
@@ -410,12 +410,19 @@ public static API importApi(String extractedFolderPath, APIDTO importedApiDTO, B
                 if (deploymentNameElement != null) {
                     String deploymentName = deploymentNameElement.getAsString();
                     if (gatewayEnvironmentsSet.contains(deploymentName)) {
+                        JsonElement deploymentVhostElement = deploymentJson.get(ImportExportConstants.DEPLOYMENT_VHOST);
+                        if (deploymentVhostElement == null) {
+                            throw new APIManagementException(
+                                    String.format("Required field \"%s\" is not defined in the deployment",
```
1akshitha, line 416: instead of throwing an exception, can't we assign the default value localhost?

pubudu538: Yes, Vhost should be an optional value.

wasuradananjith: +1

renuka-fernando: Thanks all fixed with https://github.com/wso2/carbon-apimgt/pull/10140/commits/2534b78073e669303fd191cb5f72e3fcbc59faee

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
```diff
@@ -9759,6 +9759,22 @@ public static String getStoreUrl() throws APIManagementException {
...
 
+    /**
+     * Get default (first) vhost of the given environment
+     * @param environmentName name of the environment
+     * @return default vhost of environment
+     */
+    public static VHost getDefaultVhostOfEnvironment(String environmentName) throws APIManagementException {
+        Map<String, Environment> environments = getEnvironments();
+        if (environments.get(environmentName) == null) {
+            throw new APIManagementException("Configured environment not found: " + environmentName);
```
1akshitha, line 9770: throwing such exception will break the import. Current implementation is done in a way that if there is a problem in the deployment environment, that environment will be skipped without deployment. 

@uvindra @pubudu538 @wasuradananjith shall we preserve the same experience?

1akshitha (changes requested):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
renuka-fernando: Since we are checking `if (gatewayEnvironmentsSet.contains(deploymentName))` before calling this, this exception will not be thrown, but the next exception in line 9773 will thrown, if not vhosts are configured in the environment

1akshitha: We will proceed with your way. If the vhost entered is faulty, import should abort. not skip that environment.

1akshitha (approved):

wasuradananjith (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
renuka-fernando: I think we can also remove calling `APIUtil. getDefaultVhostOfEnvironment()` since it makes another DB query, we can use existing already queried environments, and so I will remove this added util function. And will throw an exception when no vhosts are defined in the gateway environment

renuka-fernando:
Do not set vhost in deployment environment file when export API (or API Project). So when importing the exported API, the default vhost of the new environment is selected.

uvindra:
> Do not set vhost in deployment environment file when export API (or API Project). So when importing the exported API, the default vhost of the new environment is selected.

I think it might be restrictive to assume that this will always be the case when a user is importing to a new environment. The best option seems to be allowing the vhost to remain in the environment.yaml, but at the same time we can allow users to override this value to any that they choose via param file at the time of import. This seems to give the user either option to achieve whatever use case they might have

pubudu538:
Unless we make the vhost value optional, users have to do changes to the exported revision to change the hostname value. If we go with the optional field, then only we can export/import API without doing any changes to the project.

uvindra:
> Unless we make the vhost value optional, users have to do changes to the exported revision to change the hostname value. If we go with the optional field, then only we can export/import API without doing any changes to the project.

As I mentioned above, users dont need to change the project but can override the vhost via external param.yaml file if that is their requirement. I get that the vhost value can be optional at the time of importing. But if we drop it altogether at the time of export its pretty much discarded totally

renuka-fernando:
Related integration tests in the folk: https://github.com/renuka-fernando/carbon-apimgt/actions/runs/708033699
Failed in Actions also failed in master of folk github action: https://github.com/renuka-fernando/carbon-apimgt/actions/runs/709362551

Also failed same tests in local for both branches
`master` and `import-export-vhost`

Master local branch logs: [log.tar.gz](https://github.com/wso2/carbon-apimgt/files/6247604/log.tar.gz)

renuka-fernando:
We discussed with @uvindra and @pubudu538  offline and we decided to support overriding vhost with param file via https://github.com/wso2/product-apim-tooling/issues/679 and when exporting API revision we are not setting the vhost in the `deployment_environments.yaml` file. But when importing API if vhost is present in the file `deployment_environments.yaml` then we set the vhost of it otherwise wet the default vhost of the gateway environment.

uvindra (approved):

pubudu538 (approved):

PR #10161, opened 2021-03-31
Fix location rewrite when ISKMReverseProxy enabled

CrowleyRajapakse:
## Issues
Fixes https://github.com/wso2/product-apim/issues/10446


tharindu1st:
not nedded for newer version

PR #10162, opened 2021-03-31
[UX] - Update api details UI

chanaka3d:
Fixing: https://github.com/wso2/product-apim/issues/10680

The buisness plans boxes are not matching the proposed design.

Requires rest api change.

**Existing data**
Screen Shot 2021-03-31 at 2 30 55 PM

**Requires additional info to satisfy the following design**
Screen Shot 2021-03-31 at 2 31 28 PM

image

image

image


tmkasun (approved):

chanaka3d:
For the async APIs, this is how the business plans and the endpoint URL looks. And how the tryout button changes.

Screen Shot 2021-04-02 at 8 36 02 AM

This is what happens when the Tryout button clicks when the following conditions satisfies.

User logged in.
API is not subscribed to any app.
DefaultApplication is there.
The API is not async api
The API is not advertised.

Screen Shot 2021-04-02 at 3 24 58 AM

tmkasun:
We have a tryout page for webhooks, So if it's a webhook/web sub I think we can keep the tryout button, Anyway showing the definition in text here is not very useful here IMO , @sajithshn @lasanthaS ?

image

IMO, This message should be intuitive in the Try Out page rather than showing it as a text

PR #10197, opened 2021-04-01
Fix issue API is not visible in Publisher for searching

YasasRangika:
### Purpose
To solve the API is not visible in Publisher for searching after adding Devportal visibility roles.

### Goal
Fixes: https://github.com/wso2/product-apim/issues/10597

### Approach
Used system registry instead of using user registry when searching content.


YasasRangika:
Workflows: https://github.com/YasasRangika/carbon-apimgt/actions/runs/709397930

malinthaprasan (approved):

PR #10200, opened 2021-04-02
Fix supporting long values in policy templates

tharikaGitHub:
This PR fixes https://github.com/wso2/product-apim/issues/9919

This is reproduced in Public branch not during creation of the policy from UI, but during server startup.


tharikaGitHub:
Workflows: https://github.com/tharikaGitHub/carbon-apimgt/actions/runs/710389214

tharindu1st (approved):

PR #10229, opened 2021-04-03
Change the color and size of the API overview stepper.

menakaj:
Fixes https://github.com/wso2/product-apim/issues/10731

Screenshot - 2021-04-03T124002 731


erangatl (changes requested):
Usually green colour is used to give the feeling of complete. I would suggest to have a green colour.

tmkasun:
@menakaj We had an offline chat regarding the color scheme, as @erangatl  has mentioned green color has been used for indicating the completed steps as it's the most frequently used color for success/completion messages, Yet we decided to get these colors from the theme so that if someone prefers a different color scheme for active, completed & inactive states that can change it via the theme.

Thanks for the PR :) will keep the git issue open until we do the above fix

PR #10244, opened 2021-04-06
Fix UI issues

Krishanx92:
Fix https://github.com/wso2/product-apim/issues/10745
Fix https://github.com/wso2/product-apim/issues/10609
Fix https://github.com/wso2/product-apim/issues/10757
Fix https://github.com/wso2/product-apim/issues/10787

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Shared/CustomSplitButton.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Shared/CustomSplitButton.jsx
@@ -24,17 +24,14 @@ export default function CustomSplitButton(props) {
     const anchorRef = React.useRef(null);
     const [selectedIndex, setSelectedIndex] = React.useState(1);
 
-    const handleClick = () => {
-        if (`${options[selectedIndex]}` === 'Save') {
+    const handleClick = (event, index) => {
+        setSelectedIndex(index);
+        setOpen(false);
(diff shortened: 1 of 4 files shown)


tmkasun (approved):

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Shared/CustomSplitButton.jsx
```diff
@@ -91,7 +92,7 @@ export default function CustomSplitButton(props) {
                                             <MenuItem
                                                 key={option}
                                                 selected={index === selectedIndex}
-                                                onClick={(event) => handleMenuItemClick(event, index)}
+                                                onClick={(event) => handleClick(event, index)}
```
tmkasun, line 95: Using array index as a key to find an element is a bit flaky, Can't we use an id here, and also `key` should be a string. not an object, hope `option` is a string value.

> A “key” is a special string attribute you need to include when creating lists of elements.
[Docs](https://reactjs.org/docs/lists-and-keys.html#basic-list-component)

tmkasun (approved):

features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Shared/CustomSplitButton.jsx
Krishanx92: Yes,
const options = ['Save and deploy', 'Save'];
Option is string value.

PR #10247, opened 2021-04-06
set AM_API_COMMENTS.ENTRY_POINT default to DEVPORTAL

isharac:
Fix https://github.com/wso2/product-apim/issues/10090


isharac:
this fix is only required in migration scripts. Hence closing.

PR #10274, opened 2021-04-08
Align KID values of the backend JWT with the JWKS endpoint

msm1992:
Fixes https://github.com/wso2/product-apim/issues/9881

test fixes : https://github.com/wso2/product-apim/pull/10828


tharindu1st:
This needs to be removed from the 4.0.0 release.

PR #10275, opened 2021-04-08
Fix APIM-10758: Subscribing to default endpoint gives an error

chanikag:
This is to fix https://github.com/wso2/product-apim/issues/10758

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Resources/components/AddOperation.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Resources/components/AddOperation.jsx
@@ -290,7 +290,7 @@ function AddOperation(props) {
                         name='target'
                         value={newOperations.target}
                         onChange={({ target: { name, value } }) => newOperationsDispatcher(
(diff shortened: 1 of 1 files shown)


features/apimgt/org.wso2.carbon.apimgt.publisher.feature/src/main/resources/publisher/source/src/app/components/Apis/Details/Resources/components/AddOperation.jsx
```diff
@@ -290,7 +290,7 @@ function AddOperation(props) {
                         name='target'
                         value={newOperations.target}
                         onChange={({ target: { name, value } }) => newOperationsDispatcher(
-                            { type: name, value: value.startsWith('/') ? value : `/${value}` },
+                            { type: name, value },
```
tmkasun, line 293: This string comprehension append a `/` to the path if the user forgot to add it, I'm not sure how removing that going to fix `default` path issue

chanikag: When it comes to webhooks adding a '/' doesn't make sense for topics. However the original issue that is mentioned is eventhough the default topic name is _default in UI in DB it is added as /_default and that causes the runtime issue. Therefore as a fix we can remove adding / for topics

PR #10278, opened 2021-04-08
Fix sorting and spacing for scope mappings

RakhithaRR:
This PR fixes the sorting and spacing issues in Scope Mappings table.
Fixes https://github.com/wso2/product-apim/issues/10604

### DIFF
--- a/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
@@ -121,6 +121,7 @@ function AdminTableBody(props) {
         setSelected(newSelected);
     };
     const isSelected = (name) => selected.indexOf(name) !== -1;
+    const emptyRows = rowsPerPage - Math.min(rowsPerPage, rows.length - page * rowsPerPage);
 
     return (
         <TableBody>
@@ -162,6 +163,11 @@ function AdminTableBody(props) {
                         </TableRow>
                     );
                 })}
+            {emptyRows > 0 && (
+                <TableRow style={{ height: 60 * emptyRows }}>
+                    <TableCell colSpan={6} />
+                </TableRow>
+            )}
         </TableBody>
     );
 }
--- a/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
+++ b/features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
@@ -45,8 +45,8 @@ function extractRoleName(role) {
  */
 function descendingComparator(a, b, orderBy) {
     if (orderBy === 'role') {
-        const roleA = extractRoleName(a, orderBy);
-        const roleB = extractRoleName(b, orderBy);
+        const roleA = extractRoleName(a);
+        const roleB = extractRoleName(b);
         if (roleB < roleA) {
             return -1;
         }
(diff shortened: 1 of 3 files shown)


features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
```diff
@@ -106,7 +121,6 @@ function AdminTableBody(props) {
         setSelected(newSelected);
     };
     const isSelected = (name) => selected.indexOf(name) !== -1;
-    const emptyRows = rowsPerPage - Math.min(rowsPerPage, rows.length - page * rowsPerPage);
```
tmkasun, line 109: @RakhithaRR Can you please attach a screenshot , Without this , the table will be shrunk to available number of rows not showing full table

features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
```diff
@@ -34,11 +44,16 @@ import { useTableContext } from './AdminTableContext';
  * @returns
  */
 function descendingComparator(a, b, orderBy) {
-    if (b[orderBy] < a[orderBy]) {
-        return -1;
-    }
-    if (b[orderBy] > a[orderBy]) {
-        return 1;
+    if (orderBy === 'role') {
+        const roleA = extractRoleName(a, orderBy);
```
tmkasun, line 48: In the method definition, you accepting only 1 argument, here passing 2 args? , This `descendingComparator ` meant to be a generic comparator given the `orderBy ` value (column name), But here we only use it for role comparison ?

tmkasun (approved):

features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
RakhithaRR: Earlier there were 6 empty rows in the table as mentioned in the issue.
Screenshot from 2021-04-08 17-24-22

tmkasun: yes that was intentional, because when there is only 1 role this table will have only 1 row wich does not looks good

features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
RakhithaRR: Removed extra args with c198645.

The only sortable column in the Scope Mappings table is the `Roles` column. Earlier the comparator was looking for `orderBy` attribute value in the args passed (a and b) to this function but they were missing. Here, I've updated it to check whether orderBy is `role` and if so,  extract the correct role name from the args and compare them.

features/apimgt/org.wso2.carbon.apimgt.admin.feature/src/main/resources/admin/source/src/app/components/RolePermissions/AdminTable/AdminTableBody.jsx
RakhithaRR: I see. However having a large number of empty rows when the row count is low and rows per page is high also does not look that good.

tmkasun: I see, but why would someone select 100 Rows per page when there are only 2-3 roles in total?  the worst cases here is having 9 empty rows when rows per page are 10, having 10 roles when rows per page is 50 like that

RakhithaRR: Reverted this change with c0e0fb9

PR #10291, opened 2021-04-09
Disable the incorrect transition

RAVEENSR:
Fixes https://github.com/wso2/product-apim/issues/10889 issue.

Before:

before

After:

after


tmkasun:
I think this is a valid transition, Those transition buttons are generated from REST response, I think we would support publishing directly from prototyped state (User has to change the endpoint type)

RAVEENSR:
> I think this is a valid transition, Those transition buttons are generated from REST response, I think we would support publishing directly from prototyped state (User has to change the endpoint type)

We have restricted this transition from the REST API level. That is why we needed to change this UI. The relevant REST API fix was done from https://github.com/wso2/carbon-apimgt/pull/10238

malinthaprasan:
We'll have to allow this transition as it is a valid one. Will need to slightly modify the REST API fix.

PR #10312, opened 2021-04-10
Fix the issue of API Creators can delete Published APIs

YasasRangika:
### Purpose
Fixes: https://github.com/wso2/product-apim/issues/10275

### Goals
To restrict API Creators can delete Published or Deprecated APIs

### Approach
Checked user's role and the status of the API before calling the deleteAPI.

Actions: https://github.com/wso2/carbon-apimgt/actions/runs/735521276

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/impl/ApisApiServiceImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/impl/ApisApiServiceImpl.java
@@ -531,7 +531,7 @@ public class ApisApiServiceImpl implements ApisApiService {
             if (comment != null) {
                 String[] tokenScopes = (String[]) PhaseInterceptorChain.getCurrentMessage().getExchange()
                         .get(RestApiConstants.USER_REST_API_SCOPES);
-                if (Arrays.asList(tokenScopes).contains("apim:admin") || comment.getUser().equals(username)) {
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/impl/ApisApiServiceImpl.java
```diff
@@ -1511,6 +1511,13 @@ public Response deleteAPI(String apiId, String ifMatch, MessageContext messageCo
                 RestApiUtil.handleConflict("Cannot remove the API because following resource paths " +
                         usedProductResources.toString() + " are used by one or more API Products", log);
             }
+
+            // check user has publisher role and API is in published or deprecated state
+            if (!APIUtil.isRoleExistForUser(username, "Internal/publisher") && (
+                    APIConstants.PUBLISHED.equalsIgnoreCase(api.getStatus()) || APIConstants.DEPRECATED
+                            .equalsIgnoreCase(api.getStatus()))) {
+                RestApiUtil.handleConflict(username + " cannot remove the API", log);
+            }
```
malinthaprasan, line 1520: We cannot use "Internal/publisher" role always to check for a publisher role. This can be changed from scope assignments.

We need to check this using scopes. Need to check user has "apim:api_publish" scope instead of `APIUtil.isRoleExistForUser(username, "Internal/publisher")`

Scopes of the user can be found using:
```
            String[] tokenScopes =
                    (String[]) PhaseInterceptorChain.getCurrentMessage().getExchange()
                            .get(RestApiConstants.USER_REST_API_SCOPES);
```

malinthaprasan (changes requested):

malinthaprasan (approved):

PR #10365, opened 2021-04-12
Fix issue with doc content indexing

msm1992:
Fixes https://github.com/wso2/product-apim/issues/10946

Approach :

* Here I have introduced a new toggle field (within the code, the field does not need to be added to document rxt)
* The issue earlier was because the document indexer did not get invoked because there was no change on the actual document artifact upon a document content update.
* The new toggle field will contain a boolean value which will be inverted each time a new document content update is done, thus a change in underlying document artifact and the document indexer will get triggered.


malinthaprasan (approved):

malinthaprasan:
Action https://github.com/msm1992/carbon-apimgt/actions/runs/740845526

PR #10400, opened 2021-04-15
Fix #10949

Arshardh:
Fixes wso2/product-apim#10949


Arshardh:
Fixes https://github.com/wso2/product-apim/issues/10949

malinthaprasan (approved):

sanjeewa-malalgoda:
Even result set null or not we use same compoundResult to set results in return object. Im not sure how this fixed issue. Maybe i understood this wrong way.

Arshardh:
The result count only what was wrong, the actual results were fine since earlier.

tmkasun (approved):

PR #10425, opened 2021-04-19
Fix onboarding page permissions for publisher user

tharikaGitHub:
Fixes https://github.com/wso2/product-apim/issues/11059 and adds an improvement to it so that a publisher user will not be able to click on any create api option. When the card is flipped, the below text is shown.

Screenshot 2021-04-18 at 12 27 45


tharikaGitHub:
Closing as the fix will be different based on https://github.com/wso2/product-apim/issues/11059#issuecomment-822536288

PR #10456, opened 2021-04-27
Rename Publisher REST API packages

malinthaprasan:
Rename Publisher REST API java packages with removing "v1"

Workflows: https://github.com/malinthaprasan/carbon-apimgt/actions/runs/789032363


malinthaprasan:
We'll not be doing this due to the customizations done in 8.x branch (choreo) and this package renaming can cause many conflicts.

PR #10459, opened 2021-04-28
Adding application/json content type to export API Rest API

HiranyaKavishani:
Fix : https://github.com/wso2/product-apim-tooling/issues/360

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
@@ -4269,6 +4269,10 @@ paths:
               schema:
                 type: string
           content:
+            application/json:
+              schema:
+                type: string
+                format: binary
             application/zip:
               schema:
                 type: string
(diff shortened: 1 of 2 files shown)


chashikajw (approved):

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
```diff
@@ -4269,6 +4269,10 @@ paths:
               schema:
                 type: string
           content:
+            application/json:
+              schema:
+                type: string
+                format: binary
```
malinthaprasan, line 4275: How can we send a json with binary format..

malinthaprasan:
I thought the above issue is already fixed in the latest pack from https://github.com/wso2/carbon-apimgt/pull/10444

Could you please double check in the latest pack?

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
HiranyaKavishani: Can't we use binary format for non-entity-body?

HiranyaKavishani:
I only built the publisher webApp from latest and tested with GA pack.I Could able to reproduce the issue and the fix was fixed that issue. When I analyzed fix https://github.com/wso2/carbon-apimgt/pull/10444 , seems it is fixed at error response flow by changing `header(RestApiConstants.HEADER_CONTENT_TYPE, RestApiConstants.DEFAULT_RESPONSE_CONTENT_TYPE) -> type(MediaType.APPLICATION_JSON_TYPE)` where the fix has gone for util method. I'll check it and close the PR, if it is fixed

HiranyaKavishani:
Tested in latest pack. Yes it has been fixed. Thanks @malinthaprasan

PR #10479, opened 2021-05-10
Support importing keys for KMs configured with 'no oauth apps'

praminda:
Fixes wso2/product-apim#11125

if oauth app creation is disabled for the KM, we should treat it as an out of band app and use the mapped keys provided with the app.
implementation to avoid interface changes.
new overloaded method for `requestApprovalForApplicationRegistration` was introduced with `isImported` arg to identify if the request is coming from `import-app` API. This is then used to provide error msg when rest API is being used to generate keys for `create oauth apps` disabled KMs.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
@@ -4079,6 +4079,7 @@ public class APIConsumerImpl extends AbstractAPIManager implements APIConsumer {
     }
 
     @Override
+    @Deprecated
     public Map<String, Object> requestApprovalForApplicationRegistration(String userId, String applicationName,
                                                                          String tokenType, String callbackUrl,
                                                                          String[] allowedDomains, String validityTime,
@@ -4086,8 +4087,8 @@ public class APIConsumerImpl extends AbstractAPIManager implements APIConsumer {
                                                                          String jsonString,
                                                                          String keyManagerName, String tenantDomain)
             throws APIManagementException {
-        return requestApprovalForApplicationRegistration(userId, applicationName,tokenType, callbackUrl, allowedDomains,
-                validityTime, tokenScope, groupingId, jsonString, keyManagerName, tenantDomain, false);
+        return requestApprovalForApplicationRegistration(userId, applicationName, tokenType, callbackUrl,
+                allowedDomains, validityTime, tokenScope, groupingId, jsonString, keyManagerName, tenantDomain, false);
     }
 
     /**
@@ -4103,7 +4104,7 @@ public class APIConsumerImpl extends AbstractAPIManager implements APIConsumer {
                                                                          String tokenScope, String groupingId,
                                                                          String jsonString,
                                                                          String keyManagerName, String tenantDomain,
-                                                                         boolean isImported)
+                                                                         boolean isImportMode)
     throws APIManagementException {
 
         boolean isTenantFlowStarted = false;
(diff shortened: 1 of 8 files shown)


uvindra (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
```diff
@@ -4086,8 +4102,9 @@ public void removeApplication(Application application, String username) throws A
                                                                          String[] allowedDomains, String validityTime,
                                                                          String tokenScope, String groupingId,
                                                                          String jsonString,
-                                                                         String keyManagerName, String tenantDomain)
-            throws APIManagementException {
+                                                                         String keyManagerName, String tenantDomain,
+                                                                         boolean isImported)
```
malinthaprasan, line 4106: shall we rename this as `mapKeysIfKMAppCreationDisabled`?

components/apimgt/org.wso2.carbon.apimgt.impl/src/test/java/org/wso2/carbon/apimgt/impl/APIConsumerImplTest.java
```diff
@@ -1041,7 +1041,7 @@ public void testRequestApprovalForApplicationRegistration() throws APIManagement
         try {
             apiConsumer
                     .requestApprovalForApplicationRegistration("1", "app1", "access", "identity.com/auth", null, "3600",
-                            "api_view", "2", null, "default", null);
+                            "api_view", "2", null, "default", null, false);
```
malinthaprasan, line 1044: Do we need these changes as there's an overloaded method

malinthaprasan (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
praminda: current usage of the argument is to map the keys, but kept the name generic so it can be used any time we need to identify if the flow is an import flow or normal key generation flow.

components/apimgt/org.wso2.carbon.apimgt.impl/src/test/java/org/wso2/carbon/apimgt/impl/APIConsumerImplTest.java
praminda: there's no need but orginally thought of deprecating the original method as it doesn't give much value to the interface.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
malinthaprasan: Sure.. but "isImported" does not clearly say to me this is App import flow from CTL.. For me it's like we need to indicated the app is already imported.. not sure its only me though.. May be better to use "importMode" true/false ?

praminda: How about `isImporting`? looks more conventional and explains what going on... WDYT? 😉

praminda: Updated to `isImportMode`

malinthaprasan (approved):

PR #10529, opened 2021-05-21
Incorporate organization in updateAPI

Shakila:


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/ApiMgtDAO.java
```diff
@@ -7202,7 +7203,7 @@ private void handleException(String msg, Throwable t) throws APIManagementExcept
         return urlMappings;
     }
 
-    public Set<URITemplate> getURITemplatesOfAPI(APIIdentifier identifier)
+    public Set<URITemplate> getURITemplatesOfAPI(APIIdentifier identifier, String organization)
```
chamilaadhi, line 7206: this method needs to be changed to work with uuid of the api. We could remove the organization @tharikaGitHub can u check wether it is changed

PR #10607, opened 2021-06-16
[8.x] Remove scopes when importing/updating openapi definitions

RakhithaRR:
This PR introduces a new REST API parameter to remove scopes when importing or updating an openapi definition.

```
importScopes:
      name: importScopes
      in: query
      description: |
        Decides whether we need to include the scopes in the swagger deifnition.
      schema:
        type: boolean
        default: true
```

This parameter will be used in POST `/apis/import-openapi` and PUT `apis/{apiId}/swagger`. 

When `importScopes` is `false`, the scope requirements under resources and scope definitions under security definitions in the original swagger definition will be removed. This will in turn prevent any local scopes being created for that API.

Related issue: https://github.com/wso2-enterprise/choreo/issues/5079


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
```diff
@@ -584,6 +584,7 @@ paths:
         - $ref: '#/components/parameters/organizationId'
         - $ref: '#/components/parameters/apiId'
         - $ref: '#/components/parameters/If-Match'
+        - $ref: '#/components/parameters/includeScopes'
```
malinthaprasan, line 587: shouldn't this be "importScopes"

RakhithaRR: Fixed with e137989

malinthaprasan (approved):

PR #10624, opened 2021-06-23
Fix the error response if the feign service call to token endpoint fails

vithu30:
Currently the `POST /applications/{applicationId}/oauth-keys/{keyMappingId}/generate-token` REST API is used to generate token from Devportal UI. If any error occurred during the token generation it returns a general error response as follows:

```json
{
   "code":900967,
   "message":"General Error",
   "description":"Server Error Occurred",
   "moreInfo":"","error":[]
}
```

This PR fixes it as:

```json
{
    "code": 900967,
    "message": "Bad Request",
    "description": "Error occurred while calling token endpoint - No Registered IDP found for the JWT with issuer name : https://{okta-domain}/oauth2/default",
    "moreInfo": "",
    "error": []
}
```

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/kmclient/KMClientErrorDecoder.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/kmclient/KMClientErrorDecoder.java
@@ -18,18 +18,39 @@ package org.wso2.carbon.apimgt.impl.kmclient;
 
 import feign.Response;
 import feign.codec.ErrorDecoder;
+import org.apache.commons.io.IOUtils;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+import java.io.IOException;
 
 import static feign.FeignException.errorStatus;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 public class KMClientErrorDecoder implements ErrorDecoder {
 
     @Override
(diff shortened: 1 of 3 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/kmclient/KMClientErrorDecoder.java
```diff
@@ -18,18 +18,39 @@
...
+                errorDescription = response.reason();
+            }
+            if (response.status() >= 400 && response.status() <= 499) {
+                return new KeyManagerClientException(response.status(), errorDescription);
+            }
+            if (response.status() >= 500 && response.status() <= 599) {
+                return new KeyManagerClientException(response.status(), errorDescription);
+            }
+        } catch (IOException | ParseException e) {
+            e.printStackTrace();
```
Nirothipan, line 53: why not log?

Nirothipan: Also, shall we use response.reason(); if exception occurs, just to wrap the IO and parse logic in try-catch adnd leave the rest as it was.

vithu30: We don't need to log this as the error is when parsing the `response.body().asInputStream()`. If there are errors when processing `response.body().asInputStream()` set the `error_description` as `response.reason()`

Nirothipan (approved):

PR #10694, opened 2021-07-22
Fix issue in passing headers, query/path params to lamda function from Gateway

vithu30:
Fixes https://github.com/wso2/product-apim/issues/11430 and sends `httpMethod` and `path` to AWS Lambda function.

The sample payload send to Lambda function would look like:
```json
{
    "headers": {
        "Accept": "*/*",
        "Accept-Encoding": "gzip, deflate, br",
        "Accept-Language": "en,en-US;q=0.5",
        "activityid": "401688ab-a955-42ee-b16e-96ab26ff31ec"
    },
    "pathParameters": {
        "xxxx": "xxxxx"
    },
    "queryStringParameters": {
        "xxxx": "xxxxx"
    },
    "body": {
        "key1": "value1",
        "key2": "value2"
    },
    "httpMethod": "POST",
    "path": "/resource/{path}"
}
```


tharikaGitHub (approved):

binodmx:
You need to add following config in `deployment.toml` to send the request parameters to AWS Lambda function.

```
[apim.lambda_mediator_config]
pass_request_params = true
```

vithu30:
> You need to add following config in `deployment.toml` to send the request parameters to AWS Lambda function.
> 
> ```
> [apim.lambda_mediator_config]
> pass_request_params = true
> ```

We don't need to add this configuration (in the public fix). This will the default payload format.

PR #10698, opened 2021-07-28
Improve deployment status calls

AmaliMatharaarachchi:
Change deployment info service from post to put
Remove duplicate entries from incoming deployment status payload

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
@@ -3431,7 +3431,7 @@ public class SQLConstants {
                 " INSERT INTO AM_DEPLOYMENT_REVISION_MAPPING (NAME, VHOST, REVISION_UUID, DISPLAY_ON_DEVPORTAL)" +
                         " VALUES (?,?,?,?)";
         public static final String ADD_DEPLOYED_API_REVISION =
-                " INSERT INTO AM_DEPLOYED_REVISION (NAME, VHOST, REVISION_UUID)" +
+                " REPLACE INTO AM_DEPLOYED_REVISION (NAME, VHOST, REVISION_UUID)" +
                         " VALUES (?,?,?)";
         public static final String DELETE_API_REVISION_DEPLOYMENTS_MAPPING_BY_REVISION_UUID =
                 " DELETE FROM AM_DEPLOYMENT_REVISION_MAPPING WHERE REVISION_UUID = ?";
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
@@ -3431,7 +3431,7 @@ public class SQLConstants {
                 " INSERT INTO AM_DEPLOYMENT_REVISION_MAPPING (NAME, VHOST, REVISION_UUID, DISPLAY_ON_DEVPORTAL)" +
                         " VALUES (?,?,?,?)";
         public static final String ADD_DEPLOYED_API_REVISION =
-                " REPLACE INTO AM_DEPLOYED_REVISION (NAME, VHOST, REVISION_UUID)" +
+                "INSERT INTO AM_DEPLOYED_REVISION (NAME, VHOST, REVISION_UUID)" +
                         " VALUES (?,?,?)";
         public static final String DELETE_API_REVISION_DEPLOYMENTS_MAPPING_BY_REVISION_UUID =
                 " DELETE FROM AM_DEPLOYMENT_REVISION_MAPPING WHERE REVISION_UUID = ?";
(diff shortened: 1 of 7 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
```diff
@@ -3431,7 +3431,7 @@
                 " INSERT INTO AM_DEPLOYMENT_REVISION_MAPPING (NAME, VHOST, REVISION_UUID, DISPLAY_ON_DEVPORTAL)" +
                         " VALUES (?,?,?,?)";
         public static final String ADD_DEPLOYED_API_REVISION =
-                " INSERT INTO AM_DEPLOYED_REVISION (NAME, VHOST, REVISION_UUID)" +
+                " REPLACE INTO AM_DEPLOYED_REVISION (NAME, VHOST, REVISION_UUID)" +
```
sumedhassk, line 3434: If we are using `REPLACE` do we need line https://github.com/wso2/carbon-apimgt/pull/10698/files#diff-19a032cb823fd4f31e79376db6427e305b79949a26979f42d5e7ea0c607cdff2R8802?  i.e ` apiMgtDAO.removeDeployedAPIRevision(apiId, environmentsToRemove);`

components/apimgt/org.wso2.carbon.apimgt.internal.service/src/main/java/org/wso2/carbon/apimgt/internal/service/impl/ApisApiServiceImpl.java
```diff
@@ -92,36 +92,38 @@ public Response apisGet(String xWSO2Tenant, String apiId, String context, String
...
+                for (DeployedEnvInfoDTO deployedEnvInfoDTO : deployedAPIRevisionDTO.getEnvInfo()) {
+                    DeployedAPIRevision deployedAPIRevision = new DeployedAPIRevision();
+                    deployedAPIRevision.setRevisionUUID(revisionUUID);
+                    String environment = deployedEnvInfoDTO.getName();
+                    if (environments.get(environment) == null) {
+                        RestApiUtil.handleBadRequest("Gateway environment not found: " + environment, log);
+                    }
+                    deployedAPIRevision.setDeployment(environment);
+                    deployedAPIRevision.setVhost(deployedEnvInfoDTO.getVhost());
+                    if (org.apache.commons.lang3.StringUtils.isEmpty(deployedEnvInfoDTO.getVhost())) {
```
sumedhassk, line 116: Wouldn't it be better to have this as an import?

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
AmaliMatharaarachchi: it's required. because there might be old records with different revisionuuids for the same API in same env. replace won't catch it because of the different revisionUUID

tharindu1st: lets change this to remove and insert model

AmaliMatharaarachchi: We did the removing before adding from here even earlier. https://github.com/wso2/carbon-apimgt/pull/10698/files#diff-19a032cb823fd4f31e79376db6427e305b79949a26979f42d5e7ea0c607cdff2R8802. However I noticed there were intermittent error logs saying "duplicate entries violating primary key constraint." I suspected that it can be due to duplicate entries in the same batch. Hence added duplicates removing logic. To be on the safe side added "replace" here as well.  I guess we can keep it as insert since we are removing duplicates from the message.

components/apimgt/org.wso2.carbon.apimgt.internal.service/src/main/resources/api.yaml
```diff
@@ -550,7 +550,7 @@ paths:
   # Deployed revision information
   #--------------------------------------------
   /apis/deployed-revision:
-    post:
+    put:
```
malinthaprasan, line 553: /apis/deployed-revisions:
    patch:

AmaliMatharaarachchi, line 553: make it patch. and make /apis/deployed-revisions

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/constants/SQLConstants.java
malinthaprasan: Lets not use "REPLACE" as it can swallow and difficult to find inconsistencies that can happen.

AmaliMatharaarachchi: fixed with https://github.com/wso2/carbon-apimgt/pull/10698/commits/661bcb9e2f6b0a28662af8e1acc8d42bab5af7e0

malinthaprasan (approved):

PR #10702, opened 2021-07-29
Adding new interface for organization deletion

HiranyaKavishani:
This change will add a new interface to delete org collection

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
@@ -150,7 +150,7 @@ public interface APIPersistence {
      * @param org   Organization the API is owned by
      * @throws APIPersistenceException
      */
-    void deleteAPIs(Organization org) throws APIPersistenceException;
+    void deleteAllAPIs(Organization org) throws APIPersistenceException;
 
     /**
      * Search APIs to be displayed on Publisher API listing
--- a/components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
@@ -145,6 +145,14 @@ public interface APIPersistence {
     void deleteAPI(Organization org, String apiId) throws APIPersistenceException;
 
     /**
+     * Delete APIs
+     *
+     * @param org   Organization the API is owned by
+     * @throws APIPersistenceException
+     */
+    void deleteAPIs(Organization org) throws APIPersistenceException;
+
+    /**
      * Search APIs to be displayed on Publisher API listing
      *
      * @param org         Organization the APIs are owned by
--- a/components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
@@ -145,7 +145,7 @@ public interface APIPersistence {
     void deleteAPI(Organization org, String apiId) throws APIPersistenceException;
 
     /**
-     * Delete APIs
+     * Delete all APIs
      *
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
```diff
@@ -144,6 +144,14 @@ void deleteAPIRevision(Organization org, String apiUUID, String revisionUUID, in
      */
     void deleteAPI(Organization org, String apiId) throws APIPersistenceException;
 
+    /**
+     * Delete APIs
+     *
+     * @param org   Organization the API is owned by
+     * @throws APIPersistenceException
+     */
+    void deleteAPIs(Organization org) throws APIPersistenceException;
```
chamilaadhi, line 153: if you add deleteAPIs , then you need to pass list of apis. if you want to delete all the apis  then i think name should be deleteAllAPIs()

components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/RegistryPersistenceImpl.java
```diff
@@ -875,6 +875,11 @@ public void deleteAPI(Organization org, String apiId) throws APIPersistenceExcep
         }
     }
 
+    @Override
+    public void deleteAPIs(Organization org) throws APIPersistenceException {
+
```
chamilaadhi, line 880: add a comment for keeping it empty

components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
HiranyaKavishani: No. we can just delete whole collection calling following way. We don't need the API list
MongoCollection<MongoDBPublisherAPI> collection = MongoDBConnectionUtil.getPublisherCollection(org.getName());
collection.drop();

components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
```diff
@@ -144,6 +144,14 @@ void deleteAPIRevision(Organization org, String apiUUID, String revisionUUID, in
      */
     void deleteAPI(Organization org, String apiId) throws APIPersistenceException;
 
+    /**
+     * Delete all APIs
+     *
+     * @param org   Organization the API is owned by
+     * @throws APIPersistenceException
+     */
+    void deleteAllAPIs(Organization org) throws APIPersistenceException;
```
malinthaprasan, line 153: This is not only about APIs right.. Even thought we consider about APIs only now, objective of this method is to delete anything that belongs to the organization? 
```suggestion
    void purge(Organization org) throws APIPersistenceException;
```

components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
malinthaprasan: This is renamed right @HiranyaKavishani

components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
malinthaprasan: Please update the discussion points we had @HiranyaKavishani

malinthaprasan (approved):

components/apimgt/org.wso2.carbon.apimgt.persistence/src/main/java/org/wso2/carbon/apimgt/persistence/APIPersistence.java
HiranyaKavishani: Yes. But as per the organization deletion use case, the deletion of the MongoDB collection is done under the APIPurge functionality.So a better name should be similar to `deleteAllAPI`

chamilaadhi (approved):

PR #10725, opened 2021-08-09
Add event publisher parameters

sdkottegoda:
Supports passing parameters in the for event hub publishers in the format of:
```
<EventPublisherConfiguration>
   <Properties>
      <Property name="connectionString">connection_string_value</Property>
   </Properties>
</EventPublisherConfiguration>
```
Supported deployment.toml configuration would be:
```
[apim.event_hub.publish.properties]
'connectionString'="sampleConnectionString"
'testParam2'="sampleTestParam2"
```

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
@@ -1749,6 +1749,7 @@ public class APIManagerConfiguration {
                 eventHubReceiverConfiguration.setJmsConnectionParameters(properties);
                 eventHubConfigurationDto.setEventHubReceiverConfiguration(eventHubReceiverConfiguration);
             }
+
             OMElement eventPublisherElement =
                     omElement.getFirstChildWithName(new QName(APIConstants.KeyManager.EVENT_PUBLISHER_CONFIGURATIONS));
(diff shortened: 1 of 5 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
```diff
@@ -1773,12 +1774,42 @@ private void setEventHubConfiguration(OMElement omElement) {
                 if (eventTypeElement != null) {
                     eventHubPublisherConfiguration.setType(eventTypeElement.getText().trim());
                 }
+                if (Boolean.parseBoolean(System.getenv("FEATURE_FLAG_REPLACE_EVENT_HUB"))) {
```
tharindu1st, line 1777: can't we set this by default without having a flag

tharindu1st (changes requested):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
sdkottegoda: This is a flag to indicate that this feature in progress. Unless someone specifically enables it, nobody sees the feature.

tharindu1st (approved):

malinthaprasan (approved):

PR #10745, opened 2021-08-19
fix REST PUT API /apis/{apiId}/swagger error 500

bmonteiro:
when using scopes => apim:api_create apim:api_manage
{
    "access_token": "c10a2418-aaa1-34e7-bbd0-8f7ecfe95c83",
    "refresh_token": "332a10a2-5263-3930-a1db-d4b939cb19d6",
    "scope": "apim:api_create apim:api_manage",
    "token_type": "Bearer",
    "expires_in": 3600
}

the APIM gives this error:

{"code":500,"message":"Internal server error","description":"Error while updating the swagger definition of the API: 83a20a71-5000-40fa-9f7c-aa168b3cdc88 - 900380:Insufficient permission to update the API::Updating the API is restricted as as it is PUBLISHED.","moreInfo":"","error":[]}

while using scopes =>  apim:api_create apim:api_publish
{
    "access_token": "dcbaa7d1-fdd7-37bb-87f5-6dee7e5a43c1",
    "refresh_token": "baa74507-56af-3c08-a55a-106a892437a4",
    "scope": "apim:api_create apim:api_publish",
    "token_type": "Bearer",
    "expires_in": 3017
}

then it is OK / 200 HTTP code

the docs here https://apim.docs.wso2.com/en/latest/reference/product-apis/publisher-apis/publisher-v2/publisher-v2/#operation/getAPISwagger should be updated specially here (scope apim:api_create alone also gives 500 error):

image

It should be  apim:api_create AND apim:api_publish Together

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.common/src/main/java/org/wso2/carbon/apimgt/rest/api/common/RestApiConstants.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.common/src/main/java/org/wso2/carbon/apimgt/rest/api/common/RestApiConstants.java
@@ -46,9 +46,11 @@ public final class RestApiConstants {
     public static final String USER_REST_API_SCOPES = "user_rest_api_scopes";
 
     public static final String API_IMPORT_EXPORT_SCOPE = "apim:api_import_export";
+    public static final String API_MANAGE_SCOPE = "apim:api_manage";
     public static final String CREATOR_SCOPE = "apim:api_create";
     public static final String ADMIN_SCOPE = "apim:admin";
     public static final String PUBLISHER_SCOPE = "apim:api_publish";
+
 
     public static final String DEFAULT_RESPONSE_CONTENT_TYPE = APPLICATION_JSON;
 
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/impl/ApisApiServiceImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/impl/ApisApiServiceImpl.java
@@ -831,6 +831,7 @@ public class ApisApiServiceImpl implements ApisApiService {
         for (String scope : tokenScopes) {
             if (RestApiConstants.PUBLISHER_SCOPE.equals(scope)
                     || RestApiConstants.API_IMPORT_EXPORT_SCOPE.equals(scope)
+                    || RestApiConstants.API_MANAGE_SCOPE.equals(scope)
                     || RestApiConstants.ADMIN_SCOPE.equals(scope)) {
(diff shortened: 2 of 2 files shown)


malinthaprasan:
Hi @bmonteiro,
First of all, thank you very much for the contribution and seems like this is a case we have missed during recent fixes.

While agreeing that we need to fix this case, there are a few cases we need to attend here. 

Currently having two scopes for updating swagger PUT /swagger which mean we allow either "apim:api_create" **OR** "apim:api_manage" to update the swagger. Here, the need for "apim:api_publish" is a bit of a special case because, we are checking for this scope in order to check the user has "Publisher/Product Manager" permission to update an already published API. If the API is not PUBLISHED, we can update the API with only having "apim:api_create".

However, we have recently introduced "apim:api_manage" with the intention of including "apim:api_create, apim:api_view, apim:api_delete, apim:api_publish .. " meaning with only having "apim:api_manage", you can do any operation of an API, which should include updating APIs while in PUBLISHED state as well. This case we have missed to update the validation at below code segment in [ApisApiServiceImpl.java](https://github.com/wso2/carbon-apimgt/blob/574947bbe81af4ad89d4a3a5cbfeb08c4c7807ba/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/impl/ApisApiServiceImpl.java#L825-L845). We need to include "apim:api_manage" scope here to the list of OR conditions.

```java
    private void validateAPIOperationsPerLC(String status) throws APIManagementException {
        ....
        for (String scope : tokenScopes) {
            if (RestApiConstants.PUBLISHER_SCOPE.equals(scope)
                    || RestApiConstants.API_IMPORT_EXPORT_SCOPE.equals(scope)
                    || RestApiConstants.API_MANAGE_SCOPE.equals(scope) // <<<< "apim:api_manage". Include the constant in RestApiConstants.java
                    || RestApiConstants.ADMIN_SCOPE.equals(scope)) {
                updatePermittedForPublishedDeprecated = true;
                break;
            }
        }
```
We can keep the current list of scopes in PUT /swagger and do the above mentioned fix.

bmonteiro:
@malinthaprasan makes sense, when do you expect that fix to be packaged and pushed to docker hub v4.x ?

bmonteiro:
@malinthaprasan I still believe the docs should be clear when to use any specific scope (PUBLISHED API use this scope OR New API use that scope, kind of explanation). It will avoid confusion when users just read the docs without checking the source code.

malinthaprasan (approved):

malinthaprasan:
> @malinthaprasan I still believe the docs should be clear when to use any specific scope (PUBLISHED API use this scope OR New API use that scope, kind of explanation). It will avoid confusion when users just read the docs without checking the source code.

Yes agreed. Tracked the case here https://github.com/wso2/docs-apim/issues/4527 

> @malinthaprasan makes sense, when do you expect that fix to be packaged and pushed to docker hub v4.x ?
Currently we do not push on demand fixes to public docker hub and a WSO2 subscription is required to get on-demand fixes via docker.wso2.com. The next GA release will be available for public via docker hub (APIM 4.1) in most probably 2022 Q1

PR #10782, opened 2021-09-08
introduce apimConfigService

tharindu1st:
Refactor Configuration retrieval


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/config/APIMConfigService.java
```diff
@@ -0,0 +1,144 @@
+package org.wso2.carbon.apimgt.impl.config;
```
malinthaprasan, line 1: License

malinthaprasan (approved):

PR #10788, opened 2021-09-10
Increase the charactor length limit of the API name attribute from 50 to 60

HasithaAthukorala:
This is to increase the character length limit from 50 to 60 ( Since the limit in Choreo is 60 )


chalitha1989:
@malinthaprasan Shall we review and merge this PR?

malinthaprasan (approved):

PR #10817, opened 2021-09-29
Fix the authorization header per tenant issue

Krishanx92:


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
```diff
@@ -10945,6 +10945,9 @@ components:
             Is Cross Tenant Subscriptions Enabled
           example: false
           default: false
+        authorizationHeader:
+          type: string
+          description: Authorization Header
```
vithu30, line 10950: shall we add en example as well? we need to update the same in docs as well?

vithu30 (approved):

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
Krishanx92: Fixed.

PR #10829, opened 2021-10-01
Get organization given a consumer key

fazlan-nazeem:
This is needed for the new ChoreTokenIssuer which will be used in Choreo APIM


fazlan-nazeem:
Found a method we could reuse. Hence closing this PR

PR #10843, opened 2021-10-07
move identity provider code into impl

tharindu1st:


components/apimgt/org.wso2.carbon.apimgt.rest.api.admin.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/v1/impl/KeyManagersApiServiceImpl.java
```diff
@@ -117,13 +84,18 @@ public Response keyManagersKeyManagerIdDelete(String keyManagerId, MessageContex
...
-        APIUtil.logAuditMessage(APIConstants.AuditLogConstants.KEY_MANAGER,
-                new Gson().toJson(keyManagerConfigurationDTO), APIConstants.AuditLogConstants.DELETED,
-                RestApiCommonUtil.getLoggedInUsername());
-        return Response.ok().build();
+            APIUtil.logAuditMessage(APIConstants.AuditLogConstants.KEY_MANAGER,
+                    new Gson().toJson(keyManagerConfigurationDTO), APIConstants.AuditLogConstants.DELETED,
+                    RestApiCommonUtil.getLoggedInUsername());
+            return Response.ok().build();
+        } else {
+            RestApiUtil.handleResourceNotFoundError(RestApiConstants.RESOURCE_KEY_MANAGER, keyManagerId, log);
```
wasuradananjith, line 96: Wouldn't it be better if we throw an APIManagementException with an exception code here so that the global thrower will throw it?

tharindu1st: Since we in RestAPI layer we don't need to throw API Management Exception from here.

PR #10852, opened 2021-10-11
fixed network connection issue in schema loading

hasuniea:
This PR fixes the wso2/product-apim#11542


dushaniw (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/AsyncApiParser.java
```diff
@@ -45,6 +49,176 @@
 
 public class AsyncApiParser extends APIDefinition {
 
+    String metaSchema = "{\n" +
```
tharindu1st, line 52: can we put this to a file in the resource file to jar without hardcoding inside code?

PR #10930, opened 2021-11-29
Fix https://github.com/wso2/product-apim/issues/12004

shilmyhasan:
Fix https://github.com/wso2/product-apim/issues/12004

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
@@ -378,7 +378,8 @@ public class AMDefaultKeyManagerImpl extends AbstractKeyManager {
         ClientInfo request = createClientInfo(oAuthApplicationInfo, oauthClientName, true);
         ClientInfo createdClient;
         try {
-            createdClient = dcrClient.updateApplication(oAuthApplicationInfo.getClientId(), request);
+            createdClient = dcrClient.updateApplication(Base64.getUrlEncoder().encodeToString(
+                    oAuthApplicationInfo.getClientId().getBytes(StandardCharsets.UTF_8)), request);
             return buildDTOFromClientInfo(createdClient, new OAuthApplicationInfo());
         } catch (KeyManagerClientException e) {
             handleException("Error occurred while updating OAuth Client : ", e);
@@ -395,7 +396,8 @@ public class AMDefaultKeyManagerImpl extends AbstractKeyManager {
 
         ClientInfo updatedClient;
         try {
-            updatedClient = dcrClient.updateApplicationOwner(owner, oAuthApplicationInfo.getClientId());
+            updatedClient = dcrClient.updateApplicationOwner(owner, Base64.getUrlEncoder().encodeToString(
+                    oAuthApplicationInfo.getClientId().getBytes(StandardCharsets.UTF_8)));
             return buildDTOFromClientInfo(updatedClient, new OAuthApplicationInfo());
         } catch (KeyManagerClientException e) {
             handleException("Error occurred while updating OAuth Client : ", e);
@@ -411,7 +413,8 @@ public class AMDefaultKeyManagerImpl extends AbstractKeyManager {
         }
 
         try {
-            dcrClient.deleteApplication(consumerKey);
+            dcrClient.deleteApplication(Base64.getUrlEncoder().encodeToString(
+                    consumerKey.getBytes(StandardCharsets.UTF_8)));
         } catch (KeyManagerClientException e) {
             handleException("Cannot remove service provider for the given consumer key : " + consumerKey, e);
         }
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
```diff
@@ -424,7 +426,13 @@ public OAuthApplicationInfo retrieveApplication(String consumerKey) throws APIMa
         }
 
         try {
-            ClientInfo clientInfo = dcrClient.getApplication(consumerKey);
+            encoded = false;
+            String isConsumerKeyEncoded = System.getProperty(ENCODE_CONSUMER_KEY, "false");
+            if (isConsumerKeyEncoded.equalsIgnoreCase("true")) {
```
mushthaq33, line 431: Shall we swap this to avoid NPE?

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
```diff
@@ -565,7 +573,13 @@ public OAuthApplicationInfo mapOAuthApplication(OAuthAppRequest appInfoRequest)
         //check whether given consumer key and secret match or not. If it does not match throw an exception.
         ClientInfo clientInfo;
         try {
-            clientInfo = dcrClient.getApplication(consumerKey);
+            encoded = false;
+            String isConsumerKeyEncoded = System.getProperty(ENCODE_CONSUMER_KEY, "false");
+            if (isConsumerKeyEncoded.equalsIgnoreCase("true")) {
```
mushthaq33, line 578: Same here

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
shilmyhasan: The default value is set to be "false" in the above method to get the system property, hence it can not become null

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
shilmyhasan: The default value is set to be "false" in the above method to get the system property, hence it can not become null

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
mushthaq33: But better to have it the other way around as a best practice. Shall we change it?

bhathiya: +1, or we have to first convert it to boolean and check with `== true`.

malinthaprasan (changes requested):
Do we need a system property for this? `@RequestLine("GET /{clientId}?encoded={encoded}")` any parameter we send in the URL should always be URL encoded..

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
shilmyhasan: Addressed in b1a9d7e66f2d8ee05ee815b7e2d632dc91eeb734

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
shilmyhasan: Addressed in b1a9d7e66f2d8ee05ee815b7e2d632dc91eeb734

shilmyhasan:
> Do we need a system property for this? `@RequestLine("GET /{clientId}?encoded={encoded}")` any parameter we send in the URL should always be URL encoded..

The system property is used to decide whether the consumer key should be encoded or not based on the customers requirement. Based on the above decision only we set the query param and decides whether its encoded to be decoded at key managers end.

bhathiya:
@shilmyhasan try to address coverage issue comments by codecov.

tharindu1st (approved):

shilmyhasan:
> @shilmyhasan try to address coverage issue comments by codecov.

As per the discussion with @mushthaq33  , It seems that the unit test can not be written as we do not support creating application with consumers keys which involves special chars through our rest apis

PR #10961, opened 2021-12-20
Fix artifact retrieval issue when gateway worker node is started first

ashera96:
## Purpose

- Resolves https://github.com/wso2/product-apim/issues/12126

This PR fixes the issue where the API invocations result in a 404 response when the `Gateway Profile` is started first.

## RCA

The artifact retrieval step fails with an artifact synchronization exception when the gateway profile is started first. This is encountered with the inability to retrieve the existing artifacts since the control plane is not up yet. In order to fix the aforementioned issue, a delaying factor is introduced to the thread that is trying to retrieve the artifacts


ashera96:
Unable to verify this fix due to the distrubuted setup related issue mentioned in [1]. Once that is fixed, we can consider merging this PR.

1. https://github.com/wso2/product-apim/issues/12155

tharindu1st (approved):

PR #10986, opened 2022-01-27
Bump Rest API version

HiranyaKavishani:
This PR will update the REST API version to v3


malinthaprasan:
Why do we upgrade REST API version from a major version?
Do we have any breaking changes?

HiranyaKavishani:
Since this is not for a major release we will make the rest api upgrade version to v2.1.

PR #10989, opened 2022-01-28
Bump Rest API versions

HiranyaKavishani:
This PR will bump publisher, devportal, admin, gateway rest API version to v2.1 version and the rest APIs are still supported with relevant major version (v2)


HiranyaKavishani:
Close the issue as the change is already done

PR #11012, opened 2022-02-07
Refine Solace deployment flow to support Async API update.

Sarangan0219:
### Purpose

- Fixes https://github.com/wso2/product-apim/issues/12210

### User Stories

When deploying a new revision,

-  The older snapshot of API and API product in Solace is un deployed 
-  The newer revision will be deployed


Chamindu36 (approved):

Sarangan0219:
We Don't need this implementation as we decided to proceed with existing revision flow for the Solace APIs.

PR #11018, opened 2022-02-08
Fix pagination in /subscription GET when an applicationId query param is there

msm1992:
**Purpose**

Fixes https://github.com/wso2/product-apim/issues/12097

**Approach**

Total subscriptions count is fetched from a call to getSubscriptionCount
Pagination params are set to subscriptionListDTO


RakhithaRR (approved):

tharindu1st:
@msm1992 do we have a integration test to cover this scenario

chamilaadhi:
@msm1992 can you check and merge this plz

PR #11039, opened 2022-02-11
Bump Rest API Versions

HiranyaKavishani:
Rest API versions are changed as bellows

* publisher :  v2--> v3 (major version: v3 , latest version : v3)
* admin :  v2 --> v3 (major version: v3 , latest version : v3)
* devportal : v2 --> (major version: v2 , latest version : v2.1)
* gateway : v2 --> (major version: v2 , latest version : v2.1)
* serviceCatelog : v0 --> v1 
* devops : v1 --> v0 (previously-v1)


malinthaprasan:
> Rest API versions are changed as bellows
> 
> * publisher :  v2--> v3
> * admin :  v2 --> v3
> * devportal : v2
> * gateway : v2 -->  v2.1
> * serviceCatelog : v0 --> v1
> * devops : v1 --> v0 (previously-v1)

For gateway, shall we avoid this v2.1 change in the URL? As @HiranyaKavishani mentioned, Gateway API doesn't have breaking changes but if we make this change in the URL, it itself will become a breaking change.
I think, we can keep this as v2 for now and later (probably after alpha, handle both v2, v2.1 internally)

Krishanx92 (approved):

PR #11041, opened 2022-02-11
Remove deprecated mediation policy related REST APIs from Publisher

ashera96:


mushthaq33:
There are conflicts. Shall we resolve them?

PR #11098, opened 2022-03-01
Remove multiple allowed field from the operation policy specification

1akshitha:


1akshitha:
will be added via https://github.com/wso2/carbon-apimgt/pull/11105

PR #11127, opened 2022-03-09
Adding gatewayVendor property for migration property update validation

HiranyaKavishani:
$Subject


HiranyaKavishani:
We will remove the gateway vendor feature from migration

PR #11133, opened 2022-03-09
Fix the issue with default API context

RakhithaRR:
Fixes https://github.com/wso2/product-apim/issues/12568


Krishanx92 (approved):

gspadotto:
Shouldn't this line be changed as well?
https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/impl/SubscriptionDataStoreImpl.java#L583

following the same logic as: 
https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/impl/SubscriptionDataStoreImpl.java#L709-L711

PR #11166, opened 2022-03-16
Display subscription operation as the resource path in analytics graphs

msm1992:
Screenshot 2022-03-16 at 17 30 40

**Approach**
* A new field ctx was introduced to the  inboundMessageContext to hold ChannelHandlerContext
* API_ELECTED_RESOURCE is set as a property to ctx

Related Issue : https://github.com/wso2/product-apim/issues/12197

Note : The wildcard entries in the screenshot are from invocations before applying this fix.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/inbound/websocket/request/GraphQLRequestProcessor.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/inbound/websocket/request/GraphQLRequestProcessor.java
@@ -34,6 +34,7 @@ import org.wso2.carbon.apimgt.gateway.handlers.graphQL.GraphQLConstants;
 import org.wso2.carbon.apimgt.gateway.handlers.graphQL.analyzer.SubscriptionAnalyzer;
 import org.wso2.carbon.apimgt.gateway.handlers.graphQL.utils.GraphQLProcessorUtil;
 import org.wso2.carbon.apimgt.gateway.handlers.streaming.websocket.WebSocketApiConstants;
+import org.wso2.carbon.apimgt.gateway.handlers.streaming.websocket.WebSocketUtils;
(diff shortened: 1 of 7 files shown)


components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/inbound/websocket/request/GraphQLRequestProcessor.java
```diff
@@ -87,6 +88,8 @@ public InboundProcessorResponseDTO handleRequest(int msgSize, String msgText,
                             // subscription operation name
                             String subscriptionOperation = GraphQLProcessorUtil.getOperationList(operation,
                                     inboundMessageContext.getGraphQLSchemaDTO().getTypeDefinitionRegistry());
+                            WebSocketUtils.setApiPropertyToChannel(inboundMessageContext.getCtx(),
```
dushaniw, line 91: Dont we need to add the same fix to GraphQLResponseProcessor in handleResponse()? If someone use the same connection for two different subscriptions what would be the behavior?

dushaniw:
@msm1992 in the screenshot the entry with 4 hits for the same API, were these added before the fix ?

msm1992:
Yes, those 4 entries were there before this fix as well.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/inbound/websocket/request/GraphQLRequestProcessor.java
msm1992: fixed with https://github.com/wso2/carbon-apimgt/pull/11166/commits/fff020f1e84428a67f6601c736bbcbc1b2ca6cf9. Tested with multiple subscriptions in the same connections and shown below is show it looks in portal, correct resource path is mentioned in the table.
Screenshot 2022-03-18 at 06 35 00

dushaniw (approved):

Krishanx92 (approved):

PR #11194, opened 2022-03-22
Fix error log during server startup with no APIs

tharikaGitHub:
This PR fixes https://github.com/wso2/product-apim/issues/12730


tharikaGitHub:
Closing as per the comment in https://github.com/wso2/product-apim/issues/12730#issuecomment-1075176700

PR #11273, opened 2022-05-23
changes for migration client revamping 

hasuniea:


chamilaadhi:
closing this since there is a plan to re-implement the functionality in a different way

dushaniw:
Instead of using a system property, it was discussed to do this in a better way. This will be tracked with https://github.com/wso2/api-manager/issues/748 for 4.2.0.

PR #11337, opened 2022-07-21
Enable stack trace for >=500

ThusharaSampath:
Enable stack trace for 500 for further investigation on https://github.com/wso2-enterprise/choreo/issues/14234

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
@@ -544,6 +544,9 @@ public enum ExceptionCodes implements ErrorHandler {
      * @param errorDescription The error description.
      */
     ExceptionCodes(long errorCode, String msg, int httpErrorCode, String errorDescription) {
+        if(httpErrorCode>=500){
+                this.stackTrace=true;
+        }
         this.errorCode = errorCode;
         this.errorMessage = msg;
         this.httpStatusCode = httpErrorCode;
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
```diff
@@ -544,6 +544,9 @@ public enum ExceptionCodes implements ErrorHandler {
      * @param errorDescription The error description.
      */
     ExceptionCodes(long errorCode, String msg, int httpErrorCode, String errorDescription) {
+        if(httpErrorCode>=500){
+                this.stackTrace=true;
+        }
```
pubudu91, line 549: Shouldn't the fix be to use the other constructor which accepts the `stackTrace` flag for 5xx errors? 🤔

malinthaprasan: Ideally we need the stack traces for Server Errors, which are 5xx errors.
Other method can also be used, but that is causing changes to all the exception code initialization which I think is unnecessary as we can do the same thing in a simple logic as above.

malinthaprasan, line 549: Lets format the code @ThusharaSampath

malinthaprasan:
@tharindu1st pls check

malinthaprasan (approved):

components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
```diff
@@ -544,6 +544,9 @@ public enum ExceptionCodes implements ErrorHandler {
      * @param errorDescription The error description.
      */
     ExceptionCodes(long errorCode, String msg, int httpErrorCode, String errorDescription) {
+        if (httpErrorCode >= 500) {
+            this.stackTrace = true;
```
tharindu1st, line 548: can't we make stack trace true for specific error codes? without making globally.

malinthaprasan: Ideally, we should be showing the stacktraces for all 500 errors. So even if we are putting them to specific error codes, we'll have to be put them into all the 500 type error codes. From this we are avoiding that and make the stack trace loggable by default for all 500 type of error codes. If the stack trace of any error code should not be logged, we can override this by using the other constructor where we can pass the stacktrace boolean.

PR #11372, opened 2022-08-09
Skip the API that is not stored in APIM DB in devportal API search

IsuruBoyagane15:
Resolves https://github.com/wso2-enterprise/choreo/issues/14546

If an API present in the mongoDB search result is not persisted in the APIM DB, APIM DB retrival call will fail for that API. And will give a 500 status code to devportal API search call (`devportal/v2/apis`). This fix will skip such an APIs so that remaining search results will be sent with a 200 status code.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
@@ -5637,8 +5637,14 @@ public class APIConsumerImpl extends AbstractAPIManager implements APIConsumer {
                 List<Object> apiList = new ArrayList<>();
                 for (DevPortalAPIInfo devPortalAPIInfo : list) {
                     API mappedAPI = APIMapper.INSTANCE.toApi(devPortalAPIInfo);
-                    mappedAPI.setRating(APIUtil.getAverageRating(mappedAPI.getUuid()));
-                    apiList.add(mappedAPI);
+                    try {
+                        mappedAPI.setRating(APIUtil.getAverageRating(mappedAPI.getUuid()));
+                        apiList.add(mappedAPI);
+                    } catch (APIManagementException e) {
+                        if (log.isDebugEnabled()) {
+                            log.debug("Getting Average rating failed for API: " + mappedAPI.getUuid());
+                        }
+                    }
                 }
                 apiSet.addAll(apiList);
                 result.put("apis", apiSet);


ShalkiWenushika (approved):

ShalkiWenushika:
Let's add this PR to the 9.22.x branch since we will be using that branch in choreo after the choreo apim migration.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
```diff
@@ -5637,8 +5637,14 @@ public Map<String, Object> searchPaginatedAPIs(String searchQuery, String organi
...
-                    mappedAPI.setRating(APIUtil.getAverageRating(mappedAPI.getUuid()));
-                    apiList.add(mappedAPI);
+                    try {
+                        mappedAPI.setRating(APIUtil.getAverageRating(mappedAPI.getUuid()));
+                        apiList.add(mappedAPI);
+                    } catch (APIManagementException e) {
+                        if (log.isDebugEnabled()) {
+                            log.debug("Getting Average rating failed for API: " + mappedAPI.getUuid());
+                        }
+                    }
```
pubudu91, line 5647: The `searchPaginatedAPIs()` method declares that it throws `APIManagementException`. So catching it here and not rethrowing means the flow is now altered. i.e., a previous flow where `searchPaginatedAPIs()` terminated with an exception will now return as if nothing happened. Is that behavioural change intended?

IsuruBoyagane15: Yes. 
The intention is to do `apiList.add(mappedAPI);` to only `mappedAPI`s (in the `for` loop) that did not get `APIManagementException` when calling `APIUtil.getAverageRating()`.

Earlier it threw `APIManagementException` and considered as an error. Here it will skip the erroneous iteration of the `for ` loop and add the rest to the list.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIConsumerImpl.java
```diff
@@ -5637,8 +5637,14 @@ public Map<String, Object> searchPaginatedAPIs(String searchQuery, String organi
                 List<Object> apiList = new ArrayList<>();
                 for (DevPortalAPIInfo devPortalAPIInfo : list) {
                     API mappedAPI = APIMapper.INSTANCE.toApi(devPortalAPIInfo);
-                    mappedAPI.setRating(APIUtil.getAverageRating(mappedAPI.getUuid()));
-                    apiList.add(mappedAPI);
+                    try {
+                        mappedAPI.setRating(APIUtil.getAverageRating(mappedAPI.getUuid()));
+                        apiList.add(mappedAPI);
+                    } catch (APIManagementException e) {
```
malinthaprasan, line 5643: I think we should not just skip this error and log as debug. This need to be logged as an error because this is an actual error. WDYT?

pubudu91: From what I understood from https://github.com/wso2/carbon-apimgt/pull/11372#discussion_r941069575, in this particular case this is not a hindrance to the regular flow i.e., whether the exception gets thrown or not is irrelevant to continuing on with the flow. So maybe more of a warning kinda thing? 🤔

IsuruBoyagane15: Above suggestions is also added to the PR to `9.22.x` branch. (https://github.com/wso2/carbon-apimgt/pull/11375)

PR #11428, opened 2022-09-01
Add linter custom rules API for publisher

imsuneth:
https://github.com/wso2/api-manager/issues/657

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
@@ -6817,6 +6817,45 @@ paths:
       operationId: getSettings
 
   ######################################################
+  # The linter custom rules resource APIs
+  ######################################################
+  /linter-custom-rules:
+    get:
+      tags:
+        - Linter Custom Rules
+      summary: Get linter custom rules.
+      description: |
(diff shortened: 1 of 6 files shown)


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
```diff
@@ -6817,6 +6817,45 @@ paths:
...
+          headers:
+            Content-Type:
+              description: |
+                The content type of the body.
+              schema:
+                type: string
+          content:
+            application/json:
+              schema:
+                type: string
```
rmsamitha, line 6844: Can't we introduce a schema for this instead of a string? Can you provide an example json payload here so we can decide the feasibility.

rmsamitha (approved):

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml
imsuneth: The response payload is a set of rules defined according to spectral custom rules definition ([sh api-manager.sh -Dmigrate -Dcomponent=identity](https://meta.stoplight.io/docs/spectral/e5b9616d6d50c-custom-rulesets)). Therefore, the schema will be dynamic.

PR #11436, opened 2022-09-05
Add a filter to filter out blocked unsubscribed users for deleteAPI

SavinduDimal:
Fixes wso2/api-manager/issues/238
Removed checking for subscription status when removing a subscription


chamilaadhi (commented):
Let's try to fix the API deletion logic instead of removing this from the unsubscribe section

chamilaadhi (approved):

PR #11439, opened 2022-09-06
Use an JSON array to throw the error payload in GraphQL

PasanT9:
Use an JSON array to throw the error payload in GraphQL instead of an single JSON object.


dushaniw (approved):

PasanT9:
**The API Manager Build with Test / build-product (1, group1) (pull_request)** build fails with the following test case failures.

- GraphqlSubscriptionTestCase.testGraphQLAPIInvocationForComplexity
- GraphqlSubscriptionTestCase.testGraphQLAPIInvocationWithInvalidPayload

Fixes for these failures will be covered by https://github.com/wso2/product-apim/pull/12964

PR #11462, opened 2022-09-13
[APK] API service impl refactor test fail fix

chameerar:
Fixes the integration test failures regarding API revisions deploying with invalid environment test case.


RakhithaRR (approved):

rmsamitha:
Please link the related git issue. If there is no git issue you should create a git issue and link. Need to do this as a practice.

PR #11626, opened 2022-11-10
Fix error adding WSO2-IS as Key Manager from Admin portal

YasasRangika:
### Purpose
To fix the NPE error when adding WSO2-IS as Key Manager from the Admin portal.

### Goal
Fixes: https://github.com/wso2/api-manager/issues/885

### Approach
- Validation added to validate before searching a value for revokeOneTimeTokenEndpoint since this is only for resident KM.
- UI validation added for other required fields: userInfoEndpoint and scopeManagementEndpoint

Related PR: https://github.com/wso2/apim-apps/pull/379

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
@@ -701,14 +701,6 @@ public class AMDefaultKeyManagerImpl extends AbstractKeyManager {
                     .concat(getTenantAwareContext().trim()).concat
                             (APIConstants.KeyManager.KEY_MANAGER_OPERATIONS_USERINFO_ENDPOINT);
         }
-        String revokeOneTimeTokenEndpoint;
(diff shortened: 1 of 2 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/AMDefaultKeyManagerImpl.java
```diff
@@ -756,15 +748,28 @@ public void loadConfiguration(KeyManagerConfiguration configuration) throws APIM
...
-                .client(new ApacheFeignHttpClient(APIUtil.getHttpClient(revokeOneTimeTokenEndpoint)))
-                .encoder(new GsonEncoder())
-                .decoder(new GsonDecoder())
-                .logger(new Slf4jLogger())
-                .requestInterceptor(new BasicAuthRequestInterceptor(username, password))
-                .requestInterceptor(new TenantHeaderInterceptor(tenantDomain))
-                .errorDecoder(new KMClientErrorDecoder())
-                .target(RevokeClient.class, revokeOneTimeTokenEndpoint);
+
+        if (!configuration.getType().equals(APIConstants.KeyManager.WSO2_IS_KEY_MANAGER_TYPE)) {
```
tharikaGitHub, line 752: Shall we interchange the sides of the comparing parameters in equals? This would prevent a possible null pointer exception.

tharikaGitHub (approved):

PR #11658, opened 2022-11-21
Handling Null Point Exception | JWT Conditional Policy 

Tharanidk:
When we set up the JWT conditional rate limiting policy under advanced throttling policies there was a NullPointException popping up without throttling out at the expected limit at the time a non-authenticated resource is invoked. That issue is fixed with this PR. Now for non-authenticated resources, it throttles out from the default limit as the security limit is not applicable.


components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/throttling/ThrottleConditionEvaluator.java
```diff
@@ -221,6 +221,10 @@ private boolean isJWTClaimPresent(AuthenticationContext authenticationContext, C
 
     private boolean isJWTClaimPresent(AuthenticationContext authenticationContext, ConditionDto.JWTClaimConditions
             condition) {
+                
+                if (authenticationContext.getCallerToken() == null) {
+                    return false;
+                }
```
senthuran16, line 227: Let's indent this block appropriately

senthuran16 (approved):

PR #11744, opened 2023-01-03
Enabling multiple reporters to publish analytics parallelly.

SarithRavI:
Current implementation supports only one reporter to publish data (which means only one analytics dashboard can be used at a time). Currently custom reporters are allowed and the default reporter sends data to Choreo Insights. 
This PR enables to support multiple reporters. Multiple reporters can be included as key-value pairs under enforcer analytics config properties map. keys should be in the following regex form: 
```shell
"^publisher//.reporter[1-9][0-9]*//.class$"   
```
example:
```shell
"publisher.reporter1.class" = "moesif.analytics.reporter.MoesifReporter"
"publisher.reporter2.class" = "org.wso2.am.analytics.publisher.reporter.cloud.DefaultAnalyticsMetricReporter"
```

Existing ways of publishing analytics are still applicable. For ease, lets refer the way above example includes reporters as  'multi reporter listing'. We set a hierarchy for all of these ways as follows (top method denotes the method highest in hierarchy):

- Method 1: Defining a reporter using existing "publisher.reporter.class"
- Method 2: Setting "type" key to "elk", which will send analytics to ELK.
- Method 3: multi reporter listing

Whenever more than one of the above methods are used, the outcome will be the outcome of the method which is higher in the hierarchy. For example if both Method 1 and Method 2 methods were used the analytics will be sent only to the destination defined by the reporter given in the value of "publisher.reporter.class". In case of none of the methods are used the data will be sent to Choreo Insight using the default reporter.

Note worthy features:
- No numerical limit for multi reporter listing as long as latency is not an issue.
- In a failure of one reporter, the implementation will switch to the next reporter.
- Error logs and Exception messages denote from which reporter the particular error/exception was raised (Note if the error/exception occured in CounterMetric level then it will be the associated CounterMetric implementation of the reporter). 

Testing:
Successfully tested against default reporter and a custom reporter.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.common.analytics/pom.xml
+++ b/components/apimgt/org.wso2.carbon.apimgt.common.analytics/pom.xml
@@ -43,6 +43,7 @@
         <dependency>
             <groupId>org.wso2.am.analytics.publisher</groupId>
             <artifactId>org.wso2.am.analytics.publisher.client</artifactId>
+            <version>${org.wso2.am.analytics.publisher.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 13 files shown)


SarithRavI:
This PR is faulty.
This PR has two separate features together. So I'm closing this.

PR #11780, opened 2023-01-18
Active socket connections should not be able to send new messages, after token is revoked

Tharsanan1:
Resolves: https://github.com/wso2/api-manager/issues/1053


dushaniw (approved):

dushaniw (approved):

dushaniw:
Failing test cases in the workflow build `APIScopeTestCase` and `APIM4765ResourceOrderInSwagger` are not related to the fix. Hence merging.

PR #11784, opened 2023-01-19
Support export/import API policies with Synapse and CC gateway types

wasuradananjith:
## Purpose
Earlier there was an error when export/importing an API Policy that supports both Synapse and Choreo Connect gateways. This will fix that issue.
Fixes: https://github.com/wso2/api-manager/issues/1274

## Goals
Support export/import API policies with Synapse and CC gateway types

## Approach
Check the extension type of the definition file and if it is ".gotmpl", then set the gateway type accordingly.


lakshithagunasekara (approved):

wasuradananjith:
The failed tests in the checks are APIM4765ResourceOrderInSwagger.swaggerResourceOrderTest, APIMIntegrationBaseTest.undeployAndDeleteAPIRevisionsUsingRest and APIScopeTestCase.testSetScopeToResourceTestCase which are related to APIs, not to API Policies. Hence, these failures are not related to this fix.

ashera96 (approved):

PR #11790, opened 2023-01-22
Fixed nimbus class not found issue.

prasa7:


Arshardh:
There is no such issue reproducing atm. Hence closing.

PR #11791, opened 2023-01-23
Check revoked token cache before gateway token cache

Tharsanan1:
Related to: https://github.com/wso2/api-manager/issues/1053


Tharsanan1:
The changes in this PR no longer valid

PR #11837, opened 2023-02-06
Set autocommit to true, to get visibility over other transactions' changes when defaultAutoCommit=false

Tharsanan1:
Resolves: https://github.com/wso2/api-manager/issues/1270

We have a configuration property called defaultAutoCommit. When this property is set to zero and if the database transaction isolation level is REPEATABLE-READ or higher, lookup queries do not have visibility over changes that happened after a connection creation. In order to support both true and false for this defaultAutoCommit configuration we should always set auto-commit to true before doing any database lookups.


Tharsanan1:
This fix is needed to resolve the race condition when we use MYSQL database with 'REPEATABLE-READ' isolation level when we have config 'auto-commit = true'. But for now we are recommending the users to only use 'auto-commit = false' in the product configuration. So closing this PR. But we have to keep in mind that if we are to support 'auto-commit = true' configs we have to do similar code changes in all the places where we do database queries.

PR #11869, opened 2023-02-17
Remove No API Artifacts error log on server startup

DinithHerath:
## Purpose

- `[2023-02-09 09:27:11,464] ERROR - GlobalThrowableMapper No API Artifacts` is getting printed in the server startup logs.
- This happens only when there is no APIs deployed in the APIM. Ideally this happens in fresh APIM packs.
- Fixed https://github.com/wso2/api-manager/issues/1406

## Approach
- Made the exception throwing cases into `INFO` logs since those scenarios are not falling in to exceptions.
- Refactored code to remove unreachable areas.


dushaniw (approved):

malinthaprasan:
please check the behaviour of this change with choreo connect when running with APIM
AFAIR, we had to revert a similar change in the past

DinithHerath:
@malinthaprasan We have tested the functionality with choreo([Choreo Connect 1.2.0-Alpha](https://github.com/wso2/product-microgateway/releases/tag/v1.2.0-alpha)) and apim(local setup of beta pack with modified jar in both full and control plane profile) and the functionalities are all working with expected behaviours.

malinthaprasan (approved):

PR #11906, opened 2023-03-01
Bug fix: Populate x-throttle-limit extension with default  throttle limit for operations based on x-throttling-tier

VirajSalaka:
populate x-throttle-limit extension (using -1 for request count which indicates the unlimited policy) for all the operations in the swagger, if the operations does not have the extension.

P.S.
Correct approach would be process the x-throttling-tier to generate the x-throttling-limit when the operation has that property.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS2Parser.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS2Parser.java
@@ -872,6 +872,12 @@ public class OAS2Parser extends APIDefinition {
             for (Map.Entry<HttpMethod, Operation> entry : operationMap.entrySet()) {
                 Operation operation = entry.getValue();
                 operation.setVendorExtension(APIConstants.X_WSO2_APP_SECURITY, appSecurityExtension);
+                if (!operation.getVendorExtensions().containsKey(APIConstants.SWAGGER_X_THROTTLING_LIMIT)) {
+                    // set default throttling limit if not defined for operation in openapi definition. This is expected
+                    // from the choreo console.
+                    operation.setVendorExtension(APIConstants.SWAGGER_X_THROTTLING_LIMIT,
+                            APIUtil.getDefaultThrottleLimit());
+                }
             }
         }
         swagger.setVendorExtension(APIConstants.X_WSO2_APP_SECURITY, appSecurityExtension);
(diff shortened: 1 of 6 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/definitions/OAS2Parser.java
```diff
@@ -872,6 +872,12 @@ public String getOASDefinitionForPublisher(API api, String oasDefinition) throws
             for (Map.Entry<HttpMethod, Operation> entry : operationMap.entrySet()) {
                 Operation operation = entry.getValue();
                 operation.setVendorExtension(APIConstants.X_WSO2_APP_SECURITY, appSecurityExtension);
+                if (!operation.getVendorExtensions().containsKey(APIConstants.SWAGGER_X_THROTTLING_LIMIT)) {
+                    // set default throttling limit if not defined for operation in openapi definition. This is expected
+                    // from the choreo console.
+                    operation.setVendorExtension(APIConstants.SWAGGER_X_THROTTLING_LIMIT,
```
slahirucd7, line 878: Just to clarify:
If there's an existing operation with `50KPerMin` don't we need to convert that to the `x-throttling-limit` value (IMO we can use the existing method to get that value. I.E method with the switch case)?
+
Can we update the comment in line 866 ?

VirajSalaka: good point, yeah we need to fix it in that way. If x-throttling-tier is available then we need to populate this limit accordingly.

components/apimgt/org.wso2.carbon.apimgt.impl/src/test/java/org/wso2/carbon/apimgt/impl/definitions/OAS2ParserTest.java
```diff
@@ -147,4 +151,42 @@ public void testRemoveResponsesObjectFromOpenAPI20Spec() throws Exception {
         Assert.assertEquals(oas2Parser.removeResponsesObject(swagger,swaggerWithoutResponsesObject),
                 oas2Parser.removeResponsesObject(swagger,swaggerWithResponsesObject));
     }
+
+    @Test
```
slahirucd7, line 155: If possible let's add a description.

slahirucd7 (approved):

components/apimgt/org.wso2.carbon.apimgt.impl/src/test/java/org/wso2/carbon/apimgt/impl/definitions/OAS2ParserTest.java
VirajSalaka: Since we use the same method signature as in the acutal method I guess description is not required. (Same approach followed throughout the other tests as well)

PR #11917, opened 2023-03-08
Return only active non-external KMs via internal service

malinthaprasan:
Related: https://github.com/wso2-enterprise/choreo/issues/19243

cc: @VirajSalaka


components/apimgt/org.wso2.carbon.apimgt.internal.service/src/main/java/org/wso2/carbon/apimgt/internal/service/impl/KeymanagersApiServiceImpl.java
```diff
@@ -49,7 +49,7 @@ public Response keymanagersGet(String xWSO2Tenant, MessageContext messageContext
             if (StringUtils.isNotEmpty(organizationFromQueryParam) &&
                     organizationFromQueryParam.equalsIgnoreCase(APIConstants.ORG_ALL_QUERY_PARAM)) {
                 Map<String, List<KeyManagerConfigurationDTO>> keyManagerConfigurationsPerOrg =
-                        apiAdmin.getAllKeyManagerConfigurations();
+                        apiAdmin.getAllActiveKeyManagerConfigurations();
```
VirajSalaka, line 52: Shouldn't we fix the else path well? For PDPs

malinthaprasan: Yes.. and seems like org specific KMs are not sent at all when the org parameter is sent.. will fix that.. thanks for the pointer

malinthaprasan: Fixed

VirajSalaka (approved):

PR #11924, opened 2023-03-13
Fix Slo Issue

chashikajw:
### Description
This will fix the saml SLO issue in IS and WS2 API Manager setup

In this pr the isUserAuthorized check will be removed since it is not a valid check and according to the check all the users should have the Application role.

Note: isUserAuthorized check was added recently by https://github.com/wso2/carbon-apimgt

### Fixes
- https://github.com/wso2/api-manager/issues/824


chashikajw:
Closing this for the moment since this needs to analyse further and open a new pr for this for the master branch if this needs.

PR #11941, opened 2023-03-23
Fix unable to add API specific policy using YAML format via Publisher REST API

thisaltennakoon:
Fix https://github.com/wso2/api-manager/issues/1635


thisaltennakoon:
Opened a new PR: https://github.com/wso2/carbon-apimgt/pull/12237

PR #11976, opened 2023-04-20
Change opaque token to JWT for portals

chamilaadhi:
In this fix
- Change the default token type to JWT
- Remove audience validation for product rest API


chamilaadhi:
This fix is sent in a different pr for token persistence feature

PR #12075, opened 2023-07-05
Username is null in tenant flow

pecheneg12:
In tenant workflow method complete() call CarbonContext.getThreadLocalCarbonContext().getUsername() always return null.


hisanhunais:
Closing since the PR was stale for a long time. Please re-open if the fix is needed.

PR #12077, opened 2023-07-10
Allow registering external IDPs.

VirajSalaka:
Objective is to support third party IDPs in loosely coupled manner from api manager point of view. API manager will not communicate with the IDP at all, but it would store it so that the gateway can directly call the external IDP and validate all by itself. 

We bring in the tokenType call `EXTERNAL` to support this feature. And when for this type, issuer, JWKS URL, Token Endpoint, Authorize endpoint, Revoke Endpoint, Scope claim key and ConsumerKey claim key is mandatory.


components/apimgt/org.wso2.carbon.apimgt.internal.service/src/main/resources/api.yaml
```diff
@@ -1218,6 +1218,7 @@ definitions:
           - EXCHANGED
           - DIRECT
           - BOTH
+          - EXTERNAL
```
VirajSalaka, line 1221: Remove BOTH type from the Rest API level

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIAdminImpl.java
```diff
@@ -725,14 +776,39 @@ private String getDecryptedValue(String value) throws APIManagementException {
...
+        if (oldKeyManagerConfiguration == null) {
+            throw new APIManagementException("Key Manager Configuration not found for ID: "
+                    + keyManagerConfigurationDTO.getUuid(), ExceptionCodes.KEY_MANAGER_NOT_FOUND);
+        }
+        if (KeyManagerConfiguration.TokenType.EXTERNAL.equals(keyManagerTokenType)
+                ^ KeyManagerConfiguration.TokenType.EXTERNAL.equals(
+                        KeyManagerConfiguration.TokenType.valueOf(oldKeyManagerConfiguration.getTokenType()
+                                .toUpperCase()))) {
+            throw new APIManagementException("Cannot switch between tokenType when " +
+                    "the Key Manager's Token Type is External", ExceptionCodes.KEY_MANAGER_NOT_FOUND);
```
malinthaprasan, line 801: Let's fix the Error code

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIAdminImpl.java
```diff
@@ -116,6 +118,20 @@ public class APIAdminImpl implements APIAdmin {
...
+            APIConstants.KeyManager.SCOPES_CLAIM,
+            APIConstants.KeyManager.CERTIFICATE_TYPE,
+            APIConstants.KeyManager.CERTIFICATE_VALUE
+    };
+
+    private static final String[] requiredEndpointConfigurationsForExternalKeyManagers = {
+            APIConstants.KeyManager.AUTHORIZE_ENDPOINT,
+            APIConstants.KeyManager.TOKEN_ENDPOINT,
+            APIConstants.KeyManager.REVOKE_ENDPOINT
+    };
```
malinthaprasan, line 133: Shall we move this to APIConstants or any other better place

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIAdminImpl.java
```diff
@@ -116,6 +118,20 @@ public class APIAdminImpl implements APIAdmin {
     protected ApiMgtDAO apiMgtDAO;
     protected ChoreoApiMgtDAO choreoApiMgtDAO;
 
+    private static final String[] requiredConfigurationsForExternalKeyManagers = {
```
malinthaprasan, line 121:
```suggestion
    private static final String[] REQUIRED_CONF_FOR_EXT_KMS = {
```

malinthaprasan (approved):

malinthaprasan (approved):

PR #12084, opened 2023-07-17
Fix security for disabled scenario for API addition considering open API import

slahirucd7:
As of now when there are no extensions in the imported API definition, security is not going to be applied. With this PR.  it will be fixed.


rashm1n (approved):
LGTM

PR #12095, opened 2023-07-20
Fixing issue in details of additional Properties in /apis API

HiranyaKavishani:
Fix: https://github.com/wso2/api-manager/issues/1997


HiranyaKavishani:
Close PR as it is unnecessary to include additional properties for **/apis** REST API.

PR #12124, opened 2023-08-15
Implement version range subscription get from API ID

chathuranga95:
$subject

Adds, query param: `includeFromVersionRange: boolean`.

When `includeFromVersionRange=true`, the `GET /subscriptions` for `apiId` will return the subscriptions which are inherited from the version range.

When `includeFromVersionRange=false` or not specified, the exact subscriptions will be returned as per the current implementation.


components/apimgt/org.wso2.carbon.apimgt.rest.api.store.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/store/v1/impl/SubscriptionsApiServiceImpl.java
```diff
@@ -110,18 +111,45 @@ public Response subscriptionsGet(String apiId, String applicationId, String grou
...
+                    apiName = apiTypeWrapper.getApi().getId().getApiName();
+                }
+                if (includeFromVersionRange) {
+                    String versionRange;
+                    try {
+                        versionRange = apiConsumer.getSubscriptionVersionRange(apiVersion, apiTypeWrapper.getUuid());
+                    } catch (APIManagementException e) {
+                        RestApiUtil.handleBadRequest("Version range is not accepted for the API", log);
+                        return null;
+                    }
```
malinthaprasan, line 134: Do we need a try catch here? Just throw the error from the method itself. It will be handled from the REST API level.

malinthaprasan: public Response subscriptionsGet(String apiId, String applicationId, String groupId, String xWSO2Tenant,
                                     Boolean includeFromVersionRange, Integer offset, Integer limit,
                                     String ifNoneMatch, MessageContext messageContext) throws APIManagementException

PR #12235, opened 2024-01-30
Fix an issue when importing API products with overwrite query parameter set to true

SavinduDimal:
### Purpose

- Resolves https://github.com/wso2/api-manager/issues/2432
- When importing API products with overwrite query parameter set to true in the API request, default API product version value ("1.0.0") was used. In this PR it is changed to the correct version of the API product.


HeshanSudarshana (approved):

SavinduDimal:
Merging the PR since only `codecov/patch` has failed

PR #12312, opened 2024-03-07
Remove the Client Secret View option in OAUTH2 Endpoint security

shnrndk:
## Problem
For both Basic and digest auth, the password cant be seen by anyone who has access.
But in case of OAuth 2.0, if the API is public, all the users can see the masked client secret by clicking on view option, which is not available for passwords.

## Approach
Handled in the password scenario in Basic and Digest Authentication and used a similar approach for the clientSecret of OAuth secured endpoints.


tharikaGitHub:
Merging as the test failure will be fixed with https://github.com/wso2/product-apim/pull/13466

tharikaGitHub (approved):

PR #12381, opened 2024-03-27
pdfbox bump

GihanAyesh:
### Purpose
pdfbox bump to 3.0.1 and add necessary pdfbox-io orbit jar

The reason to create pdfbox-io as a orbit jar is to remove sun-misc import


AnuGayan (approved):

tharikaGitHub:
Merging as the test failure in the PR builder is an intermittent test failure.

PR #12412, opened 2024-04-02
[Test] [Public] Add unit test for doc name validation

dulithsenanayake:
This PR adds a unit test for doc name validations.


tharikaGitHub:
Merging as this is a unit test and the carbon-apimgt build is successful.

tharikaGitHub (approved):

PR #12497, opened 2024-07-22
Public fixes for Fix operation policies not updating issue with API import flow and Avoid gateway from sending an expired backend JWT token to the backend endpoint

piyumaldk:
Porting 4.3 fixes which is referred this PR


piyumaldk:
Built the related components successfully with unit tests

RakhithaRR (approved):

PR #12608, opened 2024-09-27
Fix Transformer Issue

Avishka-Shamendra:
## Purpose
This PR fixes an issue identified with the transformer used

PR #12645, opened 2024-10-08
Replace Transformer with Secured Transformer

Avishka-Shamendra:
## Purpose

Re sending https://github.com/wso2/carbon-apimgt/pull/12608 
This PR fixes an issue identified with the transformer used


Avishka-Shamendra:
Hi all,

This PR requires a SAXON bump, which we will do later

Thank you
Avishka

PR #12709, opened 2024-11-25
Fix Default API Context Generation Issue

dakshina99:
### Description
This PR resolves an issue with generating the default version context for APIs. Previously, the implementation unintentionally removed multiple occurrences of /version-like segments from the API context, even when these segments were not part of the version.

### Fix
The fix utilizes the CONTEXT_TEMPLATE stored in the database to generate the default version context. Specifically, all occurrences of /{version} and {version} are removed from the CONTEXT_TEMPLATE, and the resulting string is used as the context for the default version.
- Resolves https://github.com/wso2/api-manager/issues/3364

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/entity/API.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/entity/API.java
@@ -35,6 +35,7 @@ public class API implements CacheableEntity<String> {
     private String name = null;
     private String version = null;
     private String context = null;
+    private String contextTemplate = null;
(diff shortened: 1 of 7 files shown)


AnuGayan (approved):

components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/model/entity/API.java
```diff
@@ -35,6 +35,7 @@ public class API implements CacheableEntity<String> {
     private String name = null;
     private String version = null;
     private String context = null;
+    private String contextTemplate = null;
```
SavinduDimal, line 38: Shall we remove the null assignment since it is redundant? Check similar places as well.

PR #12720, opened 2024-12-06
B2B Scripts and enable application sharing for Orgs

HiranyaKavishani:
Fix : https://github.com/wso2/api-manager/issues/3413


HiranyaKavishani:
This is already added in public code. So closing the PR

PR #12734, opened 2025-01-03
extend-monetization-phase2

Chilliwiddit:
## Purpose
To publish additional metrics for each API event

## Goals
Allow users to monetize APIs using metrics of their choice

## Approach
Measure and append different metric of an API request for each event that is published which includes:
-Payload size
-GraphQL Complexity
-GraphQL Depth


Chilliwiddit:
PR closed as it does not point to the feature branch and moved to [#13088](https://github.com/wso2/carbon-apimgt/pull/13088)

PR #12735, opened 2025-01-03
extend-monetization-phase3

Chilliwiddit:
## Purpose
To separate the adding and removing of monetized subscriptions logic from the billing engine

## Goals
Allows the subscriptions workflows to be more modular and customizable by allowing the interchange of different workflow classes

## Approach
Move the monetized subscription logic to a seperate class where they will be called be a the monetizationImpl class


Chilliwiddit:
PR closed as it does not point to the feature branch and moved to [#13089](https://github.com/wso2/carbon-apimgt/pull/13089)

PR #12830, opened 2025-02-05
Fix endpoint certificate retrieval by using FQDN

nimsara66:
## Purpose
Fix the issue where the 'General Endpoint Configurations' section in the Publisher Portal does not display applied certificates from other APIs with the same domain name.

image

## Approach
The current implementation for retrieving certificates in the Publisher Portal relies on an exact match with the applied endpoint of the API. As a result, if another API shares the same domain name but has a different endpoint path, the associated certificate is correctly applied underneath during backend communication but does not appear in the UI.
Ex: `https://google.com` and `https://google.com/api/v1`

This PR addresses the issue by replacing the exact match (`=` operator) with the `like` operator and wildcard (`%`) syntax. This ensures that certificate retrieval is based on the fully qualified domain name (FQDN) rather than the exact endpoint, providing a more accurate and consistent UI representation.

image

## Related to 
Issue: https://github.com/wso2/api-manager/issues/3527
Internal: https://github.com/wso2-enterprise/wso2-apim-internal/issues/8460

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/CertificateMgtDAO.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/CertificateMgtDAO.java
@@ -35,6 +35,7 @@ import org.wso2.carbon.apimgt.impl.utils.APIUtil;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -42,6 +43,8 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
+
+import org.wso2.carbon.apimgt.impl.APIConstants;
 
 /**
  * This class handles data base transactions related to the server certificate management.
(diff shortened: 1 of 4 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dao/CertificateMgtDAO.java
```diff
@@ -341,6 +342,23 @@ public List<CertificateMetadataDTO> getCertificates(String alias, String endpoin
...
 
         if (StringUtils.isNotEmpty(alias) || StringUtils.isNotEmpty(endpoint)) {
+            if (StringUtils.isNotEmpty(endpoint)) {
+                try {
+                    if (log.isDebugEnabled()) {
+                        log.debug("The endpoint is not empty. Generating fqdn from endpoint " + endpoint);
+                    }
+
+                    // Remove any placeholders (curly braces with contents inside)
+                    String sanitizedEndpoint = endpoint.replaceAll("\\{.*?}", "");
```
AnuGayan, line 352: Lets make these Constants

nimsara66: Extracted the regex pattern into [impl/APIConstants.java](https://github.com/wso2/carbon-apimgt/pull/12830/files#diff-53180db9d61eccf226f337dec15f32bb7e951677a1e8b930daef9c9c30828d17) as contants

AnuGayan (approved):

PR #12843, opened 2025-02-06
Fix incorrect parent org id check when adding new org

NethmiRanasinghe:


tharikaGitHub:
It is better if we can add a description for the PR.

PR #12927, opened 2025-02-14
Enclose regex patterns with single quotes for proper parsing of escape characters.

Oshanath:


Avishka-Shamendra:
Shall we add a PR description?

PR #13154, opened 2025-06-23
Change AM_APPLICATION_REGISTRATION table

DDH13:
Changed the Inputs type from VARCHAR to BLOB


features/apimgt/org.wso2.carbon.apimgt.core.feature/src/main/resources/sql/mysql.sql
```diff
@@ -1728,7 +1728,7 @@ CREATE TABLE IF NOT EXISTS AM_APPLICATION_REGISTRATION (
     APP_ID INT,
     TOKEN_TYPE VARCHAR(30),
     TOKEN_SCOPE VARCHAR(1500) DEFAULT 'default',
-    INPUTS VARCHAR(1000),
+    INPUTS LONGBLOB,
```
shilmyhasan, line 1731: Is there any specific reason that we have set the type to LONGBLOB instead of BLOB?

Since we are converting a columns of varchar(1000) the BLOB should be well sufficient as it contains 64kb size. As Memory and other overhead are higher for LONGBLOB, it would be ideal to use BLOB if its possible.

DDH13: This concern was discussed beforehand within the team, and it was decided to use LONGBLOB to be on the safe side

PR #13462, opened 2025-10-04
Fix internal server error when adding deny policy with invalid condition

s-vamshi:
$subject

Fix: [Adding an deny policy with invalid condition result in internal server error #4253](https://github.com/wso2/api-manager/issues/4253)

changed from handleInternalServerError to handleBadRequest to return 400 bad request when POST on deny-policies is called

https://github.com/user-attachments/assets/e9280084-f7ec-4b78-90b0-579cefd9cc7b

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.admin.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/v1/impl/ThrottlingApiServiceImpl.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.admin.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/v1/impl/ThrottlingApiServiceImpl.java
@@ -1480,7 +1480,7 @@ public class ThrottlingApiServiceImpl implements ThrottlingApiService {
             } else {
                 String errorMessage = "Error while adding Blocking Condition. Condition type: "
                         + body.getConditionType() + ", " + "value: " + body.getConditionValue() + ". " + e.getMessage();
-                RestApiUtil.handleInternalServerError(errorMessage, e, log);
+                RestApiUtil.handleBadRequest(errorMessage, e, log);
             }
         } catch (URISyntaxException e) {
             String errorMessage = "Error while retrieving Blocking Condition resource location: Condition type: "


components/apimgt/org.wso2.carbon.apimgt.rest.api.admin.v1/src/main/java/org/wso2/carbon/apimgt/rest/api/admin/v1/impl/ThrottlingApiServiceImpl.java
```diff
@@ -1480,7 +1480,7 @@ public Response throttlingDenyPoliciesPost(String contentType, BlockingCondition
             } else {
                 String errorMessage = "Error while adding Blocking Condition. Condition type: "
                         + body.getConditionType() + ", " + "value: " + body.getConditionValue() + ". " + e.getMessage();
-                RestApiUtil.handleInternalServerError(errorMessage, e, log);
+                RestApiUtil.handleBadRequest(errorMessage, e, log);
```
tharikaGitHub, line 1483: @s-vamshi with this fix, all other APIManagementExceptions that do not fall under resource already exists error in this method would return a 400 status code which will not be the ideal response. There can be 500 errors as well. We will have to handle this case differently. You can have a look at other serviceImpl level code exception handling we have done to get some idea.

s-vamshi: @tharikaGitHub thanks for reviewing this pr.
I agree with you the fix for this shouldn't be this way. I have overlooked. I will get back with a different approach

s-vamshi: In ApiMgtDAO class we are throwing APIManagementException when something invalid is given as input in addBlockConditions method
may plan is to create a new exception which extends APIManagementException and to handle this newly created exception

@tharikaGitHub can you please let me know your thoughts on this?

tharikaGitHub: Hi @s-vamshi, yes you can do something similar. You should be able to do this by throwing an ExceptionCode. We don't have to write a new exception class, but instead can throw a specific exception code with the APIManagement exception so that we can identify that the issue is due to an invalid user input. Check if we could use existing exception codes for this. If not we will have to introduce new. We have done this in some of the methods of the ApiMgtDAO class.

s-vamshi: @tharikaGitHub got it, will go with this approach!

s-vamshi: @tharikaGitHub for setting `errorcode ` while creating new exception code should next incremental code be used or a different no.?

tharikaGitHub: Hi @s-vamshi, yes you can use an incremental number, but the error code should be a non existing one previously.

PR #13868, opened 2026-06-11
Fix wildcard certificate deletion silently breaking unrelated APIs

### LINKED ISSUE #5053: Wildcard certificate deletion silently breaks unrelated APIs sharing the same hostname
### Description

Hi Team,

When two APIs share the same hostname (e.g. via a wildcard cert) but have different endpoint URLs, deleting the certificate linked to the first API does not warn about the second API's dependency, causing silent SSL failures on the second API. From the perspective of API-2's owner and its consumers, the API breaks without any changes on their side, which is confusing since there is no indication in the Publisher UI that anything certificate related was changed.

As per the customer, publisher UI should detect that the certificate covers a wildcard hostname (*.cbre.com) and warn that other APIs with endpoints on the same hostname domain are implicitly dependent on this certificate before allowing deletion.
Actual behavior is publisher UI only checks for APIs that explicitly uploaded the same certificate. Since API-2 never uploaded a certificate, no dependency is recorded in AM_CERTIFICATE_METADATA then the UI shows no warning and allows deletion, then silently breaking API-2.

Thanks,
Shameera

### Steps to Reproduce

Create API-1 with endpoint: https://test.cbre.com/api
→ Upload SSL certificate via Publisher UI for this endpoint
→ Verify API-1 invocation works (200 OK)

Create API-2 with endpoint: https://prod.cbre.com/api
→ Do NOT upload any certificate (both hosts share the same wildcard cert *.cbre.com)
→ Verify API-2 invocation works (200 OK) — SSL succeeds because the cert is already in the truststore

From API-1, navigate to:
Endpoints → General Endpoint Config → Certificates → Delete certificate

Publisher UI shows NO warning about API-2 depending on this certificate

Confirm deletion

Invoke API-2 → SSL handshake failure

### Version

4.6.0.27

_No response_

Tharsanan1:
## Summary

- When API-1 uploads a wildcard endpoint cert (e.g. `*.am.wso2.com`) and API-2 uses a different subdomain of the same wildcard (e.g. `idp.am.wso2.com`) without explicitly uploading a cert, deleting the cert for API-1 silently breaks API-2 with no warning.
- Added `searchPaginatedAPIsByCertificate` which parses X.509 SANs from the certificate (with wildcard-aware matching) and returns all APIs whose endpoint configs match the wildcard domain.
- The certificate usage endpoint (`GET /endpoint-certificates/{alias}/usage`) now uses this instead of a plain FQDN match, so the Publisher correctly warns about all dependent APIs before allowing deletion.

## Root cause

`getCertificateUsageByAlias` was calling `searchPaginatedAPIsByFQDN(endpoint, ...)` which only matched APIs with the exact endpoint URL stored against the certificate. APIs on other subdomains of the same wildcard cert were not found.

## Changes

- **`APIProvider.java`** — new interface method `searchPaginatedAPIsByCertificate`
- **`APIProviderImpl.java`** — implementation that extracts SANs from the certificate, converts wildcard SANs (e.g. `*.am.wso2.com` → `.am.wso2.com`) to substring search terms, and queries Solr with `endpointConfig:` prefix
- **`CertificateMgtUtils.java`** — new utility method `getEndpointSearchTermsFromCertificate` that parses X.509 SANs (DNS + IP) with CN fallback
- **`EndpointCertificatesApiServiceImpl.java`** — switch usage endpoint to use the new method
- **`APIProviderImplTest.java`** / **`CertificateMgtUtilTest.java`** — unit tests

## Test plan

- [ ] Upload a wildcard cert (`*.am.wso2.com`) for API-1 endpoint (`api.am.wso2.com`)
- [ ] `GET /endpoint-certificates/{alias}/usage` returns both API-1 and API-2 (endpoint: `idp.am.wso2.com`)
- [ ] Publisher UI shows "wso2test-wildcard is used by 2 APIs" warning before deletion
- [ ] After deletion both APIs return HTTP 500 (PKIX SSL error)

Fixes: https://github.com/wso2/api-manager/issues/5053

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Tharsanan1:
@coderabbitai review

PasanT9 (approved):

PR #13901, opened 2026-07-02
Preserve MCP tool metadata (annotations, _meta, outputSchema) when proxying MCP servers

Shamly-Shanawaz:
### Problem
When an MCP server is proxied through API Manager ("Proxy Existing MCP Server"), the `tools/list` response only returned `name`, `description`, and `inputSchema`. Other tool-level fields sent by the upstream server (`annotations`, `_meta`, `outputSchema`, `title`, `execution`, etc.) were dropped, so clients saw less metadata than on a direct connection.

**Root cause:** Only the tool's `inputSchema` was persisted in `AM_API_URL_MAPPING.SCHEMA_DEFINITION`, and the `tools/list` builder could emit only `name`/`description`/`inputSchema`.

Related Issue : https://github.com/wso2/api-manager/issues/5120

### Fix
Retain the complete tool definition for proxied MCP servers and return all of its fields in the tools/list response. Fully backward compatible .MCP servers created before this change continue to work as before.

### Scope
* Only the proxy (`SERVER_PROXY`) path stores the richer definition.
* OpenAPI and existing-API creation paths are unchanged (their sources carry no such metadata).
* `name` and `description` continue to be served from their own columns.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/MCPInitializerAndToolFetcher.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/MCPInitializerAndToolFetcher.java
@@ -281,6 +281,25 @@ public class MCPInitializerAndToolFetcher {
     }
 
     /**
+     * Build the schema definition for a tool: all tool fields except name and description
+     * (i.e. the input schema plus any metadata such as annotations, _meta and outputSchema).
+     *
+     * @param toolJson Tool definition returned by the MCP server
+     * @return JSON object containing all tool fields except name and description
+     */
+    public static org.json.JSONObject buildToolMetadata(org.json.JSONObject toolJson) {
+
+        org.json.JSONObject toolMetadata = new org.json.JSONObject();
+        for (String key : toolJson.keySet()) {
+            if (!APIConstants.MCP.TOOL_NAME_KEY.equals(key)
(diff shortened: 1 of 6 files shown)


components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/MCPInitializerAndToolFetcher.java
```diff
@@ -280,6 +280,25 @@ private String extractJsonCandidate(String responseText) {
...
     }
 
+    /**
+     * Build the schema definition for a tool: all tool fields except name and description
+     * (i.e. the input schema plus any metadata such as annotations, _meta and outputSchema).
+     *
+     * @param toolJson Tool definition returned by the MCP server
+     * @return JSON object containing all tool fields except name and description
+     */
+    public static org.json.JSONObject buildToolMetadata(org.json.JSONObject toolJson) {
```
AqeelMuhammad: Can't we just use `JSONObject` here?

Shamly-Shanawaz: changed to JSONObject as suggested

AqeelMuhammad (dismissed):

Shamly-Shanawaz:
@coderabbitai review

PasanT9 (approved):

PR #13903, opened 2026-07-06
Improve distributed throttle counter reset reliability and add Redis connection pool config

DakshithaS:
## Purpose

Address incorrect throttling decisions in distributed deployments caused by non-atomic counter resets during time window expiry. Previously, window resets were executed via sequential decrement (`processRemove`) operations, exposing intermediate counter states to other Traffic Managers during KV store synchronization. This resulted in inconsistent reads and temporary over-allowance of requests.

## Goals

- Ensure atomic reset of distributed counters at window boundaries
- Eliminate intermediate state visibility across Traffic Managers  
- Improve consistency of throttling decisions in distributed environments
- Prevent sync thread hangs under slow/unresponsive Redis instances

---

### Problem
`JedisPool` was initialized with only a `connectionTimeout`, leaving socket reads and pool-borrow waits unbounded. Under a slow or unresponsive Redis instance, this caused sync threads to hang indefinitely, blocking throttling operations.

### Solution
- Added **`SocketTimeout`** config option to cap socket read/write operations
- Added **`MaxWaitMillis`** config option to limit pool connection wait time
- Wired both through `APIConstants`, `APIManagerConfiguration`, `DistributedThrottleConfig`, and `JedisKeyValueStoreClient`
- **SocketTimeout** defaults to `Protocol.DEFAULT_TIMEOUT` (2000ms) when not configured
- **MaxWaitMillis** defaults to **3000ms** for throttling operations, ensuring bounded waits

### Configuration

**deployment.toml**
```toml
[apim.distributed_throttling.keyvalue_store]
socket_timeout = 2000     # caps socket read/write time (ms)
max_wait_millis = 3000    # max wait for a pool connection (ms)
```

---

### Problem
The previous `reset()` implementation had multiple race conditions:

1. **Blocking Reset**: Made a synchronous Redis `SET key 0` call on the Siddhi event thread, blocking the throttle pipeline at every window boundary
2. **Old-Window Delta Race**: After `reset()` wrote `0` to Redis, in-flight sync operations could race back and overwrite with old window counts
3. **Stale Keys**: Crashed nodes left counters in Redis without expiry, causing incorrect counts when the node restarted
4. **Sequential Decrements**: Window resets via `processRemove` exposed intermediate states to concurrent sync operations

#### Non-Blocking Async Reset
- `reset()` now sets `pendingReset = true` flag and returns immediately (no blocking Redis call)
- Background sync thread detects `pendingReset` and flushes via `PSETEX key 0 <remaining-ttl>`
- Automatic retry on failure ensures eventual consistency
- Siddhi event thread never blocks on Redis I/O

#### Race Condition Guards
- **Double-check pattern**: After `unsyncedCounter.getAndSet(0)`, verify `pendingReset` again
- If `reset()` fired concurrently, discard the captured old-window delta
- Prevents old counts from overwriting the reset value

#### TTL-Aware Keys
- All Redis keys now carry a window TTL matching the time-window duration
- Implemented via `PSETEX` for resets and `PEXPIRE` for incremental updates
- Stale keys auto-expire if nodes crash mid-window
- `ThrottleStreamProcessor` propagates window expiry timestamp via `ThreadLocal`
- Aggregators refresh `storedWindowExpiry` from ThreadLocal to handle gaps (empty windows)

#### Key-Value Store Interface Extensions
- Added `setWithExpiry(key, value, ttlMillis)` → Redis `PSETEX`
- Added `expireMillis(key, ttlMillis)` → Redis `PEXPIRE`
- Integrated into `JedisKeyValueStoreClient` implementation

---

## Related Issue
- https://github.com/wso2/api-manager/issues/4744

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
@@ -618,6 +618,7 @@ public class APIManagerConfiguration {
                     OMElement password = propertiesElement.getFirstChildWithName(new QName(APIConstants.DISTRIBUTED_THROTTLE_PASSWORD));
                     OMElement databaseId = propertiesElement.getFirstChildWithName(new QName(APIConstants.DISTRIBUTED_THROTTLE_DATABASE_ID));
                     OMElement connectionTimeout = propertiesElement.getFirstChildWithName(new QName(APIConstants.DISTRIBUTED_THROTTLE_CONNECTION_TIMEOUT));
+                    OMElement socketTimeout = propertiesElement.getFirstChildWithName(new QName(APIConstants.DISTRIBUTED_THROTTLE_SOCKET_TIMEOUT));
                     OMElement isSslEnabled = propertiesElement.getFirstChildWithName(new QName(APIConstants.DISTRIBUTED_THROTTLE_IS_SSL_ENABLED));
 
                     if (host != null && StringUtils.isNotBlank(host.getText())) {
@@ -654,7 +655,15 @@ public class APIManagerConfiguration {
                             distributedThrottleConfig.setConnectionTimeout(Integer.parseInt(connectionTimeout.getText().trim()));
                         } catch (NumberFormatException e) {
                             log.warn("Invalid connectionTimeout value: " + connectionTimeout.getText(), e);
-                        }                    }
+                        }
+                    }
+                    if (socketTimeout != null) {
+                        try {
+                            distributedThrottleConfig.setSocketTimeout(Integer.parseInt(socketTimeout.getText().trim()));
+                        } catch (NumberFormatException e) {
+                            log.warn("Invalid socketTimeout value: " + socketTimeout.getText(), e);
+                        }
+                    }
                     if (isSslEnabled != null) {
                         distributedThrottleConfig.setSslEnabled(Boolean.parseBoolean(isSslEnabled.getText().trim()));
                     }
@@ -683,6 +692,8 @@ public class APIManagerConfiguration {
                                 distributedThrottleConfig.setTimeBetweenEvictionRunsMillis(Long.parseLong(propertyNode.getText()));
                             } else if (APIConstants.DISTRIBUTED_THROTTLE_NUM_TESTS_PER_EVICTION_RUNS.equals(propertyNode.getLocalName())) {
                                 distributedThrottleConfig.setNumTestsPerEvictionRun(Integer.parseInt(propertyNode.getText()));
+                            } else if (APIConstants.DISTRIBUTED_THROTTLE_MAX_WAIT_MILLIS.equals(propertyNode.getLocalName())) {
+                                distributedThrottleConfig.setMaxWaitMillis(Long.parseLong(propertyNode.getText()));
                             }
                         }
                     }
--- a/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/DistributedThrottleConfig.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/DistributedThrottleConfig.java
@@ -47,6 +47,8 @@ public class DistributedThrottleConfig {
     private int numTestsPerEvictionRun = -1;
     private boolean isSslEnabled;
     private int connectionTimeout;
+    private int socketTimeout;
+    private long maxWaitMillis = 3000L;
 
     public String getHost() {
         return host;
@@ -223,4 +225,19 @@ public class DistributedThrottleConfig {
         this.databaseId = databaseId;
     }
 
+    public int getSocketTimeout() {
+        return socketTimeout;
+    }
+
+    public void setSocketTimeout(int socketTimeout) {
+        this.socketTimeout = socketTimeout;
+    }
+
+    public long getMaxWaitMillis() {
+        return maxWaitMillis;
+    }
+
+    public void setMaxWaitMillis(long maxWaitMillis) {
+        this.maxWaitMillis = maxWaitMillis;
+    }
 }
--- a/components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
@@ -24,6 +24,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -42,7 +43,7 @@ import org.wso2.siddhi.query.api.definition.Attribute;
 public class DistributedCountAttributeAggregator extends AttributeAggregator {
 
     private static final Log log = LogFactory.getLog(DistributedCountAttributeAggregator.class);
-    private static Attribute.Type type = Attribute.Type.LONG;
+    private static final Attribute.Type type = Attribute.Type.LONG;
     private KeyValueStoreClient kvStoreClient;
     private String key;
     private final AtomicLong localCounter = new AtomicLong(0L);
@@ -68,8 +69,11 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
 
     // Static shared scheduler for all aggregators
     private static ScheduledExecutorService kvStoreSyncScheduler = null;
-    private static final ScheduledExecutorService masterScheduler =
-            Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, Thread.currentThread().getName()));
+    private static ScheduledFuture<?> masterSyncTask = null;
+    private static ScheduledExecutorService masterScheduler = null;
+    private volatile boolean pendingReset = false;
+    private volatile long storedWindowExpiry = 0L;
+    private volatile boolean keyHasTTL = false;
 
 
     /**
@@ -83,13 +87,16 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
         if (DISTRIBUTED_THROTTLE_CONFIG == null) {
             synchronized (DistributedCountAttributeAggregator.class) {
                 if (DISTRIBUTED_THROTTLE_CONFIG == null) {
-                    DISTRIBUTED_THROTTLE_CONFIG = getDistributedThrottleConfig();
+                    DistributedThrottleConfig config = getDistributedThrottleConfig();
+                    if (config != null) {
+                        // Plain writes BEFORE the volatile store — any thread that later reads
+                        // DISTRIBUTED_THROTTLE_CONFIG as non-null sees these via JMM happens-before.
+                        distributedThrottlingEnabled = config.isEnabled();
+                        corePoolSize = config.getCorePoolSize();
+                        kvStoreSyncIntervalMilliseconds = config.getSyncInterval();
+                        DISTRIBUTED_THROTTLE_CONFIG = config; // volatile write last
+                    }
                 }
-            }
-            if (DISTRIBUTED_THROTTLE_CONFIG != null) {
-                distributedThrottlingEnabled = DISTRIBUTED_THROTTLE_CONFIG.isEnabled();
-                corePoolSize = DISTRIBUTED_THROTTLE_CONFIG.getCorePoolSize();
-                kvStoreSyncIntervalMilliseconds = DISTRIBUTED_THROTTLE_CONFIG.getSyncInterval();
             }
         }
         String throttleKey = QuerySelector.getThreadLocalGroupByKey();
@@ -125,7 +132,7 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
                 long initialValue = Long.parseLong(kvStoreValue);
                 localCounter.set(initialValue);
             } else {
-                kvStoreClient.set(key, "0");
+                writeCounterValue("0");
             }
         } catch (Exception e) {
             log.error("Error initializing from key-value store for key " + key, e);
@@ -134,30 +141,107 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
     }
 
     /**
+     * Writes the given value to the key-value store with the remaining TTL of the current window.
+     * Does nothing when {@code storedWindowExpiry} is not yet known or has already passed.
+     *
+     * @param value the value to store for this aggregator's key.
+     */
+    private void writeCounterValue(String value) {
+        long windowExpiry = storedWindowExpiry;
+        if (windowExpiry > 0) {
+            long remainingMillis = windowExpiry - System.currentTimeMillis();
+            if (remainingMillis > 0) {
+                kvStoreClient.setWithExpiry(key, value, remainingMillis);
+            }
+        }
+    }
+
+    /**
      * Synchronize the local counter with the key-value store.
-     * Update the key-value store with the unsynced counter value.
+     * Pending resets (PSETEX "0") are flushed first and retried on failure; then any
+     * accumulated delta is pushed via INCRBY/DECRBY, or the current Redis value is pulled
+     * when there is no local change.
      */
     private void syncWithKVStore() {
         synchronized (kvStoreLock) {
             if (kvStoreClient == null || key == null) {
                 return;
             }
+            if (pendingReset) {
+                try {
+                    writeCounterValue("0");
+                    // Do NOT reset localCounter here — reset() already zeroed it at window boundary.
+                    // Any increments to localCounter since then belong to the new window and are valid.
+                    pendingReset = false; // cleared only on success — failure retried next tick
+                    keyHasTTL = true;
+                } catch (KeyValueStoreException e) {
+                    long currentTimeMillis = System.currentTimeMillis();
+                    if (currentTimeMillis - lastErrorLogTimestamp.get() > ERROR_LOG_INTERVAL_MS) {
+                        log.error("Error resetting counter in key-value store for key " + key, e);
+                        lastErrorLogTimestamp.set(currentTimeMillis);
+                    }
+                }
+                return;
+            }
             long currentUnsyncedCount = unsyncedCounter.getAndSet(0L);
+            if (pendingReset) {
+                // reset() fired between our initial pendingReset check and getAndSet —
+                // discard the captured old-window delta; the next tick will PSETEX 0.
+                return;
+            }
             try {
                 if (currentUnsyncedCount == 0) {
-                    localCounter.set(Long.parseLong(kvStoreClient.get(key)));
-                } else if (currentUnsyncedCount > 0) {
-                    localCounter.set(kvStoreClient.incrementBy(key, currentUnsyncedCount));
+                    String kvStoreValue = kvStoreClient.get(key);
+                    if (kvStoreValue == null) {
+                        writeCounterValue("0");
+                        localCounter.set(0L);
+                    } else if (!pendingReset) {
+                        // skip if reset() fired while waiting for the GET response —
+                        // the returned value is the old-window total and must not overwrite localCounter=0.
+                        localCounter.set(Long.parseLong(kvStoreValue));
+                    }
                 } else {
-                    localCounter.set(kvStoreClient.decrementBy(key, Math.abs(currentUnsyncedCount)));
+                    long kvStoreValue;
+                    if (currentUnsyncedCount > 0) {
+                        kvStoreValue = kvStoreClient.incrementBy(key, currentUnsyncedCount);
+                    } else {
+                        kvStoreValue = kvStoreClient.decrementBy(key, Math.abs(currentUnsyncedCount));
+                    }
+                    // INCRBY/DECRBY never sets a TTL. Stamp one the first time after a gap
+                    // (empty windows) where the key was recreated without a TTL.
+                    if (!keyHasTTL) {
+                        long expiry = storedWindowExpiry;
+                        long remainingMillis = expiry - System.currentTimeMillis();
+                        if (remainingMillis > 0) {
+                            try {
+                                kvStoreClient.expireMillis(key, remainingMillis);
+                                keyHasTTL = true;
+                            } catch (KeyValueStoreException e) {
+                                // Non-fatal: retried on the next sync tick while keyHasTTL is false.
+                                long now = System.currentTimeMillis();
+                                if (now - lastErrorLogTimestamp.get() > ERROR_LOG_INTERVAL_MS) {
+                                    log.warn("Could not set TTL via PEXPIRE for key " + key, e);
+                                    lastErrorLogTimestamp.set(now);
+                                }
+                            }
+                        }
+                    }
+                    if (!pendingReset) {
+                        // guard: reset() can fire while the INCRBY/DECRBY is in flight;
+                        // discard the result to avoid reinstating old-window counts in localCounter.
+                        localCounter.set(kvStoreValue);
+                    }
                 }
-            } catch (KeyValueStoreException e) {
+            } catch (KeyValueStoreException | NumberFormatException e) {
                 long currentTimeMillis = System.currentTimeMillis();
                 if (currentTimeMillis - lastErrorLogTimestamp.get() > ERROR_LOG_INTERVAL_MS) {
                     log.error("Error syncing with key-value store for the key " + key, e);
                     lastErrorLogTimestamp.set(currentTimeMillis);
                 }
-                unsyncedCounter.addAndGet(currentUnsyncedCount);
+                // skip restore if reset() fired after the snapshot — old-window delta must be discarded
+                if (!pendingReset) {
+                    unsyncedCounter.addAndGet(currentUnsyncedCount);
+                }
             }
         }
     }
@@ -177,11 +261,20 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
     @Override
     public Object processAdd(Object data) {
         try {
-            localCounter.incrementAndGet();
+            long newCount = localCounter.incrementAndGet();
             if (distributedThrottlingEnabled && kvStoreClient != null && key != null) {
                 unsyncedCounter.incrementAndGet();
+                // Refresh storedWindowExpiry when stale — this happens after one or more empty
+                // windows where reset() was never called (no RESET event → no reset()).
+                if (storedWindowExpiry <= System.currentTimeMillis()) {
+                    Long expiry = ThrottleStreamProcessor.getThreadLocalWindowExpiry();
+                    if (expiry != null && expiry > storedWindowExpiry) {
+                        storedWindowExpiry = expiry;
+                        keyHasTTL = false; // key was created by INCRBY after gap — needs PEXPIRE
+                    }
+                }
             }
-            return localCounter.get();
+            return newCount;
         } catch (Exception e) {
             log.error("Error in processAdd for key " + key, e);
             return localCounter.get();
@@ -190,7 +283,14 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
 
     @Override
     public Object processAdd(Object[] data) {
-        return processAdd((Object) data);
+        if (log.isDebugEnabled()) {
+            log.debug("DistributedCountAggregator: processAdd called with data: "
+                    + (data != null && data.length > 0 ? data[0] : null));
+        }
+        if (isResetRequested(data)) {
+            return reset();
+        }
+        return processAdd(data != null && data.length > 0 ? data[0] : null);
     }
 
 
@@ -204,6 +304,9 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
      */
     @Override
     public Object processRemove(Object data) {
+        if (log.isDebugEnabled()) {
+            log.debug("DistributedCountAggregator: processRemove called with data: " + data);
+        }
         try {
             localCounter.decrementAndGet();
             if (distributedThrottlingEnabled && kvStoreClient != null && key != null) {
@@ -219,31 +322,32 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
 
     @Override
     public Object processRemove(Object[] data) {
-        return processRemove((Object) data);
+        return processRemove(data != null && data.length > 0 ? data[0] : null);
     }
 
 
     /**
-     * Resets the local counter to zero.
-     * If distributed throttling is enabled, also resets the value in the distributed key-value store
-     * and clears any pending unsynced changes.
+     * Resets the local counter to zero and signals the background sync thread to PSETEX "0"
+     * on its next tick. Never blocks on a Redis network call. Transient failures are retried
+     * automatically.
      *
      * @return 0L after reset.
      */
     @Override
     public Object reset() {
-        try {
-            localCounter.set(0L);
-            if (distributedThrottlingEnabled && kvStoreClient != null && key != null) {
-                kvStoreClient.set(key, "0");
-                unsyncedCounter.set(0L); // Clear pending changes
-            }
-            return 0L;
-
-        } catch (KeyValueStoreException e) {
-            log.error("Error resetting counter for key " + key, e);
-            return 0L;
+        if (log.isDebugEnabled()) {
+            log.debug("DistributedCountAggregator: reset called");
         }
+        localCounter.set(0L);
+        if (distributedThrottlingEnabled && kvStoreClient != null && key != null) {
+            Long expiry = ThrottleStreamProcessor.getThreadLocalWindowExpiry();
+            if (expiry != null) {
+                storedWindowExpiry = expiry;
+            }
+            unsyncedCounter.set(0L);
+            pendingReset = true;
+        }
+        return 0L;
     }
 
     @Override
@@ -259,12 +363,18 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
      */
     @Override
     public void stop() {
+        if (log.isDebugEnabled()) {
+            log.debug("DistributedCountAggregator: stop called");
+        }
         try {
             // Only remove if key is not null and distributed throttling is enabled
             if (distributedThrottlingEnabled && key != null) {
                 ACTIVE_AGGREGATORS.remove(key);
                 if (kvStoreClient != null) {
-                    syncWithKVStore();
+                    syncWithKVStore(); // best-effort: flush pendingReset (PSETEX 0) if pending
+                    if (!pendingReset) {
+                        syncWithKVStore(); // best-effort: push new-window delta; both may be no-ops if Redis is down
+                    }
                 }
                 // Shutdown scheduler if no active aggregators exist
                 if (ACTIVE_AGGREGATORS.isEmpty()) {
@@ -278,6 +388,9 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
 
     @Override
     public Object[] currentState() {
+        if (log.isDebugEnabled()) {
+            log.debug("DistributedCountAggregator: currentState called");
+        }
         if (distributedThrottlingEnabled && kvStoreClient != null && key != null) {
             try {
                 syncWithKVStore();
@@ -290,21 +403,56 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
 
     @Override
     public void restoreState(Object[] state) {
+        if (log.isDebugEnabled()) {
+            log.debug("DistributedCountAggregator: restoreState called with state: " + state);
+        }
         Map.Entry<String, Object> stateEntry = (Map.Entry<String, Object>) state[0];
         long restoredValue = (Long) stateEntry.getValue();
 
-        localCounter.set(restoredValue);
-        unsyncedCounter.set(0L);
-
+        // In distributed mode, Redis is authoritative — ignore the stale single-node Siddhi snapshot.
+        // Fall back to 0 on key-absent or Redis error; the sync scheduler recovers the real value once Redis is up.
+        // In non-distributed mode, the Siddhi snapshot is the only state.
+        // Redis GET runs outside the lock to avoid blocking the sync thread during network I/O.
+        long counterToRestore = distributedThrottlingEnabled ? 0L : restoredValue;
+        boolean seedRedis = false;
         if (distributedThrottlingEnabled && kvStoreClient != null && key != null) {
             try {
-                kvStoreClient.set(key, String.valueOf(restoredValue));
+                String kvStoreValue = kvStoreClient.get(key);
+                if (kvStoreValue != null) {
+                    counterToRestore = Long.parseLong(kvStoreValue);
+                } else {
+                    seedRedis = true;
+                }
+            } catch (KeyValueStoreException | NumberFormatException e) {
+                log.error("Error reading from key-value store during restoreState for key "
+                        + key + ". Starting fresh at 0.", e);
+            }
+        }
+
+        // Lock only for the fast in-memory writes — no Redis call inside.
+        synchronized (kvStoreLock) {
+            unsyncedCounter.set(0L);
+            pendingReset = false;
+            localCounter.set(counterToRestore);
+        }
+
+        // Key absent in Redis — attempt to seed with 0; no-op if storedWindowExpiry is not yet known.
+        if (seedRedis) {
+            try {
+                writeCounterValue("0");
             } catch (KeyValueStoreException e) {
-                log.error("Error restoring state to key-value store for key "+ key, e);
+                log.error("Error seeding key-value store with 0 during restoreState for key " + key, e);
             }
         }
     }
 
+    /**
+     * Retrieves the distributed throttle configuration for the API Manager.
+     * Attempts to fetch the configuration from the service reference holder.
+     * If fetching the configuration fails, logs a warning message and returns null.
+     *
+     * @return the {@link DistributedThrottleConfig} instance if successfully loaded, or null if loading fails.
+     */
     private static DistributedThrottleConfig getDistributedThrottleConfig() {
         try {
             return ServiceReferenceHolder.getInstance()
@@ -334,13 +482,17 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
             }
             if (kvStoreSyncScheduler == null) {
                 kvStoreSyncScheduler = Executors.newScheduledThreadPool(corePoolSize, r ->
-                        new Thread(r, Thread.currentThread().getName()));
+                        new Thread(r, "apim-distributed-throttle-sync"));
+            }
+            if (masterScheduler == null) {
+                masterScheduler = Executors.newSingleThreadScheduledExecutor(r ->
+                        new Thread(r, "apim-distributed-throttle-master-sync"));
             }
 
             log.debug("Starting key-value store sync scheduler with interval: "
                     + kvStoreSyncIntervalMilliseconds + " ms, pool size: " + corePoolSize);
 
-            masterScheduler.scheduleAtFixedRate(() -> {
+            masterSyncTask = masterScheduler.scheduleAtFixedRate(() -> {
                 try {
                     CompletableFuture<?>[] futures = ACTIVE_AGGREGATORS.values().stream()
                             .map(aggregator -> CompletableFuture.runAsync(() -> {
@@ -373,21 +525,14 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
      */
     public static void shutdownScheduler() {
         synchronized (schedulerLock) {
-            if (kvStoreSyncScheduler != null && !kvStoreSyncScheduler.isShutdown()) {
-                log.debug("Shutting down key-value store sync scheduler...");
-                kvStoreSyncScheduler.shutdown();
-                try {
-                    if (!kvStoreSyncScheduler.awaitTermination(5, TimeUnit.SECONDS)) {
-                        log.warn("The key-value store sync scheduler did not terminate in time. Forcing shutdown...");
-                        kvStoreSyncScheduler.shutdownNow();
-                    }
-                } catch (InterruptedException e) {
-                    log.warn("Interrupted while shutting down key-value store sync scheduler. Forcing shutdown...");
-                    kvStoreSyncScheduler.shutdownNow();
-                    Thread.currentThread().interrupt();
-                }
+            if (masterSyncTask != null) {
+                masterSyncTask.cancel(false);
+                masterSyncTask = null;
             }
+            shutdownExecutor(kvStoreSyncScheduler, "key-value store sync scheduler");
             kvStoreSyncScheduler = null;
+            shutdownExecutor(masterScheduler, "master sync scheduler");
+            masterScheduler = null;
             schedulerStarted = false;
 
             // Shutdown the KeyValueStoreManager to close JedisPool
@@ -397,6 +542,34 @@ public class DistributedCountAttributeAggregator extends AttributeAggregator {
                 log.error("Error shutting down KeyValueStoreManager", e);
             }
         }
+    }
+
+    /**
+     * Shuts down the given executor service gracefully, forcing shutdown if it does not
+     * terminate within the timeout.
+     *
+     * @param executor The executor to shut down.
+     * @param name     Name of the executor for logging.
+     */
+    private static void shutdownExecutor(ScheduledExecutorService executor, String name) {
+        if (executor != null && !executor.isShutdown()) {
+            log.debug("Shutting down " + name + "...");
+            executor.shutdown();
+            try {
+                if (!executor.awaitTermination(5, TimeUnit.SECONDS)) {
+                    log.warn("The " + name + " did not terminate in time. Forcing shutdown...");
+                    executor.shutdownNow();
+                }
+            } catch (InterruptedException e) {
+                log.warn("Interrupted while shutting down " + name + ". Forcing shutdown...");
+                executor.shutdownNow();
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    private boolean isResetRequested(Object[] data) {
+        return data != null && data.length > 1 && Boolean.TRUE.equals(data[1]);
     }
 
 }
--- a/components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
@@ -18,6 +18,7 @@
 package org.wso2.carbon.apimgt.throttling.siddhi.extension;
 
 import java.util.AbstractMap;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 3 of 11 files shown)


components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
```diff
@@ -125,39 +132,116 @@ private void initializeFromKVStore() {
...
                 return;
             }
+            if (pendingReset) {
+                try {
+                    writeCounterValue("0");
+                    // Do NOT reset localCounter here — reset() already zeroed it at window boundary.
+                    // Any increments to localCounter since then belong to the new window and are valid.
+                    pendingReset = false; // cleared only on success — failure retried next tick
+                    keyHasTTL = true;
+                } catch (KeyValueStoreException e) {
```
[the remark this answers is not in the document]
DakshithaS: Self-healing by design: if the window expired, the Redis key's TTL already removed it, and keyHasTTL is corrected on the next processAdd

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/util/kvstore/JedisKeyValueStoreClient.java
```diff
@@ -48,6 +49,7 @@ public class JedisKeyValueStoreClient implements KeyValueStoreClient {
     private static volatile char[] password;
     private static volatile int databaseId;
     private static volatile int connectionTimeout;
+    private static volatile int socketTimeout;
```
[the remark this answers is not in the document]
DakshithaS: Already handled: the != 0 ? ... : Protocol.DEFAULT_TIMEOUT fallback is inside both config branches, and total config failure leaves host = null so no pool is ever create

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java
```diff
@@ -146,13 +149,25 @@ protected void process(ComplexEventChunk<StreamEvent> streamEventChunk, Processo
...
+                StreamEvent firstExpiredEvent = expiredEventChunk.getFirst();
+                if (firstExpiredEvent != null) {
+                    StreamEvent resetEvent = streamEventCloner.copyStreamEvent(firstExpiredEvent);
+                    resetEvent.setType(StreamEvent.Type.RESET);
+                    resetEvent.setTimestamp(expireEventTime);
+                    resetEventChunk = new ComplexEventChunk<StreamEvent>(true);
+                    resetEventChunk.add(resetEvent);
+                }
+                expiredEventChunk.clear();
             }
```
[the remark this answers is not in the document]
DakshithaS: Siddhi broadcasts RESET to every grouped aggregator instance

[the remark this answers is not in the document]
DakshithaS: @coderabbitai
in Siddhi, RESET is a broadcast signal; the group key inside the event is never read on the reset path.

Please verify against siddhi-core 3.3.0 (the version pinned in the root pom): check QuerySelector#processInBatchGroupBy (the case RESET: branch forwards the event to all attribute processors without calling constructEventKey()), and GroupByAggregationAttributeExecutor#execute (on RESET it iterates the entire aggregatorMap — one aggregator per group key — calling process() on each, which invokes reset()).

So one RESET event resets every active group's aggregator; firstExpiredEvent is only a template to carry the event type. This is the same mechanism Siddhi's own timeBatch window relies on to clear grouped aggregations.

DakshithaS:
@coderabbitai pls review and approve this

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/DistributedThrottleConfig.java
```diff
@@ -47,6 +47,8 @@ public class DistributedThrottleConfig {
     private int numTestsPerEvictionRun = -1;
     private boolean isSslEnabled;
     private int connectionTimeout;
+    private int socketTimeout;
+    private long maxWaitMillis = -1;
```
DakshithaS: set a value

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/APIManagerConfiguration.java
```diff
@@ -683,6 +692,8 @@ private void readChildElements(OMElement serverConfig,
                                 distributedThrottleConfig.setTimeBetweenEvictionRunsMillis(Long.parseLong(propertyNode.getText()));
                             } else if (APIConstants.DISTRIBUTED_THROTTLE_NUM_TESTS_PER_EVICTION_RUNS.equals(propertyNode.getLocalName())) {
                                 distributedThrottleConfig.setNumTestsPerEvictionRun(Integer.parseInt(propertyNode.getText()));
+                            } else if (APIConstants.DISTRIBUTED_THROTTLE_MAX_WAIT_MILLIS.equals(propertyNode.getLocalName())) {
+                                distributedThrottleConfig.setMaxWaitMillis(Long.parseLong(propertyNode.getText()));
```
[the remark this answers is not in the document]
DakshithaS: Leaving as-is for consistency with the other properties in this same loop (MaxTotal, MaxIdle, MinIdle, etc.), none of which have trim/try-catch either.

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/test/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/CountAttributeAggregatorWithResetTestCase.java
```diff
@@ -201,4 +203,139 @@ public void receive(long timeStamp, Event[] inEvents, Event[] removeEvents) {
...
+        inputHandler.send(new Object[]{"App1_ResourceA", "post-reset-msg", false});
+
+        execPlanRunTime.shutdown();
+
+        junit.framework.Assert.assertEquals("The targeted key should reflect the reset (0) then one new "
+                + "increment (1)", Long.valueOf(1L), lastCountPerKey.get("App1_ResourceA"));
+        junit.framework.Assert.assertEquals("A different resource under the SAME app must be untouched",
+                Long.valueOf(3L), lastCountPerKey.get("App1_ResourceB"));
+        junit.framework.Assert.assertEquals("A different app entirely must be untouched",
+                Long.valueOf(3L), lastCountPerKey.get("App2_ResourceA"));
```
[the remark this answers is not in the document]
DakshithaS: @coderabbitai I think this is a false positive — inputHandler.send() dispatches synchronously here since there's no ``@Async`` annotation and no output-rate-limiting clause on this query, so QueryCallback.receive() completes before send() returns (verified against StreamJunction.sendEvent()'s actual dispatch logic). The Thread.sleep() pattern in this file's other tests is for a different reason — those use Siddhi's built-in window.timeBatch, which genuinely buffers events for real time, unlike the custom throttler:timeBatch used here, which forwards events immediately. Could you re-verify against the actual dispatch path rather than assuming async delivery?

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/test/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAggregatorTest.java
```diff
@@ -0,0 +1,1630 @@
...
+        while (System.currentTimeMillis() < deadline1) {
+            if (fakeClient.setWithExpiryCount.get() >= 1) {
+                ttlAfterFirstReset = fakeClient.lastSetWithExpiryTTL;
+                break;
+            }
+            Thread.sleep(50);
+        }
+        Assert.assertTrue("Expected the first PSETEX reset to have fired", ttlAfterFirstReset > 0);
+        Assert.assertTrue("First reset's TTL should be close to the 3000ms window duration — got: "
+                + ttlAfterFirstReset, ttlAfterFirstReset > 0 && ttlAfterFirstReset <= 3000L);
```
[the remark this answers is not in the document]
DakshithaS: @coderabbitai I think this is a false positive — I traced syncWithKVStore()'s branch order precisely: unsyncedCounter is drained via getAndSet(0) first, and processAdd() (called synchronously right after init(), same thread, microseconds before the first scheduler tick at t=10ms) has already set it to 1 by then. So the very first tick takes the INCRBY branch — not GET — and incrementBy() creates the key in the store immediately. From that point on, every subsequent GET-branch tick finds the key present, so the "absent-key seed → PSETEX" path you're describing is never reached at all before the real boundary reset. Could you re-verify against the actual unsyncedCounter/branch-selection order rather than assuming the key starts absent through the first 3200ms?

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
```diff
@@ -125,39 +132,116 @@ private void initializeFromKVStore() {
...
         synchronized (kvStoreLock) {
             if (kvStoreClient == null || key == null) {
                 return;
             }
+            if (pendingReset) {
+                try {
+                    writeCounterValue("0");
+                    // Do NOT reset localCounter here — reset() already zeroed it at window boundary.
+                    // Any increments to localCounter since then belong to the new window and are valid.
+                    pendingReset = false; // cleared only on success — failure retried next tick
```
AnuGayan: ### 🟠 Major — correctness / throttling accuracy (wrong 429s)

**Severity:** Major. Not a security vulnerability; it is a data-correctness defect that produces
incorrect throttle decisions (over-throttling — legitimate requests rejected with 429) for up to
one full window.

`writeCounterValue()` (L149-157) silently returns **without writing** when `storedWindowExpiry <= 0`
or has already passed. This branch then treats that no-op as success:

```java
writeCounterValue("0");
pendingReset = false;   // <-- cleared even though nothing was written
keyHasTTL = true;       // <-- asserts a TTL that may not exist
```

Result: Redis keeps the **old window's total** while `localCounter` is 0. The next sync tick pushes
the new window's delta with `INCRBY` on top of that stale total, so the shared counter is inflated.
Setting `keyHasTTL = true` additionally suppresses the compensating `PEXPIRE` in the INCRBY branch
(L212-228), so the key can be left with a stale value *and* no expiry.

**How to analyze**

1. Reachability of `storedWindowExpiry <= now` at flush time: `reset()` stores the expiry published
   by `ThrottleStreamProcessor`, which advances `expireEventTime` by only **one**
   `timeInMilliSeconds` per `process()` call. If `process()` runs two or more windows late (long GC
   pause, node freeze, redeploy, or a very short window under load), the RESET is emitted with an
   expiry that is already in the past. Confirm by logging
   `storedWindowExpiry - System.currentTimeMillis()` at the top of the `pendingReset` branch.
2. Reproduce deterministically in `DistributedCountAggregatorTest` with the existing fake client:
   set `storedWindowExpiry` via reflection to `System.currentTimeMillis() - 1000`, set
   `pendingReset = true`, call `currentState()` (which calls `syncWithKVStore()`), then assert
   `fakeClient.setWithExpiryCount == 0` **and** that `pendingReset` is still `true`. Today the flag
   is cleared, which is the bug.
3. On the residual-risk question raised earlier in this PR ("self-healing, because the TTL already
   removed the key"): that argument only holds when the key actually carried a TTL. If the key was
   created by `INCRBY` and the follow-up `PEXPIRE` had failed, `keyHasTTL` was `false` — and this
   branch flips it to `true`, so nothing re-stamps it. Verify with
   `FailingOnIncrByKeyValueStoreClient`-style fake that fails only `expireMillis`.

**Probable solution** — make the write report whether it happened, and only then clear the flags:

```diff
-    private void writeCounterValue(String value) {
+    private boolean writeCounterValue(String value) {
         long windowExpiry = storedWindowExpiry;
-        if (windowExpiry > 0) {
-            long remainingMillis = windowExpiry - System.currentTimeMillis();
-            if (remainingMillis > 0) {
-                kvStoreClient.setWithExpiry(key, value, remainingMillis);
-            }
-        }
+        if (windowExpiry <= 0) {
+            return false;
+        }
+        long remainingMillis = windowExpiry - System.currentTimeMillis();
+        if (remainingMillis <= 0) {
+            return false;
+        }
+        kvStoreClient.setWithExpiry(key, value, remainingMillis);
+        return true;
     }
```

```diff
             if (pendingReset) {
                 try {
-                    writeCounterValue("0");
-                    pendingReset = false;
-                    keyHasTTL = true;
+                    if (writeCounterValue("0")) {
+                        pendingReset = false; // cleared only on a real write
+                        keyHasTTL = true;
+                    } else {
+                        // Expiry unknown or already elapsed: force a TTL re-stamp and retry
+                        // on the next tick / next reset() rather than dropping the reset.
+                        keyHasTTL = false;
+                    }
                 } catch (KeyValueStoreException e) {
```

**Prompt for an AI agent**

```
In carbon-apimgt, file:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java

Bug (Major, correctness — causes over-throttling / spurious HTTP 429 for up to one throttle
window): writeCounterValue(String) silently does nothing when storedWindowExpiry is 0 or already
in the past, but the pendingReset branch of syncWithKVStore() treats that no-op as a successful
reset: it sets pendingReset = false and keyHasTTL = true. Redis therefore keeps the previous
window's counter total while localCounter is 0, the next INCRBY stacks the new window's delta on
that stale total, and the wrongly-set keyHasTTL suppresses the compensating PEXPIRE in the
INCRBY branch.

Fix:
1. Change writeCounterValue to return boolean: false when storedWindowExpiry <= 0 or when the
   remaining TTL is <= 0; true only when kvStoreClient.setWithExpiry actually executed.
2. In the pendingReset branch of syncWithKVStore(), clear pendingReset and set keyHasTTL = true
   only when writeCounterValue("0") returned true. When it returns false, leave pendingReset set
   so the next tick retries, and set keyHasTTL = false so a PEXPIRE is re-attempted.
3. Update the other two call sites (initializeFromKVStore and the restoreState seeding path) for
   the new return type without changing their behaviour.
4. Add JUnit tests in
   src/test/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAggregatorTest.java
   using the existing in-memory fake client and the existing reflection helpers:
   - storedWindowExpiry set to a past timestamp + pendingReset = true -> assert setWithExpiry was
     never called AND pendingReset is still true after a sync.
   - storedWindowExpiry set to a future timestamp -> assert setWithExpiry called once and
     pendingReset cleared.
Keep the existing code style (commons-logging, guarded string concatenation, no new dependencies).
```

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
```diff
@@ -219,31 +322,32 @@ public Object processRemove(Object data) {
...
+            if (expiry != null) {
+                storedWindowExpiry = expiry;
             }
-            return 0L;
-
-        } catch (KeyValueStoreException e) {
-            log.error("Error resetting counter for key " + key, e);
-            return 0L;
+            unsyncedCounter.set(0L);
+            pendingReset = true;
```
AnuGayan, line 359: ### 🟡 Minor→Major (deployment-dependent) — distributed correctness under clock skew

**Severity:** Minor in a single-region cluster with tight NTP sync; Major in multi-region /
active-active. Effect is **under-throttling** (requests allowed past the limit), i.e. a quota
enforcement gap rather than a memory-safety issue.

Every Traffic Manager sets `pendingReset` at *its own* window boundary and each one issues
`PSETEX <key> 0` against the **shared** key. `timeBatch(unitTime, 0)` aligns boundaries to epoch
multiples, so nodes agree modulo clock skew — but with skew a late node zeroes counts its peers
have already accrued in the new window, and its `remainingMillis` is computed from its own (earlier)
expiry, shortening the TTL for everyone.

**How to analyze**

1. Confirm boundary alignment: `ThrottleStreamProcessor.addTimeShift()` is only used on the first
   initialization; afterwards it is `expireEventTime += timeInMilliSeconds`. So alignment holds only
   while every node's timer keeps up.
2. Two-node reproduction: run two aggregator instances against one Redis, offset the second node's
   `storedWindowExpiry` / clock by 200 ms, drive traffic continuously across a boundary, and compare
   the total `INCRBY` count issued against the value observable in Redis right after the boundary.
   The delta is the number of silently dropped requests.
3. Quantify: worst case loss ≈ (skew in ms / sync interval in ms) × per-node request rate. With the
   default 10 ms sync interval, a 100 ms skew loses roughly 10 sync ticks' worth of counts.

**Probable solution** — make the counter key window-scoped, which removes the reset entirely:

```java
// key becomes: "wso2_throttler:" + throttleKey + ":" + storedWindowExpiry
private String keyForWindow(long windowExpiry) {
    return keyPrefix + ":" + windowExpiry;
}
```

Each window gets its own key, so `INCRBY` is naturally idempotent across nodes, no node can zero
another node's in-window counts, and the existing window TTL handles cleanup. `pendingReset`,
`writeCounterValue`, `keyHasTTL` and all the `!pendingReset` double-check guards can then be
deleted. If the key shape has to stay stable for backwards compatibility, the fallback is to
suppress the reset entirely and rely purely on the TTL, which you already set.

**Prompt for an AI agent**

```
In carbon-apimgt, files:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java

Problem (distributed correctness, under-throttling): in a multi-node Traffic Manager deployment,
every node issues PSETEX <sharedKey> 0 at its own window boundary. Under clock skew between nodes,
a late node's reset zeroes counts that peer nodes already accumulated in the NEW window, so the
shared counter under-reports and requests are allowed past the configured quota.

Refactor to window-scoped Redis keys so the reset becomes unnecessary:
1. Keep the existing throttle-key prefix ("wso2_throttler:" + throttleKey) in a field, and derive
   the effective Redis key per window as prefix + ":" + storedWindowExpiry.
2. When storedWindowExpiry changes (in reset() and in the processAdd() staleness refresh),
   recompute the effective key, and reset unsyncedCounter/localCounter bookkeeping for the new
   window locally without any Redis write.
3. Because each window now has a distinct key that starts absent, INCRBY alone is correct and
   idempotent across nodes. Remove pendingReset, writeCounterValue, keyHasTTL and every
   "if (!pendingReset)" double-check guard that exists only to serialise the reset.
4. Keep the TTL behaviour: after the INCRBY that creates the key, stamp PEXPIRE with the remaining
   window time so keys self-expire; retry the PEXPIRE on the next sync tick if it fails.
5. ACTIVE_AGGREGATORS is keyed by the Redis key today — key it by the stable throttle-key prefix
   instead so an aggregator is not registered twice across windows.
6. Update DistributedCountAggregatorTest accordingly and add a test asserting that two aggregator
   instances sharing one fake KV store, with different storedWindowExpiry values, never overwrite
   each other's counters.
Preserve public behaviour of processAdd/processRemove/reset return values.
```

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
```diff
@@ -125,39 +132,116 @@ private void initializeFromKVStore() {
                 long initialValue = Long.parseLong(kvStoreValue);
                 localCounter.set(initialValue);
             } else {
-                kvStoreClient.set(key, "0");
+                writeCounterValue("0");
```
AnuGayan, line 140: ### 🟡 Minor — dead code / silent behaviour regression

**Severity:** Minor. No incorrect throttle decision results (the key is created lazily by `INCRBY`),
but the code reads as if it seeds Redis and it never does — which will mislead the next maintainer
and hid part of the previous finding.

At this point `storedWindowExpiry` is still `0L` (a freshly cloned aggregator has never seen an
event), so `writeCounterValue("0")` short-circuits and writes nothing. Previously this branch did
`kvStoreClient.set(key, "0")` and really did seed the key. The same applies to the `seedRedis` path
in `restoreState()` (L439-446), making its flag and its `try/catch` inert.

**How to analyze**

1. Add a counter to the fake KV client in `DistributedCountAggregatorTest` and assert
   `setWithExpiryCount == 0` after `init()` on an absent key — it will pass, proving the seed never
   happens. (This is the same reasoning CodeRabbit reached when it withdrew a related comment.)
2. Confirm ordering: `init()` is reached from
   `GroupByAggregationAttributeExecutor.execute -> AttributeAggregator.cloneAggregator ->
   initAggregator -> init`, which runs **inside** `nextProcessor.process()`. So
   `ThrottleStreamProcessor.getThreadLocalWindowExpiry()` is already populated there — the expiry is
   available, it is just not read.

**Probable solution** — seed `storedWindowExpiry` in `init()` so both paths become live:

```diff
         if (distributedThrottlingEnabled && throttleKey != null) {
             this.key = "wso2_throttler:" + throttleKey;
+            Long windowExpiry = ThrottleStreamProcessor.getThreadLocalWindowExpiry();
+            if (windowExpiry != null) {
+                this.storedWindowExpiry = windowExpiry;
+            }
             try {
                 this.kvStoreClient = KeyValueStoreManager.getClient();
```

Alternatively, delete both seeding branches and add a comment stating that the key is created
lazily by the first `INCRBY` and stamped with `PEXPIRE`. Either is fine — silently doing nothing is
not.

**Prompt for an AI agent**

```
In carbon-apimgt, file:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java

Problem (Minor, dead code): the "key absent -> seed with 0" branches in initializeFromKVStore()
and in restoreState() call writeCounterValue("0"), but at both call sites storedWindowExpiry is
still 0L, so writeCounterValue short-circuits and no Redis write ever happens. This silently
regressed the previous kvStoreClient.set(key, "0") behaviour and makes restoreState's seedRedis
flag and try/catch inert.

Fix (option A, preferred): in init(), after computing this.key, read
ThrottleStreamProcessor.getThreadLocalWindowExpiry() and assign it to storedWindowExpiry when
non-null. init() is invoked from GroupByAggregationAttributeExecutor.execute inside
nextProcessor.process(), so the ThreadLocal is populated. Then both seeding paths work as written.

Fix (option B): delete both seeding branches plus the seedRedis flag, and add a comment explaining
that the Redis key is created lazily by the first INCRBY and given a TTL via PEXPIRE.

Whichever option is chosen, add a JUnit test in DistributedCountAggregatorTest that starts from an
absent key and asserts the observable outcome (either exactly one setWithExpiry call during init,
or zero writes plus a documented lazy-creation path).
```

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
```diff
@@ -290,21 +403,56 @@ public Object[] currentState() {
...
+            } catch (KeyValueStoreException | NumberFormatException e) {
+                log.error("Error reading from key-value store during restoreState for key "
+                        + key + ". Starting fresh at 0.", e);
+            }
+        }
+
+        // Lock only for the fast in-memory writes — no Redis call inside.
+        synchronized (kvStoreLock) {
+            unsyncedCounter.set(0L);
+            pendingReset = false;
```
AnuGayan: ### 🟡 Minor — lost reset across a Siddhi snapshot restore

**Severity:** Minor (narrow race window, requires a snapshot restore to land between `reset()` and
the next sync tick). Impact when hit is over-throttling for one window.

`pendingReset = false` is unconditional. If a restore happens after `reset()` flagged a pending
`PSETEX 0` but before the sync thread flushed it, the reset is dropped and `localCounter` is seeded
from a Redis value that still holds the **previous** window's total.

**How to analyze**

1. Set `pendingReset = true` by reflection, prime the fake KV client with a non-zero value, call
   `restoreState(new Object[]{new AbstractMap.SimpleEntry<String, Object>("Value", 0L)})`, then run a
   sync and assert whether `setWithExpiry` was ever invoked. It is not.
2. Establish how often this can occur in production by checking who calls `restoreState` —
   `GroupByAggregationAttributeExecutor.restoreState` during persistence restore. It is rare, which
   is why this is Minor rather than Major.

**Probable solution** — carry the pending reset across the restore instead of dropping it:

```diff
         synchronized (kvStoreLock) {
             unsyncedCounter.set(0L);
-            pendingReset = false;
+            // Keep an unflushed window-boundary reset pending; dropping it would leave the
+            // previous window's total in Redis.
+            if (!pendingReset) {
+                localCounter.set(counterToRestore);
+            } else {
+                localCounter.set(0L);
+            }
-            localCounter.set(counterToRestore);
         }
```

**Prompt for an AI agent**

```
In carbon-apimgt, file:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java

Problem (Minor, correctness race): restoreState() sets pendingReset = false unconditionally. If a
Siddhi snapshot restore lands between reset() flagging a pending PSETEX 0 and the background sync
thread flushing it, the reset is silently dropped and localCounter is seeded from a Redis value
that still contains the previous window's total, causing over-throttling for one window.

Fix: inside the existing synchronized (kvStoreLock) block in restoreState(), do not clear
pendingReset. When pendingReset is true, set localCounter to 0 and leave the flag set so the sync
thread still issues the PSETEX 0. When it is false, keep the current behaviour of setting
localCounter to counterToRestore. Leave the Redis GET outside the lock as it is today.

Add a JUnit test in DistributedCountAggregatorTest: set pendingReset = true by reflection, prime
the fake KV client with a non-zero value, call restoreState with a snapshot entry, run one sync,
and assert setWithExpiry("0") was issued and localCounter is 0.
```

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java
```diff
@@ -162,22 +178,33 @@ protected void process(ComplexEventChunk<StreamEvent> streamEventChunk, Processo
...
+                    StreamEvent clonedStreamEvent = streamEventCloner.copyStreamEvent(streamEvent);
+                    clonedStreamEvent.setType(StreamEvent.Type.EXPIRED);
+                    clonedStreamEvent.setTimestamp(expireEventTime);
+                    expiredEventChunk.add(clonedStreamEvent);
                 }
-                expiredEventChunk.clear();
             }
+            expireEventTimeSnapshot = expireEventTime;
+        }
+        if (resetEventChunk != null) {
```
AnuGayan, line 190: ### 🟠 Major — concurrency: RESET dispatched outside the lock loses ordering vs. concurrent chunks

**Severity:** Major. Race condition producing **under-throttling** (quota bypass) — a correctness
and quota-enforcement gap, not memory corruption.

Before this PR the expired events were appended to the *same* `streamEventChunk` and dispatched in
one `nextProcessor.process()` call, so "reset the window" and "count the new window's events"
reached `QuerySelector` in a guaranteed order. Now they are two separate dispatches, both made
**after** `synchronized (this)` is released:

* Thread A crosses the boundary, builds `resetEventChunk`, exits the lock.
* Thread B (same window instance, chunk that does *not* cross the boundary) enters, exits, and
  dispatches its CURRENT chunk **first**.
* A then dispatches the RESET, and `DistributedCountAttributeAggregator.reset()` executes
  `localCounter.set(0)` + `unsyncedCounter.set(0)`, discarding B's already-counted new-window events.

This is strictly worse than the old code, because the reset is destructive (set to 0) rather than N
individually ordered decrements.

**How to analyze**

1. Confirm concurrency is possible: `ThrottleStreamProcessor.process()` guards only its own state
   with `synchronized (this)`; both `nextProcessor.process()` calls are outside it, and the scheduler
   timer thread and the event-flow thread(s) both enter `process()`.
2. Deterministic reproduction: subclass or instrument `nextProcessor` so that the dispatch of
   `resetEventChunk` blocks on a latch; from a second thread send a normal event that does not cross
   the boundary; release the latch; then assert the counter reflects the second thread's event. It
   will read 0.
3. Cross-check the emission order assumption in siddhi-core 3.3.0
   `QuerySelector.processInBatchGroupBy` — `case RESET` is handled per chunk, so ordering is purely
   a function of dispatch order here.

**Probable solution** — restore the single, ordered dispatch by putting the RESET at the head of the
same chunk (preferred), or dispatch the reset chunk while still holding the lock:

```diff
-        if (resetEventChunk != null) {
-            resetEventChunk.setBatch(true);
-            windowExpiryThreadLocal.set(expireEventTimeSnapshot);
-            try {
-                nextProcessor.process(resetEventChunk);
-            } finally {
-                windowExpiryThreadLocal.remove();
-            }
-            resetEventChunk.setBatch(false);
-        }
-        if (streamEventChunk.getFirst() != null) {
-            streamEventChunk.setBatch(true);
+        // RESET must be observed by the selector before this chunk's CURRENT events, otherwise a
+        // concurrent process() can have its new-window increments zeroed by our reset.
+        if (resetEventChunk != null) {
+            if (streamEventChunk.getFirst() != null) {
+                resetEventChunk.add(streamEventChunk.getFirst());
+            }
+            streamEventChunk = resetEventChunk;
+        }
+        if (streamEventChunk.getFirst() != null) {
+            streamEventChunk.setBatch(true);
             windowExpiryThreadLocal.set(expireEventTimeSnapshot);
```

(Verify with the existing `ThrottleTimeBatchWindowTestCase` group-by reset tests, which must keep
passing — RESET first, then the CURRENT events of the boundary-crossing chunk.)

**Prompt for an AI agent**

```
In carbon-apimgt, file:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java

Bug (Major, concurrency / under-throttling): process() now performs two separate
nextProcessor.process() dispatches - one for resetEventChunk and one for streamEventChunk - and
both happen after the synchronized (this) block is released. A concurrent process() invocation on
another thread can therefore dispatch its CURRENT chunk before our RESET chunk is dispatched, and
the aggregator's reset() (which sets localCounter and unsyncedCounter to 0) then discards those
already-counted new-window events. The pre-PR code avoided this by appending the expired events to
the same chunk and dispatching once.

Fix: restore a single ordered dispatch. When resetEventChunk is non-null, append the current
streamEventChunk's first event chain to the reset chunk (so the RESET event is the head of the
chunk) and dispatch only that combined chunk; otherwise dispatch streamEventChunk as today. Keep
setBatch(true)/setBatch(false) and keep the windowExpiryThreadLocal set/remove in a try/finally
around the single dispatch. Do not move Redis or blocking work inside synchronized (this).

Verify: all existing tests in
src/test/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleTimeBatchWindowTestCase.java
and CountAttributeAggregatorWithResetTestCase.java must still pass (RESET must be processed before
the CURRENT events of the boundary-crossing chunk). Add a test that dispatches a boundary-crossing
chunk concurrently with a non-boundary chunk and asserts no counted event is lost.
```

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java
```diff
@@ -146,13 +150,25 @@ protected void process(ComplexEventChunk<StreamEvent> streamEventChunk, Processo
...
+            if (sendResetEvent) {
+                StreamEvent firstExpiredEvent = expiredEventChunk.getFirst();
+                if (firstExpiredEvent != null) {
+                    StreamEvent resetEvent = streamEventCloner.copyStreamEvent(firstExpiredEvent);
+                    resetEvent.setType(StreamEvent.Type.RESET);
+                    resetEvent.setTimestamp(expireEventTime);
+                    resetEventChunk = new ComplexEventChunk<StreamEvent>(true);
+                    resetEventChunk.add(resetEvent);
+                }
+                expiredEventChunk.clear();
```
AnuGayan, line 171: ### 🟠 Major — undocumented change to the `GlobalThrottleStream` contract (no boundary output event)

**Severity:** Major as a **behaviour/contract change**; low operational risk for the shipped APIM
gateway (see analysis step 3). Not a security issue.

In siddhi-core 3.3.0 `QuerySelector`, a RESET event runs the attribute processors and `break`s —
nothing is added to `groupedEvents`/`currentComplexEventChunk`, so **no output event is emitted**.
Previously, clearing the window emitted one EXPIRED event per group key with `count == 0`, which
flowed through `throttler:emitOnStateChange(throttleKey, isThrottled)` as
`isThrottled = false` and published the **un-throttle** event to `GlobalThrottleStream`. After this
change nothing is published at the boundary; release is deferred to the first request of the next
window, or never if traffic stops.

**How to analyze**

1. Read `QuerySelector.processInBatchGroupBy` / `processGroupBy` in siddhi-core 3.3.0: `case RESET:`
   only iterates `attributeProcessorList`; only CURRENT/EXPIRED events reach `groupedEvents`.
2. Confirm the old path emitted: `ComplexEventChunk.add()` sets `last = getLastEvent(chunk)`, so the
   pre-PR `streamEventChunk.add(expiredEventChunk.getFirst())` appended the **whole** EXPIRED chain,
   and `INSERT ALL EVENTS` means `expiredOn == true`, so those events were emitted with `count == 0`.
3. Assess blast radius: `ThrottleDataHolder.isThrottled(key)` / `isAPIThrottled(key)` in
   `org.wso2.carbon.apimgt.gateway` drop the entry once `timestamp < currentTimeMillis`, and
   `EmitOnStateChange` never suppresses `isThrottled == true` (it only suppresses `false` after
   `false`). So the gateway self-heals via `expiryTimeStamp` and there is no over- or
   under-throttling. Any *other* consumer of `GlobalThrottleStream`, or a custom execution plan that
   relies on the boundary flush, does break.
4. Note that the new test `throttleTimeWindowBatchShouldResetAggregatesOnWindowExpiry` only encodes
   this implicitly via `currentEvents.size() <= 2`; nothing asserts it deliberately.

**Probable solution** — no code change required if this is intentional, but please:

1. Add it to the PR description as an explicit behaviour change ("the window boundary no longer
   emits a `count == 0` / `isThrottled == false` event; un-throttling at the gateway relies on
   `expiryTimeStamp`").
2. Add a comment here, e.g. `// RESET produces no downstream output in QuerySelector; the previous
   count==0 flush event is intentionally gone.`
3. Add an explicit regression test asserting the boundary emits **zero** output events and that the
   first event of the next window reports `count == 1`.

**Prompt for an AI agent**

```
In carbon-apimgt, file:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java

Behaviour change to document and pin with a test (Major - contract change, not a crash): emitting a
single RESET event at the window boundary instead of the previous chain of EXPIRED events means
siddhi-core 3.3.0 QuerySelector emits NO output event at the boundary (case RESET only runs the
attribute processors). Previously each group key produced an output event with count == 0, which
throttler:emitOnStateChange turned into an isThrottled=false "un-throttle" event on
GlobalThrottleStream. That publish is now gone; the gateway relies on expiryTimeStamp to release.

Tasks:
1. Add a short comment at the reset-emission site in process() stating that RESET yields no
   downstream output and that the previous count==0 flush event is intentionally removed.
2. Add a JUnit test in
   src/test/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleTimeBatchWindowTestCase.java
   with a group-by query and INSERT ALL EVENTS that: sends N events, waits past the window
   boundary WITHOUT sending anything, and asserts the QueryCallback received no additional event at
   the boundary; then sends one event in the new window and asserts the reported count is 1.
3. Do not change runtime behaviour.
```

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java
```diff
@@ -162,22 +178,33 @@ protected void process(ComplexEventChunk<StreamEvent> streamEventChunk, Processo
...
-                StreamEvent clonedStreamEvent = streamEventCloner.copyStreamEvent(streamEvent);
-                clonedStreamEvent.setType(StreamEvent.Type.EXPIRED);
-                clonedStreamEvent.setTimestamp(expireEventTime);
-                expiredEventChunk.add(clonedStreamEvent);
-            }
-            if (sendEvents) {
-                expiredEventChunk.reset();
-                if (expiredEventChunk.getFirst() != null) {
-                    streamEventChunk.add(expiredEventChunk.getFirst());
+                if (expiredEventChunk.getFirst() == null) {
```
AnuGayan, line 181: ### 🟡 Minor — `expiredEventChunk` now holds at most one event; `FindableProcessor` semantics changed silently

**Severity:** Minor. No shipped policy template is affected; the risk is to custom execution plans.

`expiredEventChunk` used to accumulate one EXPIRED clone per event in the window. With this guard it
holds **at most one** event — a placeholder used solely as the template for the RESET clone. But the
same chunk backs `find()` (L234-236) and `constructFinder()` (L238-244), i.e. this class's
`FindableProcessor` contract, and it is what `currentState()` snapshots.

**How to analyze**

1. Grep the shipped plans for joins against the window:
   `grep -rn "throttler:timeBatch" features/apimgt/*/src/main/resources/{policy_templates,conf/executionplans}`
   — every occurrence is `FROM <stream>#throttler:timeBatch(...) SELECT ... GROUP BY ... INSERT ALL
   EVENTS`, with no `JOIN`. So nothing shipped regresses.
2. Any custom plan that joins against this window (or uses it as the right-hand side of a `JOIN`)
   will now match against a single row instead of the window contents — silently, with no error.

**Probable solution** — document the narrowed contract so this is a deliberate, discoverable
decision:

```diff
-    private ComplexEventChunk<StreamEvent> expiredEventChunk = new ComplexEventChunk<StreamEvent>(true);
+    /**
+     * Holds at most ONE event per window: the placeholder cloned into the boundary RESET event.
+     * It is deliberately no longer the full window contents, so find()/constructFinder() (the
+     * FindableProcessor contract) will match against that single placeholder only. Shipped throttle
+     * policies never join against this window; custom plans that do must not rely on it.
+     */
+    private ComplexEventChunk<StreamEvent> expiredEventChunk = new ComplexEventChunk<StreamEvent>(true);
```

If joining against the window must keep working, keep accumulating every event and use only
`getFirst()` for the RESET template — at the cost of the memory the old code used.

**Prompt for an AI agent**

```
In carbon-apimgt, file:
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java

Issue (Minor, silent contract narrowing): the new "if (expiredEventChunk.getFirst() == null)" guard
means expiredEventChunk now holds at most one event instead of every event of the window. That same
chunk backs find() and constructFinder(), i.e. this class's FindableProcessor contract, and is what
currentState() snapshots. Shipped throttle policy templates never join against this window, so
nothing regresses today, but a custom execution plan that joins against it would silently match a
single row.

Task: add a Javadoc comment on the expiredEventChunk field documenting that it now holds only the
placeholder event cloned into the boundary RESET event, that it is no longer the window contents,
and the consequence for find()/constructFinder(). Also confirm restoreState() tolerates a null
state[0] (currentState() can return null when the window saw no events) and guard it if it does not.
Do not change runtime behaviour.
```

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/DistributedThrottleConfig.java
```diff
@@ -47,6 +47,8 @@ public class DistributedThrottleConfig {
     private int numTestsPerEvictionRun = -1;
     private boolean isSslEnabled;
     private int connectionTimeout;
+    private int socketTimeout;
+    private long maxWaitMillis = 3000L;
```
AnuGayan, line 51: ### 🟠 Major — default contradicts the PR description and diverges from `RedisConfig`

**Severity:** Major as a silent operational behaviour change on upgrade (a slow Redis now fails fast
instead of blocking); Minor as a code defect. Not a security issue.

`DistributedThrottleConfig.maxWaitMillis` defaults to **`3000L`**, while `RedisConfig.maxWaitMillis`
(the same logical setting, `RedisConfig.java:35`) defaults to **`-1`**. Since
`JedisKeyValueStoreClient` now always calls `poolConfig.setMaxWaitMillis(...)` for the distributed
path, an **unconfigured** deployment changes from "block until a pool connection frees"
(`JedisPoolConfig` default `-1`) to "throw `JedisExhaustedPoolException` after 3 s". The PR body
states the fallback is `-1` and that the change is "preserving backward compatibility" — the code
does not match that.

The same applies to `connectionTimeout`: `getConnectionTimeout() != 0 ? … : Protocol.DEFAULT_TIMEOUT`
moves the unconfigured case from an infinite timeout to 2000 ms.

**How to analyze**

1. Compare defaults side by side: `DistributedThrottleConfig.java:51` (`3000L`) vs
   `RedisConfig.java:35` (`-1`), then the two call sites in `JedisKeyValueStoreClient`
   (L80 and L109) — the distributed branch always applies its default, the Redis branch keeps `-1`.
2. Confirm the runtime effect: with `blockWhenExhausted = true` and `maxWaitMillis = 3000`,
   `pool.getResource()` raises `JedisException` after 3 s; `getJedis()` returns `null`, callers throw
   `KeyValueStoreException`, and `syncWithKVStore()` takes its error path and restores the delta —
   i.e. counters stall rather than block. Decide whether that is the intended upgrade behaviour.
3. Note that `!= 0` lets a negative configured value through unchanged for both timeouts.

**Probable solution** — pick one and make it explicit:

```diff
-    private long maxWaitMillis = 3000L;
+    // Matches RedisConfig.maxWaitMillis: -1 = block until a connection is available (JedisPoolConfig default).
+    private long maxWaitMillis = -1L;
```

If the bounded 3 s wait is deliberate (it is a reasonable choice for a throttling hot path), then
keep `3000L`, align `RedisConfig` to the same default, correct the PR description, and document the
new `JedisExhaustedPoolException` failure mode in the config docs. Also consider `<= 0` instead of
`!= 0` in the `JedisKeyValueStoreClient` timeout fallbacks so negative values fall back to the
default rather than reaching Jedis.

**Prompt for an AI agent**

```
In carbon-apimgt, files:
components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/DistributedThrottleConfig.java
components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/RedisConfig.java
components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/util/kvstore/JedisKeyValueStoreClient.java

Issue (Major, silent behaviour change on upgrade): DistributedThrottleConfig.maxWaitMillis defaults
to 3000L while RedisConfig.maxWaitMillis - the same logical setting - defaults to -1. Because
JedisKeyValueStoreClient now always calls poolConfig.setMaxWaitMillis() on the distributed-throttle
path, an unconfigured deployment silently moves from "block until a pool connection is available"
(JedisPoolConfig default -1) to "throw JedisExhaustedPoolException after 3s". The PR description
claims the fallback is -1 and is backward compatible.

Tasks:
1. Make the two DTO defaults consistent. Either set DistributedThrottleConfig.maxWaitMillis to -1L
   (fully backward compatible), or keep 3000L deliberately AND set the same default in RedisConfig.
   Add a short comment on the field stating the chosen semantics.
2. In JedisKeyValueStoreClient.populateKeyValueStoreConfigs(), change the connectionTimeout and
   socketTimeout fallbacks from "!= 0 ? value : Protocol.DEFAULT_TIMEOUT" to "> 0 ? value :
   Protocol.DEFAULT_TIMEOUT" so negative configured values fall back to the default instead of
   reaching Jedis.
3. Update the PR description / config documentation to state the actual defaults and the new
   fail-fast failure mode (JedisExhaustedPoolException surfacing as KeyValueStoreException, which
   makes syncWithKVStore take its error path and retain the unsynced delta).
Do not add new configuration keys.
```

AnuGayan (commented):
## Review — distributed throttle counter reset + Redis pool config

**Scope reviewed:** full diff at `1682f595`, read in context against siddhi-core **3.3.0**
(`QuerySelector`, `GroupByAggregationAttributeExecutor`, `ComplexEventChunk`), jedis **3.3.0**, the
shipped policy templates (`throttle_policy_template_{app,sub,resource}.xml`) and the gateway's
`ThrottleDataHolder`.

**Assessment.** The diagnosis is right and the direction is the correct one: replacing N sequential
`processRemove()` decrements with a single broadcast RESET plus an atomic `PSETEX key 0 <ttl>` off
the Siddhi event thread is the proper fix for the intermediate-state visibility problem. I confirmed
the central assumption independently — `GroupByAggregationAttributeExecutor.execute()` in
siddhi-core 3.3.0 does iterate the entire `aggregatorMap` on a RESET event without consulting the
group-by key, so one RESET event legitimately resets every active throttle key. Test investment is
unusually strong (36 new tests with hand-rolled failing/blocking KV clients).

**Verdict: request changes** (posted as a comment rather than a blocking review). Three items should
be addressed before merge:

1. 🟠 `writeCounterValue()` no-op is treated as a successful reset — a dropped reset leaves the old
   window's total in Redis → spurious 429s for up to a window. *(`DistributedCountAttributeAggregator.java:175`)*
2. 🟠 The RESET chunk is dispatched **outside** `synchronized (this)`, so a concurrent chunk can be
   counted before the reset arrives and then be zeroed → quota bypass. *(`ThrottleStreamProcessor.java:190`)*
3. 🟠 The window boundary no longer emits any output event, so the `isThrottled = false` un-throttle
   publish to `GlobalThrottleStream` is gone. The APIM gateway self-heals via `expiryTimeStamp`, so
   there is no functional break — but it is an undocumented change to a published stream contract
   and no test asserts it. *(`ThrottleStreamProcessor.java:171`)*

Plus 🟠 an `maxWaitMillis` default that contradicts the PR description, and four 🟡 items. Each
inline comment carries **severity**, **how to analyze/reproduce**, a **probable solution** as a diff,
and a **copy-paste prompt for an AI agent**.

### Nits (not inline)

- `DistributedCountAttributeAggregator.java:406` — `log.debug("… restoreState called with state: " + state)`
  prints `[Ljava.lang.Object;@1b6d…`; use `Arrays.toString(state)`.
- `DistributedCountAttributeAggregator.java:374-377` — `stop()` calling `syncWithKVStore()` twice
  with a `pendingReset` read outside `kvStoreLock` is hard to follow; a `flushPendingReset()` +
  `pushDelta()` split would read better.
- `DistributedCountAttributeAggregator.java:571` — `isResetRequested` uses no instance state; make it `static`.
- `JedisKeyValueStoreClient.java:280-320` — `setWithExpiry`/`expireMillis` don't validate
  `ttlMillis > 0`; Redis rejects a non-positive `PSETEX` TTL. Callers guard today, but the interface
  javadoc doesn't state the precondition.
- `ThrottleTimeBatchWindowTestCase` — unused `java.util.ArrayList` import.
- Timing-based tests (`Thread.sleep(3500)`, the 2 s/3 s boundary tests) will flake on loaded CI;
  aligning to the boundary before sending, as suggested elsewhere in this PR, is worth taking.
- `processRemove(Object)` / `processRemove(Object[])` are now effectively unreachable for the
  throttle plans since no EXPIRED events reach the selector — worth a one-line comment.

### What's good

- Redis I/O moved off the Siddhi event thread, with retry-until-success semantics.
- `isResetRequested()` finally wires the admin reset flag into the distributed aggregator. The app
  policy template already ships
  `throttler:distributedCount(messageID, cast(map:get(propertiesMap,'reset'),'bool'))`, and the old
  `processAdd((Object) data)` delegation just incremented — so admin-triggered app-counter resets
  were silently broken in distributed mode. The semantics now match
  `CountAttributeAggregatorWithReset` exactly. This is a real user-visible fix and deserves its own
  bullet in the PR description.
- Window TTLs mean a node crashing mid-window can no longer pin a counter forever.
- Scheduler lifecycle is genuinely better: named threads (`apim-distributed-throttle-sync`), the
  master scheduler created lazily and actually shut down, and `shutdownExecutor` removing duplication.
- The `getJedis()` CAS on `lastErrorLogTimestamp` fixes real log-storm double-writes.
- No DB, schema, or REST API surface touched — those review dimensions don't apply.

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java
DakshithaS: @AnuGayan This is the behavior change we discussed: emitStateChanges will no longer receive any output from the RESET event.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/dto/DistributedThrottleConfig.java
DakshithaS: As agreed, set maxWaitTime to 3000L.

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/DistributedCountAttributeAggregator.java
DakshithaS: Fixed

components/apimgt/org.wso2.carbon.apimgt.throttling.siddhi.extension/src/main/java/org/wso2/carbon/apimgt/throttling/siddhi/extension/ThrottleStreamProcessor.java
DakshithaS: Both nextProcessor.process() calls (reset and stream chunks) now execute inside synchronized(this) rather than after it releases. This closes not just the intra-call ordering issue but the actual cross-thread race described (scheduler timer vs. concurrent event thread) — merging into one chunk (as suggested) would only fix ordering within a single process() invocation, not between two independent ones.

DakshithaS:
@coderabbitai please review and approve

AnuGayan (approved):

PR #13917, opened 2026-07-10
Fix pagination boundary logic in getPaginatedSubscribedAPIsByApplication

PasanT9:
## Summary
- Fixes an off-by-one/boundary bug in `ApiMgtDAO.getPaginatedSubscribedAPIsByApplication` where the loop condition and break condition used `limit` instead of `limit + offset`, causing incorrect pagination of an application's subscriptions (e.g. `offset=2&limit=2` did not return the expected page).
- Reworks the loop to skip rows before `offset` and stop once `limit` items have been collected, which is simpler and correct regardless of `offset`.

## Source fix
https://github.com/wso2-support/carbon-apimgt/pull/8332

## Tracking issue
https://github.com/wso2/api-manager/issues/5082


RakhithaRR (dismissed):

PasanT9:
@coderabbitai seems like this is already approved. But I can't merge it yet as it says waiting for status to be reported. Can you update the status.

PasanT9:
@coderabbitai Its stuck on "CodeRabbitExpected — Waiting for status to be reported"

HeshanSudarshana (approved):

PR #13923, opened 2026-07-15
Fix NullPointerException when creating MCP Server from an existing OpenAPI v2  API

Shamly-Shanawaz:
### Problem
Creating an MCP Server from an existing REST API whose definition is Swagger 2.0 (OpenAPI v2) fails with a 500 error

### Root cause
In `PublisherCommonUtils.resolveExistingMCPBackendAPI`, the parser used to read the referenced backend API's definition was selected from the request's `openAPIVersion` query param (default `v3`) instead of the backend's actual definition version. A v3 parser reading a Swagger 2.0 document returns `null`, which then NPEs in
`generateMCPTools`.

### Fix
Select the parser from the referenced backend API's own definition using `OASParserUtil.getOASParser(...)`, so v2 and v3 backends are both parsed correctly regardless of the requested output version. The now-unused `oasParser` parameter is removed from the private method. This aligns the existing-API path with the direct-backend and import flows, which already auto-detect the version from the definition.

 **Related issue** - https://github.com/wso2/api-manager/issues/5127

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
@@ -2438,7 +2438,7 @@ public class PublisherCommonUtils {
                 if (APIConstants.API_SUBTYPE_EXISTING_API.equals(apiToAdd.getSubtype())
                         && !apiToAdd.getUriTemplates().isEmpty()) {
                     Set<URITemplate> updatedTemplates = resolveExistingMCPBackendAPI(apiToAdd, apiProvider,
-                            organization, oasParser);
+                            organization);
                     apiToAdd.setUriTemplates(updatedTemplates);
                 }
                 apiDefinition = new OAS3Parser().generateAPIDefinition(swaggerData);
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
```diff
@@ -290,6 +290,9 @@ public enum ExceptionCodes implements ErrorHandler {
     ERROR_MCP_TOOL_GENERATION_NOT_SUPPORTED(900767, "MCP tool generation not supported", 400,
             "MCP tool generation is not supported for Async API definitions"),
 
+    MCP_BACKEND_VERSION_MISMATCH(900769, "OpenAPI version mismatch", 400,
+            "The requested openAPIVersion '%s' does not match the referenced backend API definition version '%s'"),
```
PasanT9: Should be `OpenAPI version`.

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
```diff
@@ -2438,7 +2438,7 @@ public static API addAPIWithGeneratedSwaggerDefinition(API apiToAdd, String oasV
                 if (APIConstants.API_SUBTYPE_EXISTING_API.equals(apiToAdd.getSubtype())
                         && !apiToAdd.getUriTemplates().isEmpty()) {
                     Set<URITemplate> updatedTemplates = resolveExistingMCPBackendAPI(apiToAdd, apiProvider,
-                            organization, oasParser);
+                            organization);
```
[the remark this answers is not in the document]
Shamly-Shanawaz: The parser is now taken from the referenced backend API's own definition, not from the requested openAPIVersion. The backend's actual version is authoritative, so v2 and v3 apis are read correctly without the caller declaring a version, and there's no mismatch to validate.

components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
Shamly-Shanawaz: Altered the fix , hence ignoring .

PasanT9 (approved):

PasanT9 (approved):

PR #13927, opened 2026-07-17
Request/Response body support for Moesif analytics (gateway side)

vinupa:
## Purpose

- **Related issue:** https://github.com/wso2/api-manager/issues/5128
- **Companion PR (publisher mapping side, `apim-analytics-publisher`):** https://github.com/wso2/apim-analytics-publisher/pull/168

> This is one half of a **two-repo** change. This PR captures the body into the analytics
> event `properties`; the companion PR maps those properties onto the Moesif SDK event.

### Problem

The WSO2 APIM Moesif plugin is not able to capture request and response body/payload data in Moesif.

### Root cause (gateway side)

The gateway never captured the request/response payload. Headers reach Moesif precisely because
they are placed in `properties` (`requestHeaders`/`responseHeaders`); nothing similar existed for
the body.

### Intended outcome

When explicitly enabled, capture the request and response bodies and place them in the event
`properties` so the publisher can forward them to Moesif, without disrupting the proxied request/response.

### Functional goals

- Capture the request and response body (opt-in) into the event `properties`.
- Convey a per-body **transfer-encoding hint** and the body's **`Content-Type`** so the publisher
  can render it correctly independent of `send_headers`.
- **Opt-in, off by default**. Zero cost when disabled.

### Constraints & non-goals

- **Property-key contract:** the `properties` key strings must match the publisher
  (`apim-analytics-publisher`) exactly (see the table under *Approach*).
- WebSocket/async APIs do not capture bodies.
- **Bounded:**  size-capped; oversized bodies must not be buffered/re-serialized.
- **Memory-safe by default:** a body whose size cannot be bounded before building (no `Content-Length`,
  e.g. chunked) is skipped by default; capturing it is a separate, explicit opt-in.
- **Explainable:** when enabled but a body is dropped, the reason is observable.

### What changed

| File | Change |
|------|--------|
| `analytics/AnalyticsPayloadUtil.java` **(new)** | Single body-extraction utility for both request and response. |
| `analytics/AnalyticsMetricsHandler.java` | Capture the request body in `handleRequestOutFlow`, gated by `shouldCaptureRequestPayload()`. |
| `analytics/SynapseAnalyticsDataProvider.java` | Add request/response bodies (+ encodings + `requestContentType`) to event `properties`; capture the response body at event-collection time. |
| `analytics/Constants.java` | New constants (`send_payloads`, `payload_size_limit`, `capture_payloads_without_content_length`, body/encoding/content-type keys). |
| `pom.xml` | Bump `analytics.event.publisher.client.version` to the release carrying the publisher-side body support. |

### Capture rules: `AnalyticsPayloadUtil.extractPayload(...)`

| Case | Behavior |
|------|----------|
| No entity body (`NO_ENTITY_BODY`, e.g. GET/DELETE) | Skipped: no build, left byte-identical |
| `text/event-stream` (SSE), `multipart/*` | Excluded: no single materializable body |
| JSON | Captured verbatim, `transferEncoding=null` (publisher renders as structured object) |
| Text / XML / SOAP | Captured as string, `transferEncoding=null` |
| Binary | Base64-encoded, `transferEncoding="base64"` |
| Body larger than `payload_size_limit` | **Dropped**: publisher only ever gets a whole, valid body |
| No `Content-Length` (e.g. chunked) | **Skipped by default** (memory-safe); captured only when `capture_payloads_without_content_length=true` |

### Pre-build gate for Oversize handling:

```java
// If the payload advertises its size and it exceeds the limit, drop it now WITHOUT building —
// so an oversized body is never buffered into memory or re-serialized.
long declaredLength = getDeclaredContentLength(axis2MC);
if (declaredLength > sizeLimit) {
    return null;
}

// No declared Content-Length (e.g. chunked): the size cannot be bounded before building, so building
// here would materialize the whole body in memory (an OOM risk under load). Skip unless the operator
// has explicitly opted in. Applies symmetrically to request and response.
if (declaredLength < 0 && !shouldCapturePayloadsWithoutContentLength()) {
    return null;
}
RelayUtils.buildMessage(axis2MC);
```

**Why the no-`Content-Length` skip is default-off.** A body advertising a `Content-Length` over the
limit is dropped *before* building, so peak memory is bounded by `payload_size_limit`. A body with no
declared length (chunked transfer-encoding) can only be sized *after* `RelayUtils.buildMessage`
materializes it in full, so building a large chunked body purely to drop it is a real OOM/memory-pressure risk under load. So the safe choice is: **skip these bodies by default** (they are never built, and the
call still proxies normally via streaming), and expose an explicit opt-in for operators who need them
and have the headroom.

### Request-capture skip guards

`handleRequestOutFlow` gates capture on `shouldCaptureRequestPayload()`, mirroring the skips `handleResponseOutFlow` already applies, so no request is built/re-serialized for a call that will not publish an event:

| Condition | Skips capture |
|-----------|---------------|
| WebSocket subscriber path | ✅ |
| `send_payloads` set to **false** | ✅ |
| `SKIP_METRICS_PUBLISHING` set to **true** | ✅ |
| File-based API context | ✅ |

Additionally, `extractPayload` itself skips (without building) any body with no declared
`Content-Length` unless `capture_payloads_without_content_length=true`. So the default configuration
never materializes an unbounded chunked body, for either direction.

### Observability (Why is a body missing?)

When `send_payloads` is on but a body is skipped/dropped, the reason is logged at **debug** level, keyed by request/response. For example:

```
Dropping request body from analytics: declared Content-Length 250000 bytes exceeds
payload_size_limit of 100000; body not built. Increase payload_size_limit to capture it.

No response body captured for analytics: content type 'text/event-stream' is excluded
(streaming/multipart bodies are never captured).
```

### Configuration (`deployment.toml`)

```toml
[apim.analytics]
enable = true
type = "moesif"

[apim.analytics.properties]
moesifKey = "<Moesif Collector Application Id>"
send_payloads = "true"                            # opt-in; default false
payload_size_limit = "100000"                     # optional; default 100000 (bytes per captured body)
capture_payloads_without_content_length = "false" # optional; default false (memory-safe)
```

`payload_size_limit` is validated: a non-positive or non-numeric value falls back to the default
`100000` (with a warning) instead of silently dropping every body.

`capture_payloads_without_content_length` (default `false`) governs bodies that do not declare a
`Content-Length` (e.g. chunked transfer-encoding). When off, they are skipped (never built) so the default
configuration is memory-safe. Set it `true` to capture them, accepting that each such body is
materialized in full before it can be size-checked. Applies symmetrically to request and response.

### Design decisions & rationale

- **Where the request body is captured: `handleRequestOutFlow`.** The last safe point before the
  backend send; building there sets `MESSAGE_BUILDER_INVOKED` before the sender checks it, so the
  engine re-serializes the built message to the backend (forwarding preserved).
- **`Content-Type` conveyed independent of `send_headers`** so the publisher can label/parse the
  body even when header logging is off.
- **Memory safety is default-on; unbounded capture is a deliberate opt-in.** Bodies with a
  `Content-Length` are bounded by the pre-build gate; bodies without one are skipped by default
  because sizing them requires a full in-memory build. `capture_payloads_without_content_length`
  lets operators re-enable that capture where they have the headroom.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
@@ -65,7 +65,6 @@ public class AnalyticsMetricsHandler extends AbstractExtendedSynapseHandler {
 
     @Override
     public boolean handleRequestOutFlow(MessageContext messageContext) {
-        messageContext.setProperty(Constants.BACKEND_START_TIME_PROPERTY, System.currentTimeMillis());
         // Capture the request payload here (request-out flow), the last point before the backend send
         // where the request envelope is still current. Gated by the send_payloads analytics property.
         // Building here sets MESSAGE_BUILDER_INVOKED before the pass-through sender reads it, so the
@@ -83,6 +82,9 @@ public class AnalyticsMetricsHandler extends AbstractExtendedSynapseHandler {
                 }
             }
         }
+        // Mark backend-start after payload capture so the request build/serialization time counts as
+        // request-mediation latency rather than backend latency.
+        messageContext.setProperty(Constants.BACKEND_START_TIME_PROPERTY, System.currentTimeMillis());
         return true;
     }
 
@@ -99,7 +101,7 @@ public class AnalyticsMetricsHandler extends AbstractExtendedSynapseHandler {
             return false;
         }
         Object skipPublishMetrics = messageContext.getProperty(Constants.SKIP_METRICS_PUBLISHING);
-        if (skipPublishMetrics != null && (Boolean) skipPublishMetrics) {
+        if (Boolean.TRUE.equals(skipPublishMetrics)) {
             return false;
         }
         return !GatewayUtils.checkForFileBasedApiContexts(ApiUtils.getFullRequestPath(messageContext),
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
@@ -66,7 +66,44 @@ public class AnalyticsMetricsHandler extends AbstractExtendedSynapseHandler {
     @Override
     public boolean handleRequestOutFlow(MessageContext messageContext) {
         messageContext.setProperty(Constants.BACKEND_START_TIME_PROPERTY, System.currentTimeMillis());
+        // Capture the request payload here (request-out flow), the last point before the backend send
+        // where the request envelope is still current. Gated by the send_payloads analytics property.
+        // Building here sets MESSAGE_BUILDER_INVOKED before the pass-through sender reads it, so the
+        // engine re-serializes the built message to the backend (forwarding is preserved); doing this
+        // at request-in instead would run addressing before dispatch and break the send.
+        if (shouldCaptureRequestPayload(messageContext)) {
+            AnalyticsPayloadUtil.CapturedBody requestBody =
+                    AnalyticsPayloadUtil.extractPayload(messageContext,
+                            AnalyticsPayloadUtil.getPayloadSizeLimit(), "request");
+            if (requestBody != null && requestBody.getBody() != null) {
+                messageContext.setProperty(Constants.REQUEST_BODY, requestBody.getBody());
+                if (requestBody.getTransferEncoding() != null) {
+                    messageContext.setProperty(Constants.REQUEST_BODY_TRANSFER_ENCODING,
+                            requestBody.getTransferEncoding());
+                }
+            }
+        }
         return true;
+    }
+
+    /**
+     * Whether the request payload should be captured (and the message therefore built) on the
+     * request-out flow. Mirrors the skip conditions {@code handleResponseOutFlow} applies so we never
+     * build/re-serialize a request for a call that will not publish an analytics event: websocket
+     * subscriptions, file-based API contexts, and requests explicitly flagged {@code SKIP_METRICS_PUBLISHING}.
+     * Cheap checks are evaluated first and the file-based-context lookup last.
+     */
+    private boolean shouldCaptureRequestPayload(MessageContext messageContext) {
+        if (messageContext.getPropertyKeySet().contains(InboundWebsocketConstants.WEBSOCKET_SUBSCRIBER_PATH)
+                || !AnalyticsPayloadUtil.shouldSendPayloads()) {
+            return false;
+        }
+        Object skipPublishMetrics = messageContext.getProperty(Constants.SKIP_METRICS_PUBLISHING);
+        if (skipPublishMetrics != null && (Boolean) skipPublishMetrics) {
+            return false;
+        }
+        return !GatewayUtils.checkForFileBasedApiContexts(ApiUtils.getFullRequestPath(messageContext),
+                GatewayUtils.getTenantDomain());
     }
 
     @Override
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
@@ -75,9 +75,9 @@ public class AnalyticsMetricsHandler extends AbstractExtendedSynapseHandler {
                     AnalyticsPayloadUtil.extractPayload(messageContext,
                             AnalyticsPayloadUtil.getPayloadSizeLimit(), "request");
             if (requestBody != null && requestBody.getBody() != null) {
-                messageContext.setProperty(Constants.REQUEST_BODY, requestBody.getBody());
+                messageContext.setProperty(Constants.REQUEST_BODY_PROPERTY, requestBody.getBody());
                 if (requestBody.getTransferEncoding() != null) {
-                    messageContext.setProperty(Constants.REQUEST_BODY_TRANSFER_ENCODING,
+                    messageContext.setProperty(Constants.REQUEST_BODY_TRANSFER_ENCODING_PROPERTY,
                             requestBody.getTransferEncoding());
                 }
             }
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
@@ -33,6 +33,7 @@ import org.apache.synapse.transport.passthru.util.RelayUtils;
 import org.wso2.carbon.apimgt.impl.APIManagerConfiguration;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.Locale;
 import java.util.Map;
 
@@ -55,6 +56,10 @@ import static org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS;
 public class AnalyticsPayloadUtil {
 
     private static final Log log = LogFactory.getLog(AnalyticsPayloadUtil.class);
+
+    // Guards the invalid-payload_size_limit warning so a misconfigured value is logged once rather than
+    // on every capture attempt (getPayloadSizeLimit is called per request/response).
+    private static volatile boolean invalidLimitWarned = false;
 
     private AnalyticsPayloadUtil() {
     }
@@ -105,18 +110,51 @@ public class AnalyticsPayloadUtil {
             try {
                 int limit = Integer.parseInt(configs.get(Constants.PAYLOAD_SIZE_LIMIT));
                 if (limit > 0) {
+                    invalidLimitWarned = false;
                     return limit;
                 }
                 // A non-positive limit would silently drop every body (and break size math), so treat
                 // it as invalid rather than honouring it.
-                log.warn("Non-positive " + Constants.PAYLOAD_SIZE_LIMIT + " analytics property value. Using default "
-                        + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
+                warnInvalidLimitOnce("Non-positive " + Constants.PAYLOAD_SIZE_LIMIT
+                        + " analytics property value. Using default " + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
             } catch (NumberFormatException e) {
-                log.warn("Invalid " + Constants.PAYLOAD_SIZE_LIMIT + " analytics property value. Using default "
-                        + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
+                warnInvalidLimitOnce("Invalid " + Constants.PAYLOAD_SIZE_LIMIT
+                        + " analytics property value. Using default " + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
             }
         }
         return Constants.DEFAULT_PAYLOAD_SIZE_LIMIT;
+    }
+
+    /**
+     * Logs the invalid-{@code payload_size_limit} warning at most once per stretch of misconfiguration,
+     * so a bad value does not flood the logs on every per-request/response capture attempt. The flag is
+     * reset once a valid limit is read again, so a later re-misconfiguration is warned about afresh.
+     */
+    private static void warnInvalidLimitOnce(String message) {
+        if (!invalidLimitWarned) {
+            invalidLimitWarned = true;
+            log.warn(message);
+        }
+    }
+
+    /**
+     * Whether to capture payloads that do not declare a {@code Content-Length} (e.g. chunked
+     * transfer-encoding), read from the {@code capture_payloads_without_content_length} analytics
+     * property. Defaults to {@code false}.
+     *
+     * <p>Such a body cannot be size-checked before {@link RelayUtils#buildMessage} materializes it in
+     * full, so capturing a large chunked body would buffer the whole thing into memory (an OOM risk
+     * under load) only to drop it afterwards if it exceeds {@code payload_size_limit}. It is therefore
+     * off by default: with the flag off these bodies are skipped (never built), so the default
+     * configuration is memory-safe. An operator who needs them and has headroom can opt in, accepting
+     * the memory cost. Applies symmetrically to request and response capture.</p>
+     *
+     * @return true if bodies without a declared Content-Length should be captured
+     */
+    public static boolean shouldCapturePayloadsWithoutContentLength() {
+        Map<String, String> configs = APIManagerConfiguration.getAnalyticsProperties();
+        return configs != null
+                && Boolean.parseBoolean(configs.get(Constants.CAPTURE_PAYLOADS_WITHOUT_CONTENT_LENGTH));
     }
 
     /**
@@ -129,8 +167,11 @@ public class AnalyticsPayloadUtil {
      * truncated, so the publisher never has to deal with a partial/invalid body. Where the payload
      * advertises its size via a {@code Content-Length} header the check is applied <b>before</b>
      * {@link RelayUtils#buildMessage} so an oversized body is never buffered into memory or
-     * re-serialized. A chunked body with no {@code Content-Length} is still built and then dropped
-     * (the size is unknowable without reading it).</p>
+     * re-serialized. A body with no {@code Content-Length} (e.g. chunked transfer-encoding) cannot be
+     * size-checked before building, so by default it is <b>skipped</b> (never built) to keep the
+     * default configuration memory-safe; set {@code capture_payloads_without_content_length=true} to
+     * capture it anyway (it is then built in full and dropped afterwards if it exceeds the limit,
+     * accepting the memory cost). See {@link #shouldCapturePayloadsWithoutContentLength()}.</p>
      *
      * <p>Works for both the request (called from {@code handleRequestOutFlow}, just before the backend
      * send) and the response (called at event-collection time). When a body is captured,
@@ -139,8 +180,9 @@ public class AnalyticsPayloadUtil {
      * read-only build pattern used by the AI mediator and the threat-protection/schema validators.
      * Requests with no entity body (GET/DELETE) are skipped so they are left byte-identical.</p>
      *
-     * <p>The limit is compared against bytes for the pre-build {@code Content-Length} gate and the
-     * binary path, and against {@link String#length()} (characters) for text/JSON/XML.</p>
+     * <p>The limit is compared against bytes throughout: the declared {@code Content-Length} for the
+     * pre-build gate, the decoded byte count for the binary path, and the UTF-8 encoded byte length
+     * for text/JSON/XML.</p>
      *
      * <p>When {@code send_payloads} is enabled but a body is nonetheless skipped or dropped, the reason
      * is logged at debug level (keyed by {@code direction}) so operators are not left wondering why a
@@ -182,6 +224,20 @@ public class AnalyticsPayloadUtil {
                     log.debug("Dropping " + direction + " body from analytics: declared Content-Length "
                             + declaredLength + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
                             + "; body not built. Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                }
+                return null;
+            }
+
+            // No declared Content-Length (e.g. chunked transfer-encoding): the size cannot be bounded
+            // before building, so building here would materialize the whole body in memory (an OOM risk
+            // under load). Skip it unless the operator has explicitly opted in. Off by default keeps the
+            // default configuration memory-safe; applies symmetrically to request and response.
+            if (declaredLength < 0 && !shouldCapturePayloadsWithoutContentLength()) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: no Content-Length header "
+                            + "(e.g. chunked transfer-encoding), so the body size cannot be bounded before "
+                            + "building. Set " + Constants.CAPTURE_PAYLOADS_WITHOUT_CONTENT_LENGTH
+                            + "=true to capture it (materializes the full body in memory).");
                 }
                 return null;
             }
@@ -241,19 +297,23 @@ public class AnalyticsPayloadUtil {
     }
 
     /**
-     * Wraps a text/JSON/XML payload, dropping it (returning {@code null}) when it exceeds
-     * {@code sizeLimit} characters so only a whole body is ever published. A drop is logged at debug
-     * level (keyed by {@code direction}/{@code kind}) so a missing body can be explained.
+     * Wraps a text/JSON/XML payload, dropping it (returning {@code null}) when its UTF-8 encoded size
+     * exceeds {@code sizeLimit} bytes so only a whole body is ever published. Measuring bytes (not
+     * {@link String#length()}, which counts UTF-16 code units) keeps this consistent with the
+     * byte-based pre-build {@code Content-Length} gate and binary path, and with the byte semantics of
+     * {@code payload_size_limit}. A drop is logged at debug level (keyed by {@code direction}/
+     * {@code kind}) so a missing body can be explained.
      */
     private static CapturedBody capture(String payload, int sizeLimit, String transferEncoding,
                                         String direction, String kind) {
         if (payload == null) {
             return null;
         }
-        if (payload.length() > sizeLimit) {
+        int byteLength = payload.getBytes(StandardCharsets.UTF_8).length;
+        if (byteLength > sizeLimit) {
             if (log.isDebugEnabled()) {
-                log.debug("Dropping " + direction + " " + kind + " body from analytics: " + payload.length()
-                        + " characters exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                log.debug("Dropping " + direction + " " + kind + " body from analytics: " + byteLength
+                        + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
                         + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
             }
             return null;
--- a/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
@@ -1,0 +1,307 @@
+/*
+ *  Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package org.wso2.carbon.apimgt.gateway.handlers.analytics;
+
+import org.apache.axiom.om.OMElement;
+import org.apache.axiom.soap.SOAPBody;
+import org.apache.axiom.soap.SOAPEnvelope;
+import org.apache.axis2.transport.base.BaseConstants;
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.http.HttpHeaders;
+import org.apache.synapse.MessageContext;
+import org.apache.synapse.commons.json.JsonUtil;
+import org.apache.synapse.core.axis2.Axis2MessageContext;
+import org.apache.synapse.transport.passthru.PassThroughConstants;
+import org.apache.synapse.transport.passthru.util.RelayUtils;
+import org.wso2.carbon.apimgt.impl.APIManagerConfiguration;
+
+import java.io.IOException;
+import java.util.Locale;
+import java.util.Map;
+
+import javax.xml.stream.XMLStreamException;
+
+import static org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS;
+
+/**
+ * Utility for capturing the request/response payload (body) for analytics publishing.
+ *
+ * <p>Body capture is an opt-in feature (gated by the {@code send_payloads} analytics property) because it
+ * forces the pass-through message to be built into memory and captures potentially sensitive data.
+ * All content types are captured except streaming ({@code text/event-stream}) and {@code multipart/*},
+ * which have no single materializable body. JSON and text payloads are captured as-is; binary payloads
+ * are Base64-encoded (mirroring the Moesif standard, where the publisher tags them
+ * {@code transferEncoding=base64}). A payload larger than the configurable size limit is dropped
+ * (not truncated) so the publisher only ever receives a whole, valid body or nothing; the backend
+ * always receives the full body regardless.</p>
+ */
+public class AnalyticsPayloadUtil {
+
+    private static final Log log = LogFactory.getLog(AnalyticsPayloadUtil.class);
+
+    private AnalyticsPayloadUtil() {
+    }
+
+    /**
+     * Holds a captured body together with its Moesif transfer-encoding hint.
+     * {@code transferEncoding} is {@code "base64"} for pre-encoded binary payloads and {@code null}
+     * for JSON/text (the publisher decides JSON-vs-base64 for those via Moesif's {@code BodyParser}).
+     */
+    public static class CapturedBody {
+        private final String body;
+        private final String transferEncoding;
+
+        CapturedBody(String body, String transferEncoding) {
+            this.body = body;
+            this.transferEncoding = transferEncoding;
+        }
+
+        public String getBody() {
+            return body;
+        }
+
+        public String getTransferEncoding() {
+            return transferEncoding;
+        }
+    }
+
+    /**
+     * Whether request/response body capture is enabled via the {@code send_payloads} analytics property
+     * (named to match the sibling {@code send_headers} option). Defaults to {@code false}.
+     *
+     * @return true if body capture is enabled
+     */
+    public static boolean shouldSendPayloads() {
+        Map<String, String> configs = APIManagerConfiguration.getAnalyticsProperties();
+        return configs != null && Boolean.parseBoolean(configs.get(Constants.SEND_PAYLOAD));
+    }
+
+    /**
+     * Maximum number of bytes/characters captured per body, read from the {@code payload_size_limit}
+     * analytics property. Falls back to {@link Constants#DEFAULT_PAYLOAD_SIZE_LIMIT} when unset or invalid.
+     *
+     * @return the payload size limit
+     */
+    public static int getPayloadSizeLimit() {
+        Map<String, String> configs = APIManagerConfiguration.getAnalyticsProperties();
+        if (configs != null && configs.get(Constants.PAYLOAD_SIZE_LIMIT) != null) {
+            try {
+                int limit = Integer.parseInt(configs.get(Constants.PAYLOAD_SIZE_LIMIT));
+                if (limit > 0) {
+                    return limit;
+                }
+                // A non-positive limit would silently drop every body (and break size math), so treat
+                // it as invalid rather than honouring it.
+                log.warn("Non-positive " + Constants.PAYLOAD_SIZE_LIMIT + " analytics property value. Using default "
+                        + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
+            } catch (NumberFormatException e) {
+                log.warn("Invalid " + Constants.PAYLOAD_SIZE_LIMIT + " analytics property value. Using default "
+                        + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
+            }
+        }
+        return Constants.DEFAULT_PAYLOAD_SIZE_LIMIT;
+    }
+
+    /**
+     * Builds the message (materializing the pass-through stream) and returns the captured body along
+     * with its transfer-encoding hint. Streaming ({@code text/event-stream}) and {@code multipart/*}
+     * payloads are skipped. JSON and text are returned verbatim with a {@code null} encoding; binary
+     * payloads are Base64-encoded with encoding {@code "base64"}.
+     *
+     * <p>A body larger than {@code sizeLimit} is <b>dropped</b> (returns {@code null}) rather than
+     * truncated, so the publisher never has to deal with a partial/invalid body. Where the payload
+     * advertises its size via a {@code Content-Length} header the check is applied <b>before</b>
+     * {@link RelayUtils#buildMessage} so an oversized body is never buffered into memory or
+     * re-serialized. A chunked body with no {@code Content-Length} is still built and then dropped
+     * (the size is unknowable without reading it).</p>
+     *
+     * <p>Works for both the request (called from {@code handleRequestOutFlow}, just before the backend
+     * send) and the response (called at event-collection time). When a body is captured,
+     * {@link RelayUtils#buildMessage} sets {@code MESSAGE_BUILDER_INVOKED}, so the pass-through sender
+     * re-serializes the built envelope to the backend — forwarding is preserved. This is the same
+     * read-only build pattern used by the AI mediator and the threat-protection/schema validators.
+     * Requests with no entity body (GET/DELETE) are skipped so they are left byte-identical.</p>
+     *
+     * <p>The limit is compared against bytes for the pre-build {@code Content-Length} gate and the
+     * binary path, and against {@link String#length()} (characters) for text/JSON/XML.</p>
+     *
+     * <p>When {@code send_payloads} is enabled but a body is nonetheless skipped or dropped, the reason
+     * is logged at debug level (keyed by {@code direction}) so operators are not left wondering why a
+     * body is missing from Moesif.</p>
+     *
+     * @param messageContext the Synapse message context (request or response)
+     * @param sizeLimit      maximum number of characters (text) or bytes (binary) to capture
+     * @param direction      {@code "request"} or {@code "response"}, used only for debug log messages
+     * @return the captured body, or {@code null} if there is nothing capturable or it exceeds the limit
+     */
+    public static CapturedBody extractPayload(MessageContext messageContext, int sizeLimit, String direction) {
+        try {
+            org.apache.axis2.context.MessageContext axis2MC =
+                    ((Axis2MessageContext) messageContext).getAxis2MessageContext();
+
+            // Nothing to capture (and nothing to build) for entity-less requests such as GET/DELETE.
+            if (Boolean.TRUE.equals(axis2MC.getProperty(PassThroughConstants.NO_ENTITY_BODY))) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: message has no entity body "
+                            + "(e.g. GET/DELETE).");
+                }
+                return null;
+            }
+
+            String contentType = getContentType(axis2MC);
+            if (isExcludedContentType(contentType)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: content type '" + contentType
+                            + "' is excluded (streaming/multipart bodies are never captured).");
+                }
+                return null;
+            }
+
+            // Pre-build gate: if the payload advertises its size and it exceeds the limit, drop it now
+            // without building — so an oversized body is never buffered into memory or re-serialized.
+            long declaredLength = getDeclaredContentLength(axis2MC);
+            if (declaredLength > sizeLimit) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Dropping " + direction + " body from analytics: declared Content-Length "
+                            + declaredLength + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                            + "; body not built. Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                }
+                return null;
+            }
+
+            RelayUtils.buildMessage(axis2MC);
+
+            // JSON: return the text as-is; the publisher's BodyParser renders it as a structured object.
+            if (JsonUtil.hasAJsonPayload(axis2MC)) {
+                return capture(JsonUtil.jsonPayloadToString(axis2MC), sizeLimit, null, direction, "JSON");
+            }
+
+            SOAPEnvelope env = axis2MC.getEnvelope();
+            if (env == null || env.getBody() == null) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: message has no SOAP body.");
+                }
+                return null;
+            }
+            SOAPBody soapBody = env.getBody();
+            OMElement firstElement = soapBody.getFirstElement();
+            if (firstElement == null) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: message body is empty.");
+                }
+                return null;
+            }
+            env.buildWithAttachments();
+
+            if (BaseConstants.DEFAULT_BINARY_WRAPPER.equals(firstElement.getQName())) {
+                // Binary payload: the wrapper's text is the Base64 of the raw bytes. Decode, and drop
+                // the whole body if the raw bytes exceed the limit; otherwise pass the bytes through.
+                byte[] bytes = Base64.decodeBase64(firstElement.getText());
+                if (bytes.length > sizeLimit) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Dropping " + direction + " binary body from analytics: " + bytes.length
+                                + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                                + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                    }
+                    return null;
+                }
+                return new CapturedBody(Base64.encodeBase64String(bytes), Constants.TRANSFER_ENCODING_BASE64);
+            } else if (BaseConstants.DEFAULT_TEXT_WRAPPER.equals(firstElement.getQName())) {
+                // Plain text payload.
+                return capture(firstElement.getText(), sizeLimit, null, direction, "text");
+            } else {
+                // XML / SOAP payload.
+                return capture(firstElement.toString(), sizeLimit, null, direction, "XML/SOAP");
+            }
+        } catch (IOException | XMLStreamException e) {
+            log.error("Error occurred while capturing the " + direction + " message payload for analytics", e);
+            return null;
+        } catch (RuntimeException e) {
+            // Fail safe: payload capture must never disrupt the request/response flow.
+            log.error("Unexpected error while capturing the " + direction + " message payload for analytics", e);
+            return null;
+        }
+    }
+
+    /**
+     * Wraps a text/JSON/XML payload, dropping it (returning {@code null}) when it exceeds
+     * {@code sizeLimit} characters so only a whole body is ever published. A drop is logged at debug
+     * level (keyed by {@code direction}/{@code kind}) so a missing body can be explained.
+     */
+    private static CapturedBody capture(String payload, int sizeLimit, String transferEncoding,
+                                        String direction, String kind) {
+        if (payload == null) {
+            return null;
+        }
+        if (payload.length() > sizeLimit) {
+            if (log.isDebugEnabled()) {
+                log.debug("Dropping " + direction + " " + kind + " body from analytics: " + payload.length()
+                        + " characters exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                        + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+            }
+            return null;
+        }
+        return new CapturedBody(payload, transferEncoding);
+    }
+
+    /**
+     * The declared body size from the {@code Content-Length} transport header, or {@code -1} when it is
+     * absent (e.g. chunked transfer-encoding) or unparseable. Read before building so an oversized body
+     * can be skipped without buffering it. The transport-headers map is case-insensitive, matching the
+     * exact-case lookup idiom used elsewhere (e.g. {@code getRequestSize}/{@code getResponseSize}).
+     */
+    private static long getDeclaredContentLength(org.apache.axis2.context.MessageContext axis2MC) {
+        Object headersObj = axis2MC.getProperty(TRANSPORT_HEADERS);
+        if (headersObj instanceof Map) {
+            Object contentLength = ((Map) headersObj).get(HttpHeaders.CONTENT_LENGTH);
+            if (contentLength != null) {
+                try {
+                    return Long.parseLong(contentLength.toString().trim());
+                } catch (NumberFormatException e) {
+                    return -1;
+                }
+            }
+        }
+        return -1;
+    }
+
+    private static String getContentType(org.apache.axis2.context.MessageContext axis2MC) {
+        Object headersObj = axis2MC.getProperty(TRANSPORT_HEADERS);
+        if (headersObj instanceof Map) {
+            Object contentType = ((Map) headersObj).get(HttpHeaders.CONTENT_TYPE);
+            if (contentType != null) {
+                return contentType.toString();
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Streaming (SSE) and multipart payloads have no single materializable body and must never be
+     * captured (building/buffering them would be incorrect or disruptive).
+     */
+    private static boolean isExcludedContentType(String contentType) {
+        if (contentType == null) {
+            return false;
+        }
+        String lower = contentType.toLowerCase(Locale.ROOT);
+        return lower.contains("event-stream") || lower.startsWith("multipart/");
+    }
+}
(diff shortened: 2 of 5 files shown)


components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,307 @@
...
+                if (log.isDebugEnabled()) {
+                    log.debug("Dropping " + direction + " body from analytics: declared Content-Length "
+                            + declaredLength + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                            + "; body not built. Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                }
+                return null;
+            }
+
+            RelayUtils.buildMessage(axis2MC);
+
```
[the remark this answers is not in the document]
vinupa: Thanks for flagging this. Addressed in the latest commit.
Chunked bodies with no Content-Length are no longer built-then-dropped by default. A new opt-in property, `capture_payloads_without_content_length` (default false), gates them: when off, a body with no declared length is skipped before buildMessage (never materialized), so the default configuration is memory-safe; when on, operators who need these bodies and have the headroom can capture them. Applies symmetrically to request and response, and the skip reason is logged at debug level. Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
```diff
@@ -66,9 +66,46 @@ public boolean handleRequestInFlow(MessageContext messageContext) {
...
+     */
+    private boolean shouldCaptureRequestPayload(MessageContext messageContext) {
+        if (messageContext.getPropertyKeySet().contains(InboundWebsocketConstants.WEBSOCKET_SUBSCRIBER_PATH)
+                || !AnalyticsPayloadUtil.shouldSendPayloads()) {
+            return false;
+        }
+        Object skipPublishMetrics = messageContext.getProperty(Constants.SKIP_METRICS_PUBLISHING);
+        if (skipPublishMetrics != null && (Boolean) skipPublishMetrics) {
+            return false;
+        }
```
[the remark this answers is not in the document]
vinupa: This is addressed with the latest commit.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * @param messageContext the Synapse message context (request or response)
+     * @param sizeLimit      maximum number of characters (text) or bytes (binary) to capture
+     * @param direction      {@code "request"} or {@code "response"}, used only for debug log messages
+     * @return the captured body, or {@code null} if there is nothing capturable or it exceeds the limit
+     */
+    public static CapturedBody extractPayload(MessageContext messageContext, int sizeLimit, String direction) {
+        try {
+            org.apache.axis2.context.MessageContext axis2MC =
+                    ((Axis2MessageContext) messageContext).getAxis2MessageContext();
```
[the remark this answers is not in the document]
vinupa: The fail-safe is already in place. `ClassCastException` is a RuntimeException, and the entire body of `extractPayload` runs inside a catch (RuntimeException e) that logs and returns null, so a non-`Axis2 MessageContext` degrades to "no capture," never disrupting the flow. Adding an explicit `instanceof` guard would be redundant with that catch. Resolved.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,307 @@
...
+            return null;
+        }
+        if (payload.length() > sizeLimit) {
+            if (log.isDebugEnabled()) {
+                log.debug("Dropping " + direction + " " + kind + " body from analytics: " + payload.length()
+                        + " characters exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                        + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+            }
+            return null;
+        }
```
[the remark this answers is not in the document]
vinupa: This is addressed with the latest commit.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,307 @@
...
+    private static String getContentType(org.apache.axis2.context.MessageContext axis2MC) {
+        Object headersObj = axis2MC.getProperty(TRANSPORT_HEADERS);
+        if (headersObj instanceof Map) {
+            Object contentType = ((Map) headersObj).get(HttpHeaders.CONTENT_TYPE);
+            if (contentType != null) {
+                return contentType.toString();
+            }
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
vinupa: These lookups read the `TRANSPORT_HEADERS` map, which in the Synapse pass-through transport is a case-insensitive TreeMap (populated with String.CASE_INSENSITIVE_ORDER on the source/target sides), so exact case `.get(HttpHeaders.CONTENT_LENGTH)` / `.get(HttpHeaders.CONTENT_TYPE)` already resolve regardless of the wire casing. No issue here. Resolved.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,307 @@
...
+                if (bytes.length > sizeLimit) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Dropping " + direction + " binary body from analytics: " + bytes.length
+                                + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                                + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                    }
+                    return null;
+                }
+                return new CapturedBody(Base64.encodeBase64String(bytes), Constants.TRANSFER_ENCODING_BASE64);
+            } else if (BaseConstants.DEFAULT_TEXT_WRAPPER.equals(firstElement.getQName())) {
```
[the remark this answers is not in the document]
vinupa: Now bounded by default. If the body declares a `Content-Length`, the pre-build gate drops anything over `payload_size_limit` before `buildMessage`, so the binary branch decodes at most `sizeLimit` bytes. If it has no `Content-Length` (e.g. chunked), it's skipped entirely by default (`declaredLength < 0 && !shouldCapturePayloadsWithoutContentLength()`). A large chunked binary only reaches the decode under the explicit, documented `capture_payloads_without_content_length=true` opt-in. Also, `firstElement.getText()` is already the fully-materialized Base64 string, and the post-decode `bytes.length > sizeLimit` check stays as a final guard. Resolved.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/Constants.java
```diff
@@ -90,6 +90,19 @@ public static final class ERROR_CODE_RANGES {
     public static final String REQUEST_HEADER_MASK = "request_headers";
     public static final String MASK_VALUE = "*****";
 
+    public static final String SEND_PAYLOAD = "send_payloads";
```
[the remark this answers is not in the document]
vinupa: The singular constant name is intentional. It mirrors its direct sibling in the same file, `SEND_HEADER = "send_headers"` (both map a singular constant name to a plural property value). Leaving as-is for consistency. Resolved.

vinupa:
@coderabbitai full review

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,350 @@
...
+    private static String getContentType(org.apache.axis2.context.MessageContext axis2MC) {
+        Object headersObj = axis2MC.getProperty(TRANSPORT_HEADERS);
+        if (headersObj instanceof Map) {
+            Object contentType = ((Map) headersObj).get(HttpHeaders.CONTENT_TYPE);
+            if (contentType != null) {
+                return contentType.toString();
+            }
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
vinupa: `TRANSPORT_HEADERS` isn't case-preserving. The Synapse pass-through transport stores it in a case-insensitive map for both request and response, so `.get(HttpHeaders.CONTENT_LENGTH)` / `.get(HttpHeaders.CONTENT_TYPE)` resolve regardless of wire casing (lowercase HTTP/2 headers included). The oversize check and the SSE/multipart exclusion are therefore not bypassed.

This is similar to the approach already used against this map elsewhere in the gateway: `SynapseAnalyticsDataProvider.getRequestSize()` / `getResponseSize()` and `RegularExpressionProtector.isPayloadSizeExceeded()` all read `TRANSPORT_HEADERS` with a plain `.get(HttpHeaders.CONTENT_LENGTH)`.

`getRequestContentType()` iterates with `equalsIgnoreCase` for a different reason: it reads the analytics-metadata copy of the request headers, which is a case-sensitive copy. So the workaround is needed there but not for direct `TRANSPORT_HEADERS` reads. Resolved.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,350 @@
...
+     *
+     * @param messageContext the Synapse message context (request or response)
+     * @param sizeLimit      maximum number of characters (text) or bytes (binary) to capture
+     * @param direction      {@code "request"} or {@code "response"}, used only for debug log messages
+     * @return the captured body, or {@code null} if there is nothing capturable or it exceeds the limit
+     */
+    public static CapturedBody extractPayload(MessageContext messageContext, int sizeLimit, String direction) {
+        try {
+            org.apache.axis2.context.MessageContext axis2MC =
+                    ((Axis2MessageContext) messageContext).getAxis2MessageContext();
```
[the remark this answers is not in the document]
vinupa: Addressed in a previous comment as well. Already non-disruptive: `extractPayload`’s body runs inside a `catch (RuntimeException e)`, and `ClassCastException` is a `RuntimeException`, so a non-Axis2 context returns `null` without disturbing the flow. In the synapse gateway handler path the context is always an `Axis2MessageContext` (this same direct approach is used throughout the analytics package). Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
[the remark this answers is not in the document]
vinupa: Addressed in a previous comment as well. `TRANSPORT_HEADERS` isn't case-preserving. The pass-through transport stores it in a case-insensitive map (both request and response), so `.get(HttpHeaders.CONTENT_LENGTH)` / `.get(HttpHeaders.CONTENT_TYPE)` resolve regardless of wire casing. (`getRequestContentType()` iterates with `equalsIgnoreCase` only because it reads the case-sensitive `HashMap` copy of headers stashed as analytics metadata, not `TRANSPORT_HEADERS`.) Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,350 @@
...
+            OMElement firstElement = soapBody.getFirstElement();
+            if (firstElement == null) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: message body is empty.");
+                }
+                return null;
+            }
+            env.buildWithAttachments();
+
+            if (BaseConstants.DEFAULT_BINARY_WRAPPER.equals(firstElement.getQName())) {
```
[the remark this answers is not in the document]
vinupa: For the REST XML/text flow this handles, the built envelope's SOAP body has a single child element that is the full payload, so `firstElement.toString()` captures the complete body. Switching to `soapBody.toString()` would wrap every captured body in a synthetic `<soapenv:Body>…</soapenv:Body>` that the client never sent, corrupting the common case to accommodate multi-child SOAP bodies, which are rare (WS-I recommends a single body child) and would anyway need element concatenation, not the `<Body>` wrapper. Keeping `firstElement`. Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,350 @@
...
+                    return null;
+                }
+                return new CapturedBody(Base64.encodeBase64String(bytes), Constants.TRANSFER_ENCODING_BASE64);
+            } else if (BaseConstants.DEFAULT_TEXT_WRAPPER.equals(firstElement.getQName())) {
+                // Plain text payload.
+                return capture(firstElement.getText(), sizeLimit, null, direction, "text");
+            } else {
+                // XML / SOAP payload.
+                return capture(firstElement.toString(), sizeLimit, null, direction, "XML/SOAP");
+            }
```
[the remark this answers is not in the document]
vinupa: Addressed in above comment. Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,350 @@
...
+            SOAPBody soapBody = env.getBody();
+            OMElement firstElement = soapBody.getFirstElement();
+            if (firstElement == null) {
+                if (log.isDebugEnabled()) {
+                    log.debug("No " + direction + " body captured for analytics: message body is empty.");
+                }
+                return null;
+            }
+            env.buildWithAttachments();
```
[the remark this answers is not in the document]
vinupa: MTOM/SwA are transported as `multipart/related`, which `isExcludedContentType` already excludes before this line, so no attachment-bearing message reaches `buildWithAttachments`. For the non-multipart messages that do, it simply builds the envelope, and the single binary body it materializes is already bounded by the pre-build `Content-Length` gate. Switching to `build()` risks not fully materializing the binary/text wrapper content we then read via `getText()`. Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/SynapseAnalyticsDataProvider.java
```diff
@@ -514,6 +514,36 @@ public Map<String, Object> getProperties() {
             }
         }
 
+        // Request/response body (optional)
+        if (AnalyticsPayloadUtil.shouldSendPayloads()) {
+            log.debug("Including request/response body in analytics event");
```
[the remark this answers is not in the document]
vinupa: The message is a constant string with no concatenation, so there's nothing to guard, commons-logging is not operational when debug is disabled. The string-handling cost the comment describes only applies to concatenated messages (which are guarded elsewhere in this file). No change needed. Resolving.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,350 @@
...
+                // it as invalid rather than honouring it.
+                log.warn("Non-positive " + Constants.PAYLOAD_SIZE_LIMIT + " analytics property value. Using default "
+                        + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
+            } catch (NumberFormatException e) {
+                log.warn("Invalid " + Constants.PAYLOAD_SIZE_LIMIT + " analytics property value. Using default "
+                        + Constants.DEFAULT_PAYLOAD_SIZE_LIMIT);
+            }
+        }
+        return Constants.DEFAULT_PAYLOAD_SIZE_LIMIT;
+    }
```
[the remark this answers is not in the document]
vinupa: Addressed this issue in a commit. Added a _private static volatile boolean_ `invalidLimitWarned` flag. `getPayloadSizeLimit()` now routes both warnings through `warnInvalidLimitOnce(...)`, which logs at most once, and resets the flag when a valid limit is read again. So a misconfiguration warns once (not per request), and a later re-misconfiguration warns afresh.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,367 @@
...
+    /**
+     * Streaming (SSE) and multipart payloads have no single materializable body and must never be
+     * captured (building/buffering them would be incorrect or disruptive).
+     */
+    private static boolean isExcludedContentType(String contentType) {
+        if (contentType == null) {
+            return false;
+        }
+        String lower = contentType.toLowerCase(Locale.ROOT);
+        return lower.contains("event-stream") || lower.startsWith("multipart/");
```
AnuGayan: 🟠 **major — `application/x-www-form-urlencoded` bodies are captured as XML, not the form body**

For form-encoded content, Axis2's `XFormURLEncodedBuilder` materialises the body into an OM tree (`<xformValues><a>1</a><b>2</b></xformValues>`). That element matches neither `DEFAULT_BINARY_WRAPPER` nor `DEFAULT_TEXT_WRAPPER`, so it falls through to the XML branch at line 287 and `firstElement.toString()` publishes the XML representation.

**Failure scenario:** `POST /orders` with `Content-Type: application/x-www-form-urlencoded`, `Content-Length: 7`, body `a=1&b=2`. Moesif receives `requestContentType: "application/x-www-form-urlencoded"` alongside `requestBody: "<xformValues>..."` — the label and the payload disagree, and the original wire body is unrecoverable.

**Secondary concern on the same content type:** because capture sets `MESSAGE_BUILDER_INVOKED`, the pass-through sender re-serialises through `XFormURLEncodedFormatter` on the way to the backend. Parameter ordering and repeated-key handling are not guaranteed to survive that round-trip, so enabling `send_payloads` can change what the backend receives for form POSTs.

Excluding it alongside SSE/multipart fixes the wrong-content bug *and* removes the forwarding risk. (Also tightening `contains("event-stream")` → `startsWith("text/event-stream")` to match the documented intent and the `multipart/` check beside it.)

```suggestion
        String lower = contentType.toLowerCase(Locale.ROOT);
        return lower.startsWith("text/event-stream")
                || lower.startsWith("multipart/")
                || lower.startsWith("application/x-www-form-urlencoded");
```

If capturing form bodies is actually wanted, reconstruct the `k=v&…` string from the `xformValues` children instead of emitting `toString()`.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,367 @@
...
+                    }
+                    return null;
+                }
+                return new CapturedBody(Base64.encodeBase64String(bytes), Constants.TRANSFER_ENCODING_BASE64);
+            } else if (BaseConstants.DEFAULT_TEXT_WRAPPER.equals(firstElement.getQName())) {
+                // Plain text payload.
+                return capture(firstElement.getText(), sizeLimit, null, direction, "text");
+            } else {
+                // XML / SOAP payload.
+                return capture(firstElement.toString(), sizeLimit, null, direction, "XML/SOAP");
```
AnuGayan: 🟡 **minor — XML/SOAP capture keeps only the first body element**

`firstElement.toString()` drops sibling elements and the SOAP header. A SOAP body with two children publishes only the first — silently partial data, which contradicts the "whole, valid body or nothing" contract the size gate is built around.

```suggestion
                // XML / SOAP payload: serialise every child of the body so a multi-element body is
                // not silently truncated to its first element.
                Iterator<OMElement> children = soapBody.getChildElements();
                StringBuilder xml = new StringBuilder();
                while (children.hasNext()) {
                    xml.append(children.next().toString());
                }
                return capture(xml.toString(), sizeLimit, null, direction, "XML/SOAP");
```

Needs `import java.util.Iterator;`. Note deliberately **not** `soapBody.toString()` (which `getResponseSize()` uses at line ~832) — that would add a `<soapenv:Body>` wrapper that was never in the REST XML body on the wire.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,367 @@
...
+                            + "; body not built. Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                }
+                return null;
+            }
+
+            // No declared Content-Length (e.g. chunked transfer-encoding): the size cannot be bounded
+            // before building, so building here would materialize the whole body in memory (an OOM risk
+            // under load). Skip it unless the operator has explicitly opted in. Off by default keeps the
+            // default configuration memory-safe; applies symmetrically to request and response.
+            if (declaredLength < 0 && !shouldCapturePayloadsWithoutContentLength()) {
```
AnuGayan: 🟡 **minor — this skip also fires when the message is *already* built**

The rationale for the skip is "sizing requires a full in-memory build". That reasoning stops holding once `MESSAGE_BUILDER_INVOKED` is already `TRUE` — the body is materialised, the memory has been spent, and skipping just loses the data for free.

This is not hypothetical: `getResponseSize()` builds when `build_response_message = true` and runs at `SynapseAnalyticsDataProvider.java:467`, i.e. **before** the response capture at line 537. Schema validation, threat protection, the AI mediator and `McpMediator` do the same on the request side.

**Failure scenario:** `build_response_message = true`, `send_payloads = true`, chunked backend response. `getResponseSize()` builds the message; `extractPayload` then drops the body because `Content-Length` is absent — paying the full memory cost and publishing nothing.

```suggestion
            boolean alreadyBuilt = Boolean.TRUE.equals(
                    axis2MC.getProperty(PassThroughConstants.MESSAGE_BUILDER_INVOKED));
            if (declaredLength < 0 && !alreadyBuilt && !shouldCapturePayloadsWithoutContentLength()) {
```

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,367 @@
...
+                byte[] bytes = Base64.decodeBase64(firstElement.getText());
+                if (bytes.length > sizeLimit) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Dropping " + direction + " binary body from analytics: " + bytes.length
+                                + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
+                                + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
+                    }
+                    return null;
+                }
+                return new CapturedBody(Base64.encodeBase64String(bytes), Constants.TRANSFER_ENCODING_BASE64);
```
AnuGayan, line 296: 🟡 **minor — pointless base64 decode/re-encode round-trip**

The binary wrapper's text is *already* base64 of the raw bytes, so `Base64.encodeBase64String(Base64.decodeBase64(text))` returns a normalised copy of its own input — at the cost of an extra `byte[]` plus an extra `String`, ~2.3× body size in peak heap. Only the *size check* needs the decoded length, and that can be derived arithmetically.

```suggestion
                // Binary payload: the wrapper's text is already the Base64 of the raw bytes, so pass it
                // through untouched. The decoded size is derived arithmetically rather than by decoding
                // and re-encoding the whole body purely to measure it.
                String encoded = firstElement.getText();
                long decodedLength = (encoded.length() / 4L) * 3L;
                if (decodedLength > sizeLimit) {
                    if (log.isDebugEnabled()) {
                        log.debug("Dropping " + direction + " binary body from analytics: ~" + decodedLength
                                + " bytes exceeds " + Constants.PAYLOAD_SIZE_LIMIT + " of " + sizeLimit
                                + ". Increase " + Constants.PAYLOAD_SIZE_LIMIT + " to capture it.");
                    }
                    return null;
                }
                return new CapturedBody(encoded, Constants.TRANSFER_ENCODING_BASE64);
```

This leaves `org.apache.commons.codec.binary.Base64` unused — drop the import too.

If the round-trip *is* deliberate (e.g. stripping chunking/whitespace that commons-codec may have introduced), please say so in the comment — as written it reads as accidental.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,367 @@
...
+    /**
+     * Wraps a text/JSON/XML payload, dropping it (returning {@code null}) when its UTF-8 encoded size
+     * exceeds {@code sizeLimit} bytes so only a whole body is ever published. Measuring bytes (not
+     * {@link String#length()}, which counts UTF-16 code units) keeps this consistent with the
+     * byte-based pre-build {@code Content-Length} gate and binary path, and with the byte semantics of
+     * {@code payload_size_limit}. A drop is logged at debug level (keyed by {@code direction}/
+     * {@code kind}) so a missing body can be explained.
+     */
+    private static CapturedBody capture(String payload, int sizeLimit, String transferEncoding,
+                                        String direction, String kind) {
```
AnuGayan: 🔵 **nit — `transferEncoding` is dead**

It is passed `null` at all three call sites (lines 248, 284, 287). Dropping the parameter and returning `new CapturedBody(payload, null)` removes a parameter that only ever carries one value — PMD/CodeRabbit will flag it as-is.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
```diff
@@ -0,0 +1,367 @@
...
+     *
+     * @return the payload size limit
+     */
+    public static int getPayloadSizeLimit() {
+        Map<String, String> configs = APIManagerConfiguration.getAnalyticsProperties();
+        if (configs != null && configs.get(Constants.PAYLOAD_SIZE_LIMIT) != null) {
+            try {
+                int limit = Integer.parseInt(configs.get(Constants.PAYLOAD_SIZE_LIMIT));
+                if (limit > 0) {
+                    invalidLimitWarned = false;
```
AnuGayan: 🔵 **nit — volatile write on every request**

`getPayloadSizeLimit()` runs per request *and* per response, so this executes a volatile store on every successful capture. Guard it so the write only happens when it actually changes state:

```suggestion
                    if (invalidLimitWarned) {
                        invalidLimitWarned = false;
                    }
```

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
```diff
@@ -65,10 +65,49 @@ public boolean handleRequestInFlow(MessageContext messageContext) {
...
+    /**
+     * Whether the request payload should be captured (and the message therefore built) on the
+     * request-out flow. Mirrors the skip conditions {@code handleResponseOutFlow} applies so we never
+     * build/re-serialize a request for a call that will not publish an analytics event: websocket
+     * subscriptions, file-based API contexts, and requests explicitly flagged {@code SKIP_METRICS_PUBLISHING}.
+     * Cheap checks are evaluated first and the file-based-context lookup last.
+     */
+    private boolean shouldCaptureRequestPayload(MessageContext messageContext) {
+        if (messageContext.getPropertyKeySet().contains(InboundWebsocketConstants.WEBSOCKET_SUBSCRIBER_PATH)
+                || !AnalyticsPayloadUtil.shouldSendPayloads()) {
```
AnuGayan, line 100: 🟠 **major — not gated on analytics actually being enabled**

`AnalyticsPayloadUtil.shouldSendPayloads()` reads the raw static `APIManagerConfiguration.getAnalyticsProperties()`, which is populated whenever the `<Analytics>` element is parsed — **regardless of `<Enabled>`**.

Compare the sibling `shouldSendHeaders()` (`SynapseAnalyticsDataProvider.java:892`), which reads `getReporterProperties()`; that map is only assigned inside `if (this.analyticsEnabled)` (`APIManagerAnalyticsConfiguration.java:76`), so the header path inherits the enablement gate for free. This path does not.

**Failure scenario:** operator sets `enable = false` under `[apim.analytics]` but leaves `send_payloads = "true"` in `[apim.analytics.properties]` — a very plausible "turn analytics off for now" state. `handleRequestOutFlow` then calls `RelayUtils.buildMessage` on **every** request, adding a full in-memory body materialisation plus re-serialisation to the hot path, and the result is discarded because nothing publishes.

`APIUtil.isAnalyticsEnabled()` is the established idiom in this module (`WebsocketHandler`, `InboundWebSocketProcessor`, `DeliveryStatusUpdater`, …):

```suggestion
        if (!APIUtil.isAnalyticsEnabled()
                || messageContext.getPropertyKeySet().contains(InboundWebsocketConstants.WEBSOCKET_SUBSCRIBER_PATH)
                || !AnalyticsPayloadUtil.shouldSendPayloads()) {
```

Needs `import org.wso2.carbon.apimgt.impl.utils.APIUtil;`.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
```diff
@@ -65,10 +65,49 @@ public boolean handleRequestInFlow(MessageContext messageContext) {
...
+        // at request-in instead would run addressing before dispatch and break the send.
+        if (shouldCaptureRequestPayload(messageContext)) {
+            AnalyticsPayloadUtil.CapturedBody requestBody =
+                    AnalyticsPayloadUtil.extractPayload(messageContext,
+                            AnalyticsPayloadUtil.getPayloadSizeLimit(), "request");
+            if (requestBody != null && requestBody.getBody() != null) {
+                messageContext.setProperty(Constants.REQUEST_BODY, requestBody.getBody());
+                if (requestBody.getTransferEncoding() != null) {
+                    messageContext.setProperty(Constants.REQUEST_BODY_TRANSFER_ENCODING,
+                            requestBody.getTransferEncoding());
```
AnuGayan: 🟡 **minor — use a namespaced internal property key here** (see the paired comment on `Constants.java`)

`Constants.REQUEST_BODY` is `"requestBody"` — an unprefixed name, unlike every other analytics message-context property (`apim.analytics.request.start.time`, `api.analytics.user.ip`). Synapse copies all properties from the out context to the in context, so a user policy's `<property name="requestBody">` would be read back at collection time and published.

```suggestion
                messageContext.setProperty(Constants.REQUEST_BODY_PROPERTY, requestBody.getBody());
                if (requestBody.getTransferEncoding() != null) {
                    messageContext.setProperty(Constants.REQUEST_BODY_TRANSFER_ENCODING_PROPERTY,
                            requestBody.getTransferEncoding());
```

The matching read in `SynapseAnalyticsDataProvider.getProperties()` needs the same rename; the wire keys (`Constants.REQUEST_BODY` etc.) stay as-is in the `custom` map so the publisher contract is untouched.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/Constants.java
```diff
@@ -90,6 +90,17 @@ public static final class ERROR_CODE_RANGES {
...
+    public static final String REQUEST_BODY = "requestBody";
```
AnuGayan, line 94: 🟡 **minor — analytics wire keys reused as Synapse message-context property names**

`REQUEST_BODY` serves double duty: the publisher-contract key *and* the message-context property key at `AnalyticsMetricsHandler.java:78`. Every other analytics MC property in this file is namespaced (`apim.analytics.request.start.time`, `api.analytics.user.ip`, `api.analytics.cacheHit`) precisely to avoid collisions — and because Synapse copies all properties from the out context to the in context, an unprefixed `requestBody` set by any user policy would be read back here and published (or our internal value exposed to their mediation).

Keep the wire key, add separate internal keys:

```suggestion
    public static final String REQUEST_BODY = "requestBody";
    // Internal message-context property keys — namespaced like every other analytics MC property in
    // this class, so a user policy's <property name="requestBody"/> cannot collide with them.
    public static final String REQUEST_BODY_PROPERTY = "apim.analytics.request.body";
    public static final String REQUEST_BODY_TRANSFER_ENCODING_PROPERTY =
            "apim.analytics.request.body.encoding";
```

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/SynapseAnalyticsDataProvider.java
```diff
@@ -514,6 +514,36 @@ public Map<String, Object> getProperties() {
...
+            int sizeLimit = AnalyticsPayloadUtil.getPayloadSizeLimit();
+            // Request body was captured in the request flow and stashed on the message context.
+            Object requestBody = messageContext.getProperty(Constants.REQUEST_BODY);
+            if (requestBody instanceof String) {
+                custom.put(Constants.REQUEST_BODY, requestBody);
```
AnuGayan, line 525: 🟠 **major — captured bodies bypass the masking pipeline entirely (security / privacy)**

Headers go through two layers of protection before publishing:
1. `AnalyticsMetricsHandler.getUserAgentAndCopyRequestHeadersToContext` strips `Authorization` / `apikey` (`AnalyticsMetricsHandler.java:163-164`)
2. this same method applies `applyMask(..., parseMaskSet(getMaskProperties().get(REQUEST_HEADER_MASK)))` (line 492, ~25 lines above)

Bodies get **neither** — they are published verbatim, here and at line 540.

**Failure scenario:** an API fronting a token/login endpoint. `POST /token` with `client_secret=…&username=…&password=…`, or a JSON body containing PAN/SSN/PHI, is exported in full to Moesif and retained there. The header pipeline was built specifically to prevent this for `Authorization`; the body path reintroduces it through a much wider channel.

Suggested fix — reuse the existing mask infrastructure rather than inventing a new one:

```java
// Constants.java
public static final String REQUEST_BODY_MASK = "request_body";
public static final String RESPONSE_BODY_MASK = "response_body";
```

and redact the configured field names in the captured body before `custom.put(...)`, mirroring `applyMask`. For JSON, parse and replace matched values with `Constants.MASK_VALUE`; if a body cannot be parsed for redaction, drop it rather than publish it unredacted. Keep it fail-safe — a redaction error must never break the call.

At minimum: (a) skip capture for the token/revoke/introspect internal contexts, and (b) state prominently in the config docs that bodies are exported **unredacted** and that `send_payloads` must not be enabled on APIs handling credentials or regulated data.

`send_payloads` being default-off keeps this out of blocker territory, but shipping a body-export feature with no redaction hook at all will come back as a customer escalation.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/SynapseAnalyticsDataProvider.java
```diff
@@ -514,6 +514,36 @@ public Map<String, Object> getProperties() {
...
+                Object reqEncoding = messageContext.getProperty(Constants.REQUEST_BODY_TRANSFER_ENCODING);
+                if (reqEncoding instanceof String) {
+                    custom.put(Constants.REQUEST_BODY_TRANSFER_ENCODING, reqEncoding);
+                }
+                // Convey the request Content-Type so Moesif can label/parse the body even when
+                // send_headers is off. (The response Content-Type already rides as responseContentType.)
+                String reqContentType = getRequestContentType();
```
AnuGayan, line 532: 🟡 **minor — reads request-*in* headers, not the captured body's content type**

The body is captured at request-**out** (`AnalyticsMetricsHandler:73`), after mediation. If a policy transformed the payload (JSON→XML, MCP, the AI mediators), `requestContentType` describes the client's original type while `requestBody` holds the mediated one — the two disagree.

It also duplicates logic `AnalyticsPayloadUtil.getContentType()` already implements against the correct headers.

Cleaner: have `CapturedBody` carry the content type observed at capture time and stash it on the message context alongside the body, then read it back here. That deletes `getRequestContentType()` (lines 880-907) and its whole `getAnalyticsMetadata()` traversal — about 20 lines — and makes the label always agree with the bytes.

Minor inconsistency while you're here: `RESPONSE_CONTENT_TYPE` falls back to `UNKNOWN_VALUE` (line 469 / `getResponseContentType`) whereas the request side omits the key entirely when null. Worth aligning.

AnuGayan (changes requested):
## Review summary

Adds opt-in request/response body capture on the synapse gateway and places it into the analytics event `properties` for the (already-merged) [`apim-analytics-publisher#168`](https://github.com/wso2/apim-analytics-publisher/pull/168) to forward to Moesif.

The design is thoughtful and the hard part is right: the **pre-build `Content-Length` gate**, **drop-never-truncate**, **chunked bodies skipped by default**, and the `RuntimeException` catch-all so a capture bug can never break a proxied call. The PR description is excellent — the two-repo split, property-key contract and default-off rationale are all clear, and the `1.2.38` bump is correctly ordered behind the merged publisher PR.

My concerns are concentrated in three places, detailed as inline comments:

1. 🟠 **No redaction/masking path for bodies** — headers get two layers (`Authorization`/`apikey` stripped at `AnalyticsMetricsHandler:163-164`, then `applyMask` at `SynapseAnalyticsDataProvider:492`); bodies get neither.
2. 🟠 **`application/x-www-form-urlencoded` bodies are captured as XML gibberish**, and re-serialisation may alter what the backend receives.
3. 🟠 **Zero tests** for 367 lines of new request-path code.

### Verdict

🔧 **Request changes** — feature is well-engineered, but body export needs a redaction hook, form-encoded content is captured wrong, and none of it is covered by a test.

---

## Findings not anchorable to a diff line

### 🟠 No tests for 367 lines of new hot-path code

`AnalyticsPayloadUtil` has four public methods and none are exercised. `SynapseAnalyticsDataProviderTestCase.java` already exists in this package, so the harness (PowerMock + mocked `Axis2MessageContext`) is right there. Per the repo standard, this should not merge without them. The branches that most need pinning — all reachable with a mocked axis2 context, no network:

- `getPayloadSizeLimit()` — unset / valid / `0` / `-1` / `"abc"` → default fallback + warn-once
- `extractPayload` gates — `NO_ENTITY_BODY`, excluded content types, `Content-Length` over limit. **Assert `RelayUtils.buildMessage` was never called** — that is the memory-safety guarantee the whole design rests on
- `declaredLength < 0` skip vs. capture under `capture_payloads_without_content_length=true`
- byte-vs-char limit: a JSON body of 60 000 multi-byte UTF-8 characters against a limit of 100 000 must be **dropped** (`String.length()` would have kept it)
- `shouldCaptureRequestPayload` — each of the four skip conditions independently

### 🟡 Request bodies are retained on the message context for the full request lifetime

The captured string (up to `payload_size_limit`, default 100 000 bytes) lives on the MC from request-out through response-out and is copied onto the response context by Synapse. At 1 000 concurrent in-flight requests that is ~100 MB of retained heap on top of the transient build cost, and up to ~200 KB per event once the response body joins it. Worth calling out explicitly in the config docs; I'd also consider a lower default (32 KB).

### 🟡 Request bodies are absent from fault events

Capture happens in `handleRequestOutFlow`, which never runs for requests rejected before dispatch — auth failures (401), throttling (429), schema validation, guardrail hits. Those are exactly the events a Moesif user most wants a request body for. Probably out of scope here, but it should be a documented limitation rather than a surprise.

### 🔵 Doc nits

- `extractPayload`'s `@param sizeLimit` says *"maximum number of characters (text) or bytes (binary)"*, contradicting both the implementation and the paragraph three lines above it (*"The limit is compared against bytes throughout"*). Same wording on `getPayloadSizeLimit()`. Say bytes.
- `DEFAULT_PAYLOAD_SIZE_LIMIT` — a `_BYTES` suffix would make the unit self-documenting at the call site.
- `AnalyticsPayloadUtil` could be `final` given the private constructor.

---

## What's good

- The **pre-build `Content-Length` gate** is the right call and is genuinely the hard part of this feature. Dropping before `buildMessage` rather than building-then-measuring is what makes the default configuration safe, and it is documented well enough that a future maintainer won't "simplify" it away.
- **Drop, never truncate** — the publisher only ever sees a whole valid body, avoiding a class of downstream parse failures.
- **Chunked capture as a separate explicit opt-in**, with the memory trade-off spelled out, is the correct default posture.
- **Fail-safe error handling** — catching `RuntimeException` alongside `IOException`/`XMLStreamException` is exactly right for a global synapse handler.
- `shouldCaptureRequestPayload` mirrors `handleResponseOutFlow`'s skip set so no request is built for a call that will never publish — and it uses `Boolean.TRUE.equals(...)` rather than the existing `(Boolean)` cast, which is the safer idiom.
- Debug-level "why is this body missing" logging with actionable remediation text will save real support cycles.

---

<details>
<summary><b>📋 Prompt for an agent to verify &amp; fix these findings</b> (click to expand)</summary>

Paste the following into Claude Code (or your agent of choice) with this branch checked out. It is written to **verify before changing** — several findings are claims about Axis2/Synapse runtime behaviour that deserve to be confirmed against the actual builders rather than taken on faith.

````
You are working in the wso2/carbon-apimgt repo on the branch for PR #13927
("Request/Response body support for Moesif analytics"). Five files changed:
AnalyticsPayloadUtil.java (new), AnalyticsMetricsHandler.java,
SynapseAnalyticsDataProvider.java, Constants.java (all under
components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/),
and the root pom.xml.

A review raised the findings below. For EACH one: first VERIFY the claim against
the actual code and the Axis2/Synapse APIs on the classpath, report whether it is
confirmed or refuted with the evidence you found, and only then apply the fix.
Do not fix anything you could not confirm — report it as unconfirmed instead.

FINDING 1 (major, correctness) — form-urlencoded bodies captured as XML.
Claim: for Content-Type: application/x-www-form-urlencoded, Axis2's
XFormURLEncodedBuilder materialises the body into an OM tree whose first element
matches neither BaseConstants.DEFAULT_BINARY_WRAPPER nor DEFAULT_TEXT_WRAPPER, so
extractPayload falls through to the XML branch and publishes
"<xformValues><a>1</a></xformValues>" while labelling it
requestContentType=application/x-www-form-urlencoded. Secondary claim: because
capture sets MESSAGE_BUILDER_INVOKED, the pass-through sender re-serialises via
XFormURLEncodedFormatter, so parameter order / repeated keys may not survive the
round-trip to the backend.
VERIFY: find XFormURLEncodedBuilder on the classpath (check the synapse/axis2
transport jars) and confirm the wrapper QName it produces. Confirm the builder is
actually selected for that content type in the pass-through path.
FIX if confirmed: add application/x-www-form-urlencoded to isExcludedContentType.

FINDING 2 (major, security) — captured bodies bypass the masking pipeline.
Claim: headers are protected twice (Authorization/apikey stripped in
AnalyticsMetricsHandler.getUserAgentAndCopyRequestHeadersToContext, then
applyMask/parseMaskSet against REQUEST_HEADER_MASK / RESPONSE_HEADER_MASK in
SynapseAnalyticsDataProvider.getProperties), while request/response bodies are
put into `custom` verbatim. So a POST /token carrying client_secret or a
password grant, or a JSON body with PII, is exported unredacted to Moesif.
VERIFY: read the existing mask plumbing end to end — Constants.REQUEST_HEADER_MASK,
APIManagerConfiguration.getAnalyticsMaskProperties(), applyMask, parseMaskSet —
and confirm nothing already covers bodies.
FIX if confirmed: extend the same mechanism to bodies rather than inventing a new
one. Add Constants.REQUEST_BODY_MASK = "request_body" and RESPONSE_BODY_MASK =
"response_body", and redact the configured field names in the captured body before
custom.put(...). For JSON, parse and replace matched field values with
Constants.MASK_VALUE; if a body cannot be parsed, drop it rather than publish it
unredacted. Keep it fail-safe: a redaction error must never break the call.

FINDING 3 (major, wasted work) — request capture is not gated on analytics being
enabled.
Claim: AnalyticsPayloadUtil.shouldSendPayloads() reads the static
APIManagerConfiguration.getAnalyticsProperties(), which is populated whenever the
<Analytics> element is parsed regardless of <Enabled>; whereas the sibling
shouldSendHeaders() reads getReporterProperties(), which
APIManagerAnalyticsConfiguration.setAPIManagerConfiguration only assigns inside
`if (this.analyticsEnabled)`. So with enable=false but send_payloads="true" left
in place, handleRequestOutFlow calls RelayUtils.buildMessage on every request and
throws the result away.
VERIFY: confirm the <Analytics> parsing in APIManagerConfiguration is
unconditional, and confirm APIUtil.isAnalyticsEnabled() (in
org.wso2.carbon.apimgt.impl.utils) is the idiom used elsewhere in this module.
FIX if confirmed: add !APIUtil.isAnalyticsEnabled() as the first short-circuit in
shouldCaptureRequestPayload, plus the import.

FINDING 4 (minor) — the no-Content-Length skip fires even when the message is
already built.
Claim: the skip's stated rationale is "sizing requires a full in-memory build",
which no longer holds once PassThroughConstants.MESSAGE_BUILDER_INVOKED is TRUE.
This is reachable: getResponseSize() builds when build_response_message=true and
runs at SynapseAnalyticsDataProvider.java:467, before the response capture at
line 537 — so with a chunked backend response the memory cost is paid and the body
is then dropped anyway.
VERIFY: confirm the call ordering inside getProperties(), and confirm
RelayUtils.buildMessage short-circuits on MESSAGE_BUILDER_INVOKED.
FIX if confirmed: skip only when the message is not already built.

FINDING 5 (minor) — XML/SOAP capture keeps only the first body element.
Claim: firstElement.toString() drops sibling elements and the SOAP header, so a
SOAP body with two children publishes only the first — silently partial data,
contradicting the "whole valid body or nothing" contract the size gate is built
around. getResponseSize() (line ~832) uses soapbody.toString() instead.
FIX: serialise every child of the SOAP body (soapBody.getChildElements()) rather
than just the first. Do NOT switch to soapBody.toString() — that adds a
<soapenv:Body> wrapper that was never in the REST XML body on the wire.

FINDING 6 (minor) — pointless base64 decode/re-encode round-trip.
Claim: the binary wrapper's text is ALREADY base64 of the raw bytes, so
Base64.encodeBase64String(Base64.decodeBase64(text)) returns a normalised copy of
its own input at the cost of an extra byte[] plus an extra String (~2.3x body size
in peak heap). Only the size check needs the decoded length.
VERIFY: confirm OMText.getText() on a binary wrapper backed by a DataHandler
returns base64 (check the Axiom version on the classpath) — if it does not, this
finding is refuted and the round-trip is load-bearing. Also check whether
commons-codec's encodeBase64String chunks or normalises in a way the current code
depends on.
FIX if confirmed: pass the wrapper text through untouched and derive the decoded
length arithmetically. Remove the then-unused commons-codec Base64 import.

FINDING 7 (minor, collision risk) — analytics wire keys reused as Synapse
message-context property names.
Claim: Constants.REQUEST_BODY = "requestBody" is used both as the publisher
contract key and as the MC property key in handleRequestOutFlow. Every other
analytics MC property in Constants.java is namespaced (apim.analytics.*,
api.analytics.*) precisely to avoid collisions — and Synapse copies all properties
from the out context to the in context, so an unprefixed `requestBody` set by any
user policy's <property name="requestBody"> would be read back and published.
VERIFY: confirm in the Synapse source (SynapseCallbackReceiver) that out-context
properties are copied wholesale to the response context — this is also the
mechanism the existing BACKEND_START_TIME_PROPERTY relies on.
FIX if confirmed: add separate namespaced internal keys
(apim.analytics.request.body, apim.analytics.request.body.encoding) for the MC,
keeping REQUEST_BODY / REQUEST_BODY_TRANSFER_ENCODING as the wire keys only.
Update both the setProperty and the getProperty sites.

FINDING 8 (minor) — getRequestContentType() reads the wrong direction.
Claim: it reads the request-IN headers from getAnalyticsMetadata(REQUEST_HEADERS),
but the body is captured at request-OUT after mediation. If a policy transformed
the payload (JSON->XML, MCP, AI mediators), requestContentType describes the
client's original type while requestBody holds the mediated one. It also
duplicates AnalyticsPayloadUtil.getContentType(), which already reads the correct
headers.
FIX: have CapturedBody carry the content type observed at capture time, stash it
on the MC alongside the body, and read it back in getProperties(). This deletes
getRequestContentType() and the getAnalyticsMetadata traversal entirely.

FINDING 9 (nits) — apply these directly:
- capture()'s `transferEncoding` parameter is null at all three call sites: dead
  parameter, remove it.
- isExcludedContentType uses lower.contains("event-stream"); tighten to
  startsWith("text/event-stream") to match the documented intent and the
  multipart/ check next to it.
- getPayloadSizeLimit writes the volatile invalidLimitWarned on every successful
  call (i.e. every request); guard it: if (invalidLimitWarned) { ... = false; }.
- Javadoc on extractPayload's @param sizeLimit and on getPayloadSizeLimit says
  "characters (text) or bytes (binary)", contradicting the byte-based
  implementation and the paragraph above it. Say bytes.
- Make AnalyticsPayloadUtil final (it has a private constructor).

FINDING 10 (major, coverage) — add AnalyticsPayloadUtilTestCase and extend
SynapseAnalyticsDataProviderTestCase, following the existing PowerMock +
mocked-Axis2MessageContext patterns in that test package. Required cases:
- getPayloadSizeLimit: unset, valid, "0", "-1", "abc" -> default fallback; assert
  the invalid-value warning fires once, not per call.
- extractPayload returns null AND RelayUtils.buildMessage is NEVER invoked for:
  NO_ENTITY_BODY set, excluded content types, and Content-Length > limit. Verifying
  the no-build is the point of these tests.
- declaredLength < 0: skipped by default; captured when
  capture_payloads_without_content_length=true.
- byte-vs-char limit: a JSON body of 60000 multi-byte UTF-8 characters against a
  limit of 100000 must be DROPPED.
- binary path returns transferEncoding="base64"; JSON/text/XML return null
  encoding.
- shouldCaptureRequestPayload returns false for each of: websocket subscriber
  path, send_payloads=false, SKIP_METRICS_PUBLISHING=true (set as a Boolean, as
  InMemoryAPIDeployer does with type="BOOLEAN"), file-based API context — and
  analytics-disabled once FINDING 3 is applied.

WHEN DONE:
1. mvn -q -pl components/apimgt/org.wso2.carbon.apimgt.gateway -am \
     -DskipTests=false test
   and make checkstyle pass (max line length 120, see ./checkstyle.xml).
2. Report a table: finding | confirmed/refuted/unconfirmed | evidence | action taken.
3. Do NOT change the pom.xml version bump — the publisher PR is merged and 1.2.38
   is correct.
4. Do NOT weaken the pre-build Content-Length gate or the default-off posture of
   capture_payloads_without_content_length; those are the memory-safety guarantees
   of the design.
````

</details>

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
vinupa: Resolved in the latest commit. `Content-Type: application/x-www-form-urlencoded` is now skipped for capture, the same way `SSE` and `multipart` already are. Because the skip happens before the message is built, the form body is no longer turned into XML, and it's also no longer rebuilt on the way to the backend.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
vinupa: Resolved in the latest commit. The XML/SOAP path now iterates through every child element of the body and joins them, instead of taking only the first. For a normal single-element XML body the result is the same as before; a body with several elements is no longer cut down to just the first. It does not use `soapBody.toString()`, so no extra `<soapenv:Body>` wrapper is added.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
vinupa: Resolved in the latest commit. The "skip when there's no Content-Length" rule now also checks whether the message was already built earlier in the flow (e.g. by response-size calculation or a mediator). If it was, the body is already in memory, so we capture it instead of dropping it for nothing. The size limit still applies afterwards.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
vinupa: This is on purpose. The wrapper's text is decoded and encoded again so the body we send is clean, standard Base64 that Moesif can always read, rather than trusting whatever form the builder left it in. The extra memory is small (it's capped by `payload_size_limit` and released as soon as the event is built) so it doesn't weaken the size protection. Leaving it as is.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
vinupa: Resolved in the latest commit. Removed the unused parameter from the `capture()` helper (it was always passed `null`). The binary path never used that helper, it still sets the `base64` encoding directly on `CapturedBody` so nothing else changes.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsPayloadUtil.java
vinupa: Resolved in the latest commit. The warning flag is now only written when it actually changes, so a normal request no longer does a needless write on the hot path.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
vinupa: Resolved in the latest commit. Capture now checks that analytics is turned on first. If analytics is off, nothing gets published anyway, so we no longer build the message for a body that would just be thrown away. This covers both the request and response paths in one place.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/AnalyticsMetricsHandler.java
vinupa: Resolved in the latest commit. The request body is now stashed under an internal, namespaced key so it can't clash with a user-set property.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/Constants.java
vinupa: Resolved in the latest commit. The request body is now stored on the message context under internal, namespaced keys (`apim.analytics.request.body` / `apim.analytics.request.body.encoding`), like the other analytics properties, so a user policy setting `<property name="requestBody"/>` can't collide with it. The keys sent to Moesif (`requestBody`, etc.) are unchanged, so nothing changes on the publisher side.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/SynapseAnalyticsDataProvider.java
vinupa: Body masking was left out on purpose. The existing masking works on header/field names in a flat list; masking inside a body is a new problem and only works cleanly for JSON, not XML, text, or binary. Body capture is off by default, and the docs call out that bodies are sent as-is and that sensitive APIs should keep it off or rely on Moesif's own masking rules.

components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/analytics/SynapseAnalyticsDataProvider.java
vinupa: The request content type is read from the incoming request headers on purpose, so it matches what the client actually sent (and the request headers we log). It only differs from the captured body if a policy rewrites the body along the way, which is rare. And since the publisher uses this value as the event's Content-Type, taking it from the rewritten body would end up replacing the client's real Content-Type on the event. So we keep it tied to the request headers.

AnuGayan (approved):

PR #13961, opened 2026-08-14
Resolve the reference API of an MCP Server at import time

### LINKED ISSUE #5174: EXISTING_API subtype MCP Server import fails cross-environment
### Description

Hi Team,

We observed the following error when importing the MCP server if the backend API does not exist in the target environment, or if the mcp_server.yaml file does not contain the correct UUID (the target environment API UUID) for the backend API under the apiOperationMapping tag.

Creating the backend API first is a possible approach. However, the challenge is that the mcp_server.yaml file needs to be updated with the backend API UUID after the import.

```
apiOperationMapping:
      apiId: API_UUID
      apiName: API_name
      apiVersion: v1
      apiContext: /API_context/verion
      backendOperation:
        target: /***
        verb: GET
  categories: []
```

This approach is not practical when automating the deployment process using these commands because the backend API UUID changes whenever the API is imported into a new environment. Therefore, manually updating the mcp_server.yaml file with the new UUID for each import time is not a feasible solution for an automated deployment workflow. 

Additionally, the following error message does not provide sufficient details about the actual cause of the import failure:

```ERROR - GlobalThrowableMapper Error while importing API: Failed to get API```

This error does not clearly indicate whether the failure occurred because the dependent backend API does not exist in the target environment or because the API UUID specified in the mcp_server.yaml file is invalid or does not match the target environment API UUID.

Providing a more specific error message would help users identify and resolve the issue more easily.

Therefore, could you please check this and let us know an option to import mcp-server without depending on the BE API UUID(check and update the UUID internally in the yaml file)? 

We appreciate your input here. 

### Steps to Reproduce

N/A

### Version

4.6.0

_No response_

### PR DESCRIPTION
**Pull request by Oshanath**
## Purpose

Importing an `EXISTING_API` subtype MCP Server into an environment other than the one it was exported from always failed, even when the API it is built on was present in the target environment.

Fixes: https://github.com/wso2/api-manager/issues/5174

## Problem

The exported `mcp_server.yaml` carries the reference API under `apiOperationMapping`:

```yaml
apiOperationMapping:
  apiId: edf6c2dd-bfc4-4d9f-bf0c-acd2615fc7ec   # UUID of the source environment
  apiName: BackendPizzaAPI
  apiVersion: 1.0.0
  apiContext: /backendpizza/1.0.0
```

The UUID belongs to the environment the artifact was exported from, and nothing resolved it against the target environment, so the import failed with an opaque error:

```
{"code":900967,"message":"General Error","description":"Server Error Occurred"}
```

```
WARN  - ApiMgtDAO Unable to find provider for API: testAPIsMCP in the database
INFO  - ImportUtils Cannot find : testAPIsMCP-1.0. Creating it.
ERROR - GlobalThrowableMapper Error while importing API:  Failed to get API
```

Two further defects kept the reason hidden:

- `resolveExistingMCPBackendAPI` guarded on `getAPIbyUUID` returning null, but that method throws for an API that does not exist and only returns null when the UUID is blank. The guard, and the message it carried, were unreachable for the case they were written for.
- `importMCPServer` rethrew through `APIManagementException(String, Throwable)`, a constructor that hardcodes `ExceptionCodes.INTERNAL_ERROR`, flattening every import failure into a generic server error.

## Approach

**Resolve the reference.** The reference API is looked up by its name and version in the importing environment and the mapping is updated with the local UUID. This is the same resolution already used for the dependent APIs of an API Product (`updateDependentApiUuids`), and it runs before the create and the update paths so both are covered. When the artifact does not carry a name and a version, the UUID it came with is left untouched and resolved as before.

**Report the reason.** A retrieval failure whose root cause is an `APIMgtResourceNotFoundException` now leaves the reference unresolved so the existing guard is reached; any other failure keeps propagating untouched. The guard reports the new `REFERENCE_API_NOT_FOUND` (904015), a 400 that names the API, the version and the id the artifact points at, in line with the other MCP codes that report an unusable artifact. `importMCPServer` carries the original error handler forward.

```
{"code":904015,"message":"Reference API not found",
 "description":"The API this MCP Server is built on (name 'BackendPizzaAPI', version '1.0.0', id '6958d589-...') does not exist in this environment. Create that API here before importing the MCP Server."}
```

## Testing

Three unit tests added to `ImportUtilsMCPServerTest`, each verified to fail without the change it covers.

Verified on a 4.7.0-SNAPSHOT pack by running a 27 case battery against an unpatched pack and a patched pack and diffing the two. The only difference is the intended one:

```
< IMPORT mcp, referenced API gone   | 500 | 900967 | General Error
> IMPORT mcp, referenced API gone   | 400 | 904015 | Reference API not found
```

API import, API Product import (with a present and with a missing dependent API), MCP Server import, corrupt archives, a request with no file, and every read, write, delete and export against a non existent UUID are unchanged.

Cases covered end to end:

| Case | Result |
|---|---|
| Reference UUID stale, name and version present | 200, mapping moved onto the local UUID |
| Reference UUID valid, name and version absent | 200, UUID left untouched |
| Reference UUID invalid, name and version absent | 400, the id is named |
| Name and version present, no such API here | 400, the API, the version and the id are named |

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
@@ -927,6 +927,9 @@ public enum ExceptionCodes implements ErrorHandler {
             "One or more MCP tools are duplicated."),
     INVALID_MCP_BACKEND_OPERATION(904014, "Invalid MCP backend operation", 400,
             "The backend operation '%s %s' does not match any resource in the referenced API '%s'."),
+    REFERENCE_API_NOT_FOUND(904015, "Reference API not found", 400,
+            "The API this MCP Server is built on (name '%s', version '%s', id '%s') does not exist in this "
+                    + "environment. Create that API here before importing the MCP Server."),
 
     // gateway notification related codes
     GATEWAY_NOTIFICATION_BAD_REQUEST(902052, "Invalid request for gateway notification", 400,
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
@@ -41,6 +41,7 @@ import io.swagger.v3.parser.ObjectMapperFactory;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.http.HttpResponse;
@@ -57,6 +58,7 @@ import org.wso2.carbon.apimgt.api.APIComplianceException;
 import org.wso2.carbon.apimgt.api.APIDefinition;
 import org.wso2.carbon.apimgt.api.APIDefinitionValidationResponse;
 import org.wso2.carbon.apimgt.api.APIManagementException;
+import org.wso2.carbon.apimgt.api.APIMgtResourceNotFoundException;
 import org.wso2.carbon.apimgt.api.APIProvider;
 import org.wso2.carbon.apimgt.api.ErrorHandler;
 import org.wso2.carbon.apimgt.api.ExceptionCodes;
@@ -193,6 +195,7 @@ import static org.wso2.carbon.apimgt.impl.APIConstants.REPUBLISH;
 public class PublisherCommonUtils {
 
     private static final Log log = LogFactory.getLog(PublisherCommonUtils.class);
+    private static final String UNKNOWN_REFERENCE = "UNKNOWN";
     public static final String SESSION_TIMEOUT_CONFIG_KEY = "sessionTimeOut";
     static APIMGovernanceService apimGovernanceService = ServiceReferenceHolder.getInstance()
             .getAPIMGovernanceService();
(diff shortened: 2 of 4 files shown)


components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
```diff
@@ -927,6 +927,10 @@ public enum ExceptionCodes implements ErrorHandler {
             "One or more MCP tools are duplicated."),
     INVALID_MCP_BACKEND_OPERATION(904014, "Invalid MCP backend operation", 400,
             "The backend operation '%s %s' does not match any resource in the referenced API '%s'."),
+    REFERENCE_API_NOT_FOUND(904015, "Reference API not found", 400,
+            "The referenced API (name '%s', version '%s', id '%s') does not exist in this environment. Create the "
+                    + "API before importing the MCP Server, or point apiOperationMapping in mcp_server.yaml at an "
```
PasanT9: Lets not include `apiOperationMapping` as it is. Use user friendly error messages.

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
```diff
@@ -193,6 +195,7 @@
 public class PublisherCommonUtils {
 
     private static final Log log = LogFactory.getLog(PublisherCommonUtils.class);
+    private static final String UNKNOWN_REFERENCE = "unknown";
```
PasanT9: Lets use `UNKNOWN`.

components/apimgt/org.wso2.carbon.apimgt.api/src/main/java/org/wso2/carbon/apimgt/api/ExceptionCodes.java
Oshanath: Done

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/mappings/PublisherCommonUtils.java
Oshanath: Done

PasanT9 (approved):

PR #13965, opened 2026-08-17
Fix AI API deployment failure after changing the primary endpoint

### LINKED ISSUE #5184: Cannot deploy AI APIs after changing the primary endpoint and delete default endpoint
### Description

$Subject

### Steps to Reproduce

Take an APIM pack and add a production gateway (the default one is hybrid)
Start the APIM pack and go to the publisher portal.
Create an AI API.
Under the Endpoints section, add another Production endpoint.
Then change the Primary Endpoint of the API.
Then delete the old endpoint.
Try to deploy that into the configured production gateway and are able to see the error.

### Version

4.6.0

_No response_

Oshanath:
## Purpose

Fixes https://github.com/wso2/api-manager/issues/5184

An AI API cannot be deployed to a production or sandbox gateway after its primary endpoint is switched to a newly added endpoint and the original endpoint is removed.

## Root cause

AI API endpoints are stored as named endpoints and selected through `primaryProductionEndpointId` and `primarySandboxEndpointId`. The production/sandbox deployment gate checked only the legacy `endpoint_config`, so it could reject a valid AI API before the AI endpoint list was used to generate its gateway artifacts.

## Approach

Build the production and sandbox endpoint collections once before the deployment gate, then reuse them in both the gate and the existing AI API artifact generation path.

- AI APIs are checked against their named endpoint collection.
- Non-AI APIs retain the existing `endpoint_config` checks.
- Production and sandbox checks remain separate and follow the prior gate structure.
- Null endpoint configuration behavior and endpoint artifact generation remain unchanged.

## Testing

The exact production and sandbox reproductions, including super-tenant and tenant scenarios, were validated while developing the fix. Tests were not rerun after reducing the implementation to this review-focused diff.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/TemplateBuilderUtil.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/TemplateBuilderUtil.java
@@ -1176,6 +1176,18 @@ public class TemplateBuilderUtil {
         }
     }
 
+    /**
+     * Checks whether the API has an endpoint for the given deployment stage.
+     * An AI API holds its endpoints as named endpoints outside the endpoint configuration, so it is checked
+     * against the endpoint collection of that stage. Every other API is checked against its endpoint
+     * configuration as before.
+     *
+     * @param api          The API being deployed
+     * @param endpointType The deployment stage to check, either
+     *                     {@link APIConstants.APIEndpoint#PRODUCTION} or {@link APIConstants.APIEndpoint#SANDBOX}
+     * @param endpoints    The endpoints of that deployment stage, considered only for AI APIs
+     * @return True if an endpoint exists for the given deployment stage, otherwise false
+     */
     private static boolean hasEndpoint(API api, String endpointType, List<SimplifiedEndpoint> endpoints) {
 
         if (APIConstants.API_SUBTYPE_AI_API.equals(api.getSubtype())) {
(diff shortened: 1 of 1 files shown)


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/TemplateBuilderUtil.java
```diff
@@ -1176,6 +1176,24 @@ private static void addEndpointsSequence(String type, List<SimplifiedEndpoint> e
...
+
+    private static boolean hasSandboxEndpoint(API api, boolean isAiApi,
+                                              List<SimplifiedEndpoint> sandboxEndpoints) {
+
+        if (isAiApi) {
+            return !sandboxEndpoints.isEmpty();
+        }
+        return APIUtil.isSandboxEndpointsExists(api.getEndpointConfig());
+    }
+
```
ashera96: Shall we merge these two methods and decide prod or sand based on a method argument?

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/TemplateBuilderUtil.java
```diff
@@ -1176,6 +1176,24 @@ private static void addEndpointsSequence(String type, List<SimplifiedEndpoint> e
         }
     }
 
+    private static boolean hasProductionEndpoint(API api, boolean isAiApi,
```
ashera96: Do we have to pass down isAiApi boolean to this method signature? It is redundant given that first argument can be used to deduce the second right?

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/TemplateBuilderUtil.java
Oshanath: Done

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/TemplateBuilderUtil.java
Oshanath: Done

ashera96 (dismissed):

ashera96 (approved):

PR #13968, opened 2026-08-18
Fix client certificate deletion and alias reuse across API revisions

### LINKED ISSUE #5175: Client Certificate Deletion Fails with a Unique Constraint Violation When the Alias Was Previously Used by a Different API/Product and Deleted
### Description

Deleting a client certificate from an API or API Product through the Publisher Portal fails with a SQLIntegrityConstraintViolationException (ORA-00001: unique constraint violated) when the same certificate alias was previously soft-deleted from a different API in the same tenant.

### Steps to Reproduce

1. Create two APIs (API-A and API-B), or create one API and one API Product.
2. Add a client certificate with the alias test_cert to API-A. Ref: [doc](https://apim.docs.wso2.com/en/4.1.0/design/api-security/api-authentication/secure-apis-using-mutual-ssl/#:~:text=Click%20Add%20Certificate%20to%20upload%20a%20new%20client%20certificate.)
3. Delete the certificate from API-A. This performs a soft delete (REMOVED = 1).
4. Add a client certificate with the same alias (test_cert) to API-B.
5. Delete the certificate from API-B.
6. A SQLIntegrityConstraintViolationException is thrown due to a unique constraint violation on the primary key.

### Version

4.2

_No response_

ashanhr:
### Issue

Fixes wso2/api-manager#5175

### Problem

Deleting a client certificate from an API or API Product fails with `SQLIntegrityConstraintViolationException`
(`ORA-00001` on Oracle) when the same alias was previously deleted from a **different** API in the same tenant.
After that the certificate cannot be removed through the product at all — only manual SQL clears it.

`AM_API_CLIENT_CERTIFICATE` has `PRIMARY KEY (ALIAS, TENANT_ID, KEY_TYPE, REMOVED, REVISION_UUID)` and does **not**
include `API_ID`. Deleting a certificate ran a pre-delete to clear the colliding `REMOVED = 1` row followed by an
`UPDATE` setting `REMOVED = 1`, but the pre-delete was additionally filtered by `API_ID` while the key it guards is
not. A row left behind by another API therefore survived, and the `UPDATE` duplicated the key.

The same missing `API_ID` causes a second defect: `checkWhetherAliasExist` only looks at
`REVISION_UUID = 'Current API'`, so once an API removed an alias from its working copy another API could take it
even though the first API's revisions still held it. Restoring that revision then failed the same way — a
duplicate `('alias', tenantId, keyType, FALSE, 'Current API')`.

### Changes

**1. Hard delete the client certificate entry.** `CertificateMgtDAO.deleteClientCertificate` now issues a single
`DELETE` instead of pre-delete + soft-delete `UPDATE`. `REMOVED` is deliberately left out of the `WHERE` clause so
stale entries from earlier builds are cleaned up on the next delete of that alias on that API.

The `REMOVED = 1` entry was only ever consumed by the pre-revision gateway sync, which no longer exists:
`getDeletedClientCertificateAliasOfGivenKeyType` and `updateRemovedCertificatesFromGateways` have no production
callers on master — a repo-wide search finds them only in `CertificateMgtDaoTest`. Every other reader of the table
binds `REMOVED = false`.

Note this is the shape the product already uses elsewhere: restoring a revision clears the current-copy rows with
`DELETE FROM AM_API_CLIENT_CERTIFICATE WHERE API_ID = ? AND REVISION_UUID='Current API'` — a hard delete with no
`REMOVED` predicate. This change makes the delete path consistent with the restore path.

**2. Reject an alias that a revision still holds.** Adding a client certificate now also checks whether a revision
of a *different* API of the same tenant holds the alias. Revisions of the API being added to are excluded via
`API_ID <> (…)`, so certificate rotation on the same API keeps working.

**3. Distinct response code.** The new rejection returns `ALIAS_EXISTS_IN_API_REVISION` rather than reusing
`ALIAS_EXISTS_IN_TRUST_STORE`. Two reasons:

- Users would otherwise get a 409 telling them the alias exists *in the trust store*, sending them to look for
  something that is not there.
- `ImportUtils.addClientCertificates` treats `ALIAS_EXISTS_IN_TRUST_STORE` as "already present here → overwrite it".
  When the alias exists only in a revision there is no current-copy row to update, so `updateClientCertificate`
  finds nothing and returns `false` after a debug log — the API would import **silently without its client
  certificate**. The importer now logs a warning naming the alias and the API instead.

### Difference from the 4.2.0 patch

`KEY_TYPE` entered the primary key and all of these queries after 4.3.0, so the new SQL carries `KEY_TYPE` and the
prepared-statement indices differ from the 4.2.0 version. `deleteClientCertificate`,
`checkWhetherAliasExist(String keyType, …)` and the import path all take `keyType` on master. Everything else is
the same change.

### Reviewer notes

- **Behaviour change:** adding a client certificate whose alias is held by another API's revision now returns
  **409** where it previously returned 201. Intentional — it is what prevents the revision-restore failure — but it
  is user-visible. Moving an alias between APIs now requires deleting the revisions holding it first.
- **Documentation:** the mutual-SSL docs do not currently mention client-certificate alias uniqueness or revisions,
  so this constraint is undocumented today. A docs update should follow.
- **Pre-existing corruption is not repaired.** Where an alias has *already* migrated between APIs, revision restore
  keeps failing until the duplicate certificate is removed from the API that took the alias. The fix prevents
  recurrence; it does not undo existing state.
- `PRE_DELETE_CERTIFICATES`, `PRE_DELETE_CERTIFICATES_WITHOUT_APIID`, `DELETE_CERTIFICATES` and
  `DELETE_CERTIFICATES_WITHOUT_APIID` in `ClientCertificateConstants` become unreferenced. Left in place to keep the
  diff minimal — happy to remove them if preferred.
- **Every supported line carries this defect** (4.0.0 through 4.7.0 and master), so backports will be wanted.
  4.0.0–4.3.0 need the version of this change *without* `KEY_TYPE`.

### Testing

**Build:** `org.wso2.carbon.apimgt.impl`, `org.wso2.carbon.apimgt.rest.api.publisher.v1.common` and
`org.wso2.carbon.apimgt.rest.api.publisher.v1` all build clean on master with JDK 21.

**Unit:** `CertificateMgtDaoTest` — 17/17 green. `testGetDeletedClientCertificates` was updated: it asserted that a
delete leaves a `REMOVED = 1` row behind, which is exactly what this change removes. It now asserts none remains.

**Runtime:** the equivalent change was verified end to end on a 4.2.0 pack (H2, super tenant and a non-super
tenant). Master itself has been build- and unit-verified only — there is no master pack to deploy against here.

| Scenario | Before | After |
|---|---|---|
| Delete a certificate whose alias was previously deleted from another API | 500 | **200** |
| Same, one API + one API Product | 500 | **200** |
| Add an alias held by another API's revision | 201 (silently created the corruption) | **409**, specific message |
| Restore a revision after the alias migrated | 500 | prevented upstream |
| Rotation on the same API with revisions holding the alias | 201 | **201** (unchanged) |
| Add an alias actively held by another API | 409 "already exists in the trust store" | **same message** (codes not conflated) |
| `PUT` update a client certificate | 200 | **200** |
| Import an API whose alias is reserved by another API's revision | certificate imported, creating corruption | warning logged, import succeeds without it |
| Non-super tenant: all of the above | as above | as above, and the new check does not leak across tenants |

Also verified against a database first populated by the **unpatched** build and then patched in place without
resetting the data: three pre-existing `REMOVED = 1` rows cleared themselves through ordinary operations, so
affected users need no manual SQL.


ashanhr:
@coderabbitai please review

ashanhr:
@coderabbitai I have dismissed your review by another commit, so can you approve again?

PasanT9 (approved):

PR #13977, opened 2026-08-30
Fix StackOverflowError on circular ref in MCP tool schema generation

### LINKED ISSUE #5189: StackOverflowError when generating MCP tools from an API definition with a circular reference
### Description

When an MCP Server is created from an existing API whose definition contains a circular $ref (a schema that refers back to itself through one of its properties), tool generation fails with a java.lang.StackOverflowError and the Publisher returns HTTP 500. The MCP Server is not created.

This affects both Swagger 2.0 and OpenAPI 3.0 definitions.

Recursive schemas are common in real APIs, for example a folder that contains sub-folders, a comment that contains replies, or a search filter that contains nested filters.

### Steps to Reproduce

1. Create an API using the following definition:
```json
   {
     "swagger": "2.0",
     "info": { "title": "RecursiveSearch", "version": "1.0.0" },
     "host": "localhost:9443",
     "basePath": "/search/versions/1",
     "schemes": ["https"],
     "consumes": ["application/json"],
     "produces": ["application/json"],
     "paths": {
       "/search": {
         "post": {
           "operationId": "search",
           "parameters": [
             { "in": "body", "name": "queryBody", "required": true,
               "schema": { "$ref": "#/definitions/SearchRequest" } }
           ],
           "responses": { "200": { "description": "OK" } }
         }
       }
     },
     "definitions": {
       "SearchRequest": {
         "type": "object",
         "properties": {
           "query":   { "type": "string" },
           "filters": {
             "type": "array",
             "items": { "$ref": "#/definitions/SearchRequest" }
           }
         }
       }
     }
   }
```

   Note that SearchRequest contains a list of SearchRequest.

2. In the Publisher, go to MCP Servers and create a new MCP Server from an existing API, selecting the API created above as the backend.

3. Select the POST /search operation as a tool and click Create.

4. The request fails with HTTP 500. The following appears in wso2carbon.log:
```
   java.lang.StackOverflowError
       at org.wso2.carbon.apimgt.spec.parser.definitions.OAS2Parser.resolveProperty(OAS2Parser.java)
       at org.wso2.carbon.apimgt.spec.parser.definitions.OAS2Parser.resolveProperty(OAS2Parser.java)
       at org.wso2.carbon.apimgt.spec.parser.definitions.OAS2Parser.resolveModel(OAS2Parser.java)
       ... 
```

5. Repeating the same steps with the equivalent OpenAPI 3.0 definition produces the same failure, with the stack trace in OAS3Parser.resolveSchema.

### Version

4.7.0

_No response_

Piumal1999:
## Purpose

Fix https://github.com/wso2/api-manager/issues/5189

## Issue

Creating an MCP Server from a backend API whose definition contains a circular `$ref` fails — `StackOverflowError` in `OAS2Parser`, endless recursion in `OAS3Parser`.

```yaml
Comment:
  properties:
    text:   { type: string }
    parent: { $ref: '#/definitions/Comment' }   # self-reference
```

## Approach

* Circular `$ref` chains now terminate in both the Swagger 2.0 and OpenAPI 3 parsers, for every shape they take: through a field, a list, an inline object, `allOf` (both via a property and as an `allOf`-only chain), a map value type, `oneOf`, and indirect chains across several models.
* Resolution no longer modifies the parsed API definition, so the schema generated for one tool cannot affect the schemas generated for others.
* Three cases that previously emitted an unresolvable `$ref` into the tool schema are corrected: a property referencing an `allOf` model, a property written as an inline `allOf`, and a reference to a definition that does not exist (now omitted, with a warning logged).
* A misleading `Circular reference detected for model: …` warning has been removed. It was emitted for any `$ref` request body, cyclic or not, so it is not an indicator of this issue.

## What changes in the generated tool schema

A reference that closes a cycle resolves to an object with no declared properties:

```json
"parent": {
  "type": "object",
  "description": "Circular reference to 'Comment'; nested properties are omitted."
}
```

The field remains declared, so an agent still knows it exists and can provide a value; only the nested shape is omitted. Definitions without circular references are unaffected and expand exactly as before. An agent can only infer a suitable value for the circular field, since with the MCP protocol version we currently use there is no way to include the definitions.

## Known limitation

MCP protocol version 2025-06-18 defines a tool's `inputSchema` as `type`/`properties`/`required` only, so a recursive structure cannot be represented by reference. Omitting the nested shape where the cycle closes is therefore the only conformant option at this protocol version. MCP 2026-07-28 permits `$defs`/`$ref`, which would allow recursion to be expressed exactly; that has to be tracked separately since it is beyond the scope of this patch.

### DIFF
--- a/components/apimgt/org.wso2.carbon.apimgt.spec.parser/src/main/java/org/wso2/carbon/apimgt/spec/parser/definitions/OAS2Parser.java
+++ b/components/apimgt/org.wso2.carbon.apimgt.spec.parser/src/main/java/org/wso2/carbon/apimgt/spec/parser/definitions/OAS2Parser.java
@@ -58,6 +58,7 @@ import io.swagger.models.parameters.PathParameter;
 import io.swagger.models.parameters.RefParameter;
 import io.swagger.models.properties.AbstractProperty;
 import io.swagger.models.properties.ArrayProperty;
+import io.swagger.models.properties.ComposedProperty;
 import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.ObjectProperty;
 import io.swagger.models.properties.Property;
@@ -3020,6 +3021,11 @@ public class OAS2Parser extends APIDefinition {
                     }
                     return objProp;
                 }
+                // The definition the reference names does not exist. Drop the field rather than
+                // emitting the reference unresolved: the generated schema carries no definitions
+                // section, so a surviving $ref points at nothing.
+                log.warn("Component not found in reference: " + ref);
+                return null;
             } finally {
                 visitedRefs.remove(ref);
             }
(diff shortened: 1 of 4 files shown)


Piumal1999:
@coderabbitai review

PasanT9 (dismissed):

PasanT9 (approved):

PR #13996, opened 2026-09-17
Fix NPE when endpoint stage is explicit null in MCP/API deployment

DakshithaS:
## Summary
Fixes several related failures caused by the same underlying mistake: code that decides whether an
endpoint stage (`production_endpoints` / `sandbox_endpoints`) is "configured" checks key **presence**
instead of **value validity**, so an explicit JSON `null` (as opposed to the key being absent) slips
past the check and reaches code that assumes a real value.

- A NullPointerException raised during API/MCP artifact **deployment** when a payload contains an
  explicit JSON `null` for an endpoint stage, instead of omitting the key entirely.
- An HTTP 500 (`JSONException`) raised during API **revision creation** for the same explicit-null
  endpoint stage, on both regular and `failover`-type endpoints.

## Root cause
- `APIUtil#isSandboxEndpointsExists` / `#isProductionEndpointsExists` (top-level and nested GraphQL
  `http` checks) only verified key presence, not that the value was non-null. Downstream,
  `TemplateUtilContext#escapeXml` had no null guard, so a null endpoint URL surfaced as an unqualified
  NPE with no actionable message during gateway artifact generation.
- `OASParserUtil#setPrimaryConfig` and `#populateFailoverConfig` used `JSONObject.has(...)` before
  `getJSONObject(...)`. `has()` returns `true` even when the value is JSON `null`, so the following
  `getJSONObject()` call threw `JSONException: ... is not a JSONObject (class org.json.JSONObject$Null)`
  at revision-creation time, before deployment is ever reached.
- `populateFailoverConfig` also called `updateEndpointResult(primaryEndpoints, ...)` unconditionally,
  with no null check — unlike `setPrimaryConfig`, which only calls it when `primaryEndpoints != null`.
  Once `primaryEndpoints` can legitimately be `null` (an absent or now-correctly-nulled endpoint stage,
  e.g. a failover-only config with no primary endpoint), this throws a `NullPointerException` one call
  deeper.

## Changes
- `APIUtil.java`: both endpoint-existence checks (top-level and nested GraphQL) now require
  `get(...) != null` in addition to `containsKey(...)`.
- `TemplateUtilContext.java`: `escapeXml(String url)` now throws an `IllegalArgumentException` with a
  clear message when `url == null`, instead of failing with a bare NPE further down the call chain.
- `OASParserUtil.java`: `setPrimaryConfig` and `populateFailoverConfig` now use `!isNull(...)` instead of
  `has(...)` for both `production_endpoints` and `sandbox_endpoints`, so an explicit `null` is treated
  the same as an absent stage. `populateFailoverConfig`'s final `updateEndpointResult` call is now
  guarded against a null `primaryEndpoints`, matching the pattern already used in `setPrimaryConfig`.
- Added unit tests in `APIUtilTest`, `TemplateUtilContextTest`, and `TemplateBuilderUtilTest` covering
  explicit-null, absent-key, and nested GraphQL endpoint-stage combinations, plus non-regression checks
  for existing `escapeXml` behavior.

## Related Issue
- https://github.com/wso2/api-manager/issues/5213


components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/template/TemplateUtilContext.java
```diff
@@ -41,6 +41,10 @@ public VelocityContext getContext() {
     }
 
     public String escapeXml(String url) {
+        if (url == null) {
```
ashera96, line 44: Shall we add the empty string check as well?

components/apimgt/org.wso2.carbon.apimgt.spec.parser/src/main/java/org/wso2/carbon/apimgt/spec/parser/definitions/OASParserUtil.java
```diff
@@ -1758,7 +1758,10 @@ private static ObjectNode populateFailoverConfig(JSONObject endpointConfig, bool
         ObjectNode endpointResult = objectMapper.createObjectNode();
         endpointResult.set(APISpecParserConstants.ENDPOINT_URLS, endpointsArray);
         endpointResult.put(APISpecParserConstants.X_WSO2_ENDPOINT_TYPE, APISpecParserConstants.ENDPOINT_TYPE_FAILOVER);
-        return updateEndpointResult(primaryEndpoints, endpointResult);
+        if (primaryEndpoints != null) {
+            return updateEndpointResult(primaryEndpoints, endpointResult);
+        }
+        return endpointResult;
```
ashera96, line 1764: Shall we add the null check inside the `updateEndpointResult` instead of this change?

DakshithaS: Agreed — moved the null check into `updateEndpointResult`. It's private with three callers; only `populateFailoverConfig` can pass `null` (failover-only config), so behaviour is unchanged, and it also covers the pre-existing case where the primary stage is absent rather than explicitly null.

components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1.common/src/main/java/org/wso2/carbon/apimgt/rest/api/publisher/v1/common/template/TemplateUtilContext.java
DakshithaS: Checked this, including at runtime. `escapeXml("")` (and whitespace-only) already returns `""` safely; only `null` NPEs. I also tried adding the empty check on a 4.7.0 pack: an OAuth-secured API behind an unauthenticated proxy (what the Publisher UI sends, `proxyUsername: ""`) then fails at `endpoint_template.xml` line 152 (`<proxyUsername>`). Deploy-revision still returns 201, but `deployedGatewayCount` stays 0. It deploys fine with the null-only guard.

`escapeXml` is also used for non-URL fields that are legitimately empty (unauthenticated proxy username/password, empty `customParameters` for OAuth without custom params, custom param values, `$!llmProviderId`), so throwing on empty would break valid deployments. Empty endpoint URLs are already rejected at create/update time by `PublisherCommonUtils.validateEndpoints` → `APIUtil.validateEndpointURL`.

So I've kept the guard limited to `null`. Happy to add URL-specific validation at that layer if you'd like.

components/apimgt/org.wso2.carbon.apimgt.impl/src/main/java/org/wso2/carbon/apimgt/impl/utils/APIUtil.java
```diff
@@ -5049,12 +5049,12 @@ public static boolean isSandboxEndpointsExists(String endpointConfig) {
         try {
             config = (JSONObject) parser.parse(endpointConfig);
 
-            if (config.containsKey("sandbox_endpoints")) {
+            if (config.containsKey("sandbox_endpoints") && config.get("sandbox_endpoints") != null) {
```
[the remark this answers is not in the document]
DakshithaS: Addressed — thanks for catching this. Guarded both `endpoint_type` and the nested GraphQL `http` field before dereferencing them, in both `isSandboxEndpointsExists` and `isProductionEndpointsExists`. No behavior change for well-formed configs; only stops the NPE for the missing/null case.

